PR #4, opened 2020-10-31
Fix clippy errors

TaKO8Ki:
I've fixed clippy errors and added a CI job for `cargo fmt` and `cargo clippy`!


TaKO8Ki:
Now `cargo clippy` causes many warnings. Can I fix them in this pull request? Or should I do this in another pull request?

Issue #13, opened 2020-11-17
[DTLS] implement crypto_ccm.rs


rainliu:
as title

zeroclock:
@rainliu 
I've started to implement this feature and created draft PR.
If duplicate work, please reject it.

rainliu:
close this issue as it is implemented in https://github.com/webrtc-rs/webrtc/pull/22

PR #22, opened 2020-11-26
DTLS: crypto_ccm.rs implementation

zeroclock:
related issue: #13


zeroclock:
@rainliu
I've implemented crypto_ccm.rs and also added a simple test case.

The test case doesn't exist in Pion, so please let me know if you don't need it as it can be covered by another test case. I'll remove it.

Coverage check fail seems to be related to rust-lang/rust#79487 .

rainliu:
@zeroclock, thank you for adding crypto_ccm.rs and corresponding test case.

Issue #12, opened 2020-11-17
[DTLS] add more cipher_suite implementations


rainliu:
Pion implements the following cipher suites, but here only CipherSuiteTLSEcdheEcdsaWithAes128GcmSha256/CipherSuiteTLSPskWithAes128GcmSha256 and CipherSuiteTLSEcdheEcdsaWithAes256CbcSha are implemented.

```
// Supported Cipher Suites
const (
	// AES-128-CCM
	TLS_ECDHE_ECDSA_WITH_AES_128_CCM   CipherSuiteID = 0xc0ac //TODO
	TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 CipherSuiteID = 0xc0ae //TODO

	// AES-128-GCM-SHA256
	TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 CipherSuiteID = 0xc02b //Done
	TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256   CipherSuiteID = 0xc02f //TODO

	// AES-256-CBC-SHA
	TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA CipherSuiteID = 0xc00a //Done
	TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA   CipherSuiteID = 0xc014 //TODO

	TLS_PSK_WITH_AES_128_CCM        CipherSuiteID = 0xc0a4 //TODO
	TLS_PSK_WITH_AES_128_CCM_8      CipherSuiteID = 0xc0a8 //TODO
	TLS_PSK_WITH_AES_128_GCM_SHA256 CipherSuiteID = 0x00a8 //Done
)

```

rainliu:
@zeroclock, are you interested in be owner of this task?

zeroclock:
@rainliu 
Yes, I'm interested in implementing them.

rainliu:
Thanks @zeroclock for adding CCM-related cipher suites implementation!

Now we only have TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 and TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA left.

PR #29, opened 2020-12-07
[DTLS] Add RSA-related cipher suites

zeroclock:
related: #12


zeroclock:
@rainliu 
I've implemented remaining cipher suites.
For now, Rust doesn't support struct embedding as Go does, so I added `rsa` field to `CipherSuiteAes256CbcSha` / `CipherSuiteAes128GcmSha256` to be able to handle both RSA and ECDSA patterns.
If better solution, please let me know.

Issue #24, opened 2020-11-29
[DTLS] add NamedCurve::X25519 and NamedCurve::P384 support 


rainliu:
Currently only NamedCurve::P256 is supported.

https://github.com/webrtc-rs/webrtc/blob/master/dtls/src/curve/named_curve.rs

zeroclock:
Hi, @rainliu .
Can I work on this issue? I've started to implement `NamedCurve::X25519` .

k0nserv:
Merged in #254, not yet released. Will be in 0.5.0

PR #32, opened 2020-12-09
[DTLS] Add NamedCurve::X25519

zeroclock:
related: #24


zeroclock:
@rainliu
I'm ready for your review.
I've adopted `x25519_dalek::StaticSecret` as `NamedCurvePrivateKey` because `x25519_dalek::EphemeralSecret` can be used only once to get public key.

PR #47, opened 2021-01-09
Start translating e2e tests from pion

littlebenlittle:
This is a first crack at issue #42.

This is also my first crack at writing Rust, so I'm looking for feedback and suggestions. My code is almost certainly not idiomatic and I've probably made some simple mistakes with ownership, particularly across async calls.

See the [tests README](https://github.com/littlebenlittle/webrtc/blob/e2e/tests/README.md) for general info.


tests/basic.rs
```diff
@@ -99,57 +49,78 @@ pub async fn run_server(
...
+    let mut buf = vec![0_u8; 8192];
+    loop {
+        match stream.readable().await {
+            Ok(_) => {}
+            Err(e) => {
+                err_chan.send(e);
+                return;
+            }
+        }
+        match stream.try_read(&mut buf) {
```
littlebenlittle, line 77: Calling `from_utf8` on `&buf[0..n]` is resulting in "borrowed value does not live long enough". I am just still just learning about borrowing and my understanding is that the scope of `buf` is `68-124`, so I think I should be able to borrow that value here. What mistake am I making?

tests/basic.rs
```diff
@@ -0,0 +1,220 @@
...
+    stream: tokio::net::TcpStream,
+    out_buffer: Arc<Mutex<[u8; 8192]>>,
+    writer_join_handle: &mut tokio::task::JoinHandle<Result<(), std::io::Error>>,
+    reader_join_handle: &mut tokio::task::JoinHandle<Result<(), std::io::Error>>,
+) -> Result<(), std::io::Error>
+{
+    let ref arc_mx_stream = Arc::new(Mutex::new(stream));
+    let read = Arc::clone(arc_mx_stream);
+    // Read from stream into out buffer
+    *reader_join_handle = tokio::spawn( async {
```
littlebenlittle, line 38: It seem like I'm not using `Arc<Mutex<_>>` properly here. Is there another way to pass the buffer back up the execution chain?

```
future cannot be sent between threads safely

future created by async block is not `Send`

help: within `impl std::future::Future`, the trait `std::marker::Send` is not implemented for `std::sync::MutexGuard<'_, tokio::net::TcpStream>`
```

tests/basic.rs
littlebenlittle: I already restructured the code to avoid using channels before I saw this. I went back and was able to get it working with a `String` channel, though. Thank you for the suggestion!

My original intent with the channels was to keep the code structure and logic as similar as possible to the pion e2e tests. I might actually try to use channels again now that you've shown me the proper way to do it.

metaclips: Hack away😁

tests/basic.rs
metaclips: Has this been fixed? Seems outdated.

littlebenlittle: Yes, it was primarily because of not using tokio's `Mutex` implementation and instead using the one in `std`.

tests/basic.rs
```diff
@@ -0,0 +1,249 @@
...
+            server_port,
+            server_ready_tx,
+        ));
+        let timeout = Duration::from_secs(5);
+        let sleep = sleep(timeout);
+        tokio::pin!(sleep);
+        tokio::pin!(client);
+        tokio::pin!(server);
+        while !(server_complete && client_complete) {
+            tokio::select! {
```
littlebenlittle, line 214: There is a [runtime error](https://github.com/webrtc-rs/webrtc/pull/47/checks?check_run_id=1717692777#step:4:56).

```
thread 'e2e_mtu' panicked at 'unexpected task state', ...
```

littlebenlittle:
I'm opening this for review. The build is not passing, but most of logic has been filled in. I could a use a once-over by a more experienced rustacean when it comes to architecture and handling of async code.

rainliu:
@littlebenlittle, since DTLS has been moved to https://github.com/webrtc-rs/dtls repo, could you move this PR to that repo too?

littlebenlittle:
See https://github.com/webrtc-rs/dtls/pull/2

Issue #97, opened 2021-09-24
[WebRTC] double check simulcast example and play-from-disk-renegotiation example


rainliu:
Pion's simulcast example can receives two streams with rid = "f" and "h"
~~and  play-from-disk-renegotiation example  has video grid when click add video multiple times.~~

rainliu:
double check 
1). why simulcast example always receives one stream with rid="f" (full resolution)
2). why play-from-disk-renegotiation example can't add more grid

cloudwebrtc:
> 2). why play-from-disk-renegotiation example can't add more grid

fixed by https://github.com/webrtc-rs/webrtc/pull/98.

rainliu:
for 1). why simulcast example always receives one stream with rid="f" (full resolution)

according to chrome://webrtc-internals/,  webrtc-rs's bweforvideo
googAvailableReceiveBandwidth | 255280
googAvailableSendBandwidth | 300000

but pion has more than 1.5M bandwidth.

Most like due to RTCP feedback is not enabled in webrtc-rs, which cause BWE can't ramp up.

Will create another task to track RTCP feedback feature.








googActualEncBitrate | 75344
googAvailableReceiveBandwidth | 247541
googAvailableSendBandwidth | 300000
googBucketDelay | 0
googRetransmitBitrate | 0
googTargetEncBitrate | 300000
googTransmitBitrate

<br class="Apple-interchange-newline">

cloudwebrtc:
> 1). why simulcast example always receives one stream with rid="f" (full resolution)

My guess is that on_track is not triggered, because I have the same problem when testing the `reflect` example. When I try to add audio/video and echo at the same time, I can only trigger the first track.

```
Press ctlr-c to stop
Peer Connection State has changed: connected
Track rid = f has started
enter track loop f
Sending pli for stream with rid: f, ssrc: 1697313055
Sending pli for stream with rid: f, ssrc: 1697313055
Sending pli for stream with rid: f, ssrc: 1697313055
Sending pli for stream with rid: f, ssrc: 1697313055
Sending pli for stream with rid: f, ssrc: 1697313055
Sending pli for stream with rid: f, ssrc: 1697313055
```

cloudwebrtc:
When I wait for about 1 minute after connecting, h track will be triggered to display. Since then, there have been two tracks f|h.

cloudwebrtc:
[...]

PR #98, opened 2021-09-25
fix: Stop transceiver correctly.

cloudwebrtc:
fix some issues in https://github.com/webrtc-rs/webrtc/issues/97.

> 2). why play-from-disk-renegotiation example can't add more grid

After the repair, `play-from-disk-renegotiation` can `addVideo` correctly.


rainliu:
good catch! Thanks @cloudwebrtc  for the fix.

rainliu:
@cloudwebrtc, could you run "cargo fmt" and "cargo clippy"?

cloudwebrtc:
Ok I will do it right away. :)

PR #131, opened 2021-11-01
Fix underflow in operations length

k0nserv:
There was an underflow in the operations length when adding an operation
back on the queue. This happened because the length wasn't incremented,
but the operation was sent on the channel.


src/peer_connection/operation/mod.rs
```diff
@@ -53,9 +53,20 @@ impl Operations {
...
+
+        Ok(())
+    }
+
+    fn enqueue_inner(
+        op: Operation,
+        ops_tx: &Arc<mpsc::UnboundedSender<Operation>>,
+        length: &Arc<AtomicUsize>,
+    ) -> Result<()> {
+        length.fetch_add(1, Ordering::SeqCst);
```
k0nserv, line 67: I flipped the order of sending on the channel and increment the length too. Feel safer to use this ordering.

PR #142, opened 2021-12-14
Fix lifetime issue in `add_transceiver_from_track`

k0nserv:
See details in this [Stackoverflow Answer](https://stackoverflow.com/questions/68591843/async-fn-reports-hidden-type-for-impl-trait-captures-lifetime-that-does-not-a).

It seems like the compiler struggled to determine the appropriate lifetime to use for references when track was not also a reference. By making the lifetime explicit the compiler can be convinced.


src/peer_connection/mod.rs
```diff
@@ -1710,7 +1710,7 @@ impl RTCPeerConnection {
 
         let t = self
             .internal
-            .new_transceiver_from_track(direction, Arc::clone(track))
+            .new_transceiver_from_track(direction, Arc::clone(&track))
```
rainliu, line 1713: No need to clone it

k0nserv: Ah yes, right you are. Fixed

PR #143, opened 2021-12-15
Fix SDP attributes

k0nserv:
This fixes a bug where `set_local_description` would immediately trigger `on_negotiation_needed` even though a negotiation isn't needed. This happens because the [check](https://github.com/webrtc-rs/webrtc/blob/fc7e1ef9f9e7b5b6208abe808079bda23beb15c8/src/peer_connection/mod.rs#L471) for step 5.3.3 in `check_negotiation_needed` would always fail. 

`MediaDescription::attribute` did not allow distinguishing between an attribute that was missing and one that was present but had no value, such as `a=recvonly`.

This PR depends on this sibling PR in `webrtc-rs/sdp`: https://github.com/webrtc-rs/sdp/pull/9.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ repository = "https://github.com/webrtc-rs/webrtc"
 
 [dependencies]
 util = { package = "webrtc-util", version = "0.5.2" }
-sdp = "0.4.0"
+sdp = "0.5.0"
 mdns = { package = "webrtc-mdns", version = "0.4.1" }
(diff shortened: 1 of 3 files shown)


rainliu:
@k0nserv, I just bumped SDP to v0.5.0 to include your fix in sdp git repo. 

You can update sdp to v0.5.0 in Cargo.toml, in order to fix the Github Actions errors.

Thanks.

k0nserv:
Done

PR #146, opened 2021-12-18
Remove async from register_header_extension

jelmansouri:
This changed when transitioning to 0.3, was wondering if we can revert to this not being async, it doesn't seem necessary, but I might be mistaken.


rainliu:
Good catch!

PR #152, opened 2022-01-15
Fixes #151

### LINKED ISSUE #151: ReadRTCP doesn't work
Digging into how/why reading RTCP results in the data stream being corrupt.

I've

- Confirmed that in Pion, the code that I am trying to write works
- Isolated the issue to a simple repro
- Got suspicions that the issue is caused by a shared lock between RTP and RTCP, meaning the RTP only gets read when an RTCP packet comes in.

I'll dig deeper, and amend this issue as I do.

### Discussed in https://github.com/webrtc-rs/webrtc/discussions/150

<div type='discussions-op-text'>

<sup>Originally posted by **robashton** January 13, 2022</sup>
As mentioned over on the Interceptor repo, https://github.com/webrtc-rs/interceptor/issues/2 and indeed the subsequent pull request https://github.com/webrtc-rs/interceptor/pull/3, I'm wanting to get hold of the RTCP (Specifically, the sender reports) being sent by the web browser so I can do some A/V sync downstream of the rust.

The documentation for Pion suggests that we should use the receiver passed to us in OnTrack to perform this function, as this will run any interceptors and give you the final RTCP. (Hence how I discovered the issues there). ( https://github.com/pion/webrtc/blob/master/examples/rtcp-processing/main.go#L31 )

Indeed, on ingest - without the call to read_rtcp, those interceptors never get executed and presumably the rtcp being sent to us is just dropped on the floor (as far as I can tell). If I call read_rtcp, it breaks the actual *rtp* somehow, even if I do nothing with the results.

Sticking in the following code to the on_track code in the save-to-disk example means the interceptors get ran and I get some RTCP (including the sender reports in that println), but means the data stream breaks.

```rust
                    tokio::spawn(async move {
                        loop{
                            tokio::select! {
                                result = receiver.read_rtcp() => { 
                                    println!("RTCP {:?}". result);
                                }
                            }
                        }
                    });
``` 

wnat am I missing? How are we supposed to make sure the interceptors get executed on ingest and things like sender reports are made available to us?</div>

robashton:
A couple of simple changes to reduce lock-time on a shared channel, and reduce lock-time accessing shared tracks field. 

#151

### DIFF
--- a/src/rtp_transceiver/rtp_receiver/mod.rs
+++ b/src/rtp_transceiver/rtp_receiver/mod.rs
@@ -19,11 +19,11 @@ use crate::track::{TrackStream, TrackStreams};
 use interceptor::stream_info::RTPHeaderExtension;
 use interceptor::{Attributes, Interceptor};
 use std::sync::Arc;
-use tokio::sync::{mpsc, Mutex, Notify};
+use tokio::sync::{mpsc, Mutex, Notify, RwLock};
 
 pub struct RTPReceiverInternal {
     pub(crate) kind: RTPCodecType,
-    tracks: Mutex<Vec<TrackStreams>>,
+    tracks: RwLock<Vec<TrackStreams>>,
     closed_rx: Arc<Notify>,
     received_rx: Mutex<mpsc::Receiver<()>>,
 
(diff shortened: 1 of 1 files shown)


rainliu:
Concurrent programming is hard. Thank you for your hard work fixing this bug!

PR #154, opened 2022-02-07
Fix RTP extension id collision

k0nserv:
Fixes an issue where we would incorrectly map a given RTP extension ID to more than a single extension URI e.g. we'd map the id `3` to `urn:ietf:params:rtp-hdrext:sdes:mid` for an audio stream, but then map it `urn:3gpp:video-orientation` for a video stream negotiated later.

In this fix I've introduced a new field `propsed_header_extensions` which keeps track of the mappings from id to extension URI we have proposed but haven't yet negotiated completely. `header_extension` is no longer used to generate the `id` we assign a given extension.

In pseudo code here's what happens when we generate extensions.

```
# Params
# typ the stream type i.e. audio or video
# direction the stream direction 

if stream_type_negotiated(typ)
  // Use negotiated mappings from negotiated_header_extensions as before
else
  for local_extension in header_extensions
    if not_relevant(local_extension, typ, direction)
      continue
    end
    
    negotiated_extension = find_negotiated_extension(
      local_extension.uri, negotiated_header_extensions
    )
    
    # If we have already negotiated an extension id for this extension URI before use that.
    # This can happen if we negotiate e.g. an audio track and are now negotiating 
    # the first video track, in which case we want to use the previously agreed id.
    if negotiated_extension      
      add_to_output(negotiated_extension)

      continue
    end
    
    proposed_extension = find_proposed_extension(
      local_extension.uri, proposed_header_extensions
    )
    
    # If we have already proposed an extension id for the this extension we should re-use it.
    # This can happen e.g. if we proposed an extension for an audio stream but the peer didn't
    # support it. We should re-use the same id when proposing it for the video stream. 
    if proposed_extension
      add_to_output(proposed_extension)
      
      continue
    end
    
    # Lastly, if we haven't negotiated or proposed this extension yet we must
    # generate an id to propose. This ID cannot be an id we are already proposing or
    # have negotiated.
    id = find_free_id(
      local_extension, proposed_header_extensions, negotiated_header_extensions
    )
    
    if id
      proposed = add_proposed_header_extension(id, local_extension)
      
      add_to_output(proposed)
    else
      # We have exhausted all 15 possible IDs, log a warning and do nothing
    end
  end
end

```

And when updating from a remote answer or offer:

```
# Params
# extension The URI of the extension we are updating
# id The id suggested for the extension we are updating
# typ the stream type i.e. audio or video

# We only consider extension we've said we support.
for local_extension in header_extensions
  if local_extension.uri != extension
    continue
  end
  
  negotiated_extension = find_negotiated_extension(
      extension, negotiated_header_extensions
  )
  
  if negotiated_extension
    # We already have a negotiated extension ID for this extension
    if id == negotiated_extension.id
      # All good, the peer suggested the same extension id that we had.
      mark_active_for_current_type(negotiated_extension)
    else
      # Log an error, the peer changed the id of an extension that we have already negotiated
    end
  else
    # This is either the response to an extension id we proposed or a mapping the peer is 
    # suggesting. Accept it if we can
    prev_ext = negotiated_header_extensions[id]
    
    if prev_ext
      if prev_ext.uri != extension
        # This ID was already negotiated, log  warning
      end
    else
      # Accept the peer's suggestion
      mark_negotiated(id, extension, typ, negotiated_header_extensions)
    end
  end
  
  # Clear any previous proposal for this id
  proposed_header_extensions[id] = nil
end
```

### DIFF
--- a/src/api/media_engine/media_engine_test.rs
+++ b/src/api/media_engine/media_engine_test.rs
@@ -809,6 +809,8 @@ a=rtpmap:111 opus/48000/2
         .find(|ext| ext.uri == "urn:3gpp:video-orientation")
         .unwrap();
     assert_ne!(orientation.id, 1);
+    assert_ne!(orientation.id, 7);
+    assert_ne!(orientation.id, 5);
 
     Ok(())
 }
--- a/src/api/media_engine/media_engine_test.rs
+++ b/src/api/media_engine/media_engine_test.rs
@@ -721,3 +721,94 @@ async fn test_update_header_extenstion_to_cloned_media_engine() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_extension_id_collision() -> Result<()> {
+    let must_parse = |raw: &str| -> Result<SessionDescription> {
+        let mut reader = Cursor::new(raw.as_bytes());
+        Ok(SessionDescription::unmarshal(&mut reader)?)
+    };
+
+    const HEADER_EXTENSIONS: &str = "v=0
+o=- 4596489990601351948 2 IN IP4 127.0.0.1
+s=-
+t=0 0
+m=audio 9 UDP/TLS/RTP/SAVPF 111
+a=extmap:7 urn:ietf:params:rtp-hdrext:sdes:mid
+a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
+a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
+a=rtpmap:111 opus/48000/2
+";
+
+    let mut m = MediaEngine::default();
+    m.register_default_codecs()?;
+    for extension in ["urn:3gpp:video-orientation"] {
+        m.register_header_extension(
+            RTCRtpHeaderExtensionCapability {
+                uri: extension.to_owned(),
+            },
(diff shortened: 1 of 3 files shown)


k0nserv:
Hmm, about this I'm reading some more:

From https://datatracker.ietf.org/doc/html/rfc5285#section-5
> In addition, as noted above, the IDs used MUST be unique for each stream type for a given media, or for the session for session-level declarations


I think "stream type" means audio, video, or data here. So I guess the observed behaviour here is valid.

k0nserv:
I've [reproduced][0] this in Pion to rule out a bug in webrtc-rs. My understanding of how this should work is that this is a bug in both projects. Namely, extension IDs should be unique in a given session(even across different media types) when using BUNDLE.  




[0]: https://github.com/k0nserv/pion-webrtc/tree/feat/lack-of-extenion-id-collision

k0nserv:
In particular when using `BUNDLE` the extension id must be the same across all streams.

> When RTP header extensions [[RFC8285](https://www.rfc-editor.org/rfc/rfc8843.html#RFC8285)] are used in the context of this specification, the identifier used for a given extension MUST identify the same extension across all the bundled media descriptions.

_source:_ [RFC8843 Section 12](https://www.rfc-editor.org/rfc/rfc8843.html#section-12)

Issue #155, opened 2022-02-09
How to create an RTCSessionDescription structure?


ineiti:
Hi there,

I'm trying to get `webrtc-rs` to interact with the browser-webrtc available in `web_sys.RtcPeerConnection`. I started with the example here:

https://github.com/webrtc-rs/examples/blob/5a0e2861c66a45fca93aadf9e70a5b045b26dc9e/examples/data-channels/data-channels.rs

So I'm creating an `APIBuilder::new()[...].new_peer_connection(config)` that should create an `offer`, which will be sent to the `web_sys.RtcPeerConnection` implementation.

The `web_sys` then sends back an `answer` that I would like to feed to the `peer_connection`, and here is my problem: `web_sys` only gives me a `String`, not an `RTCSessionDescription` as needed by the `peer_connection`. And the `RTCSessionDescription` has a private field, so I cannot do:

```
peer_connection.set_local_description(RTCSessionDescription{
  sdp_type: RTCSdpType::Answer,
  sdp: answer_string,
  parsed: None
}
```

I looked around at the methods of `peer_connection` but couldn't find an easy way to feed the answer-string to it.

rainliu:
let desc_data = signal::decode(answer_string.as_str())?;
let answer = serde_json::from_str::<RTCSessionDescription>(&desc_data)?;

ineiti:
OK, I'll try that. But to me it would be more logical to have something like:
```
impl RTCSessionDescription {
  fn offer(sdp: &str) -> Self {
    Self {
      sdp_type: RTCSdpType::Offer,
      sdp,
      parsed: None
    }
  }
}
```

And the same for `pranswer`, `answer`, ... So it would be clearer:
```
let answer = RTCSessionDescription::answer(answer_string);
```

Would that make sense?

KallDrexx:
Adding my vote for this issue.  I had to add serde_json to my project just for this unfortunately, but it seems logical to me to be able to have a function convert a string SDP into a `RTCSessionDescription`.

sax:
Same here. I'm interacting with webrtc.rs from Elixir, where I have the SDP as a string. It seems like the only mechanism I have to make a peer connection work is to generate JSON (which I have to guess the structure of) just so that it can be decoded, but that really doesn't make sense.

rainliu:
I am not against adding new APIs to parse SDP string. 

PRs are welcome to address this issue.

sax:
[...]

PR #167, opened 2022-03-18
Add offer/answer/pranswer constructors for RTCSessionDescription

sax:
This is a proposed solution to #155


wdv4758h:
nice adding!
I was using something like:
```rust
let mut sdp = RTCSessionDescription::default();
sdp.sdp_type = RTCSdpType::Offer;
sdp.sdp = offer.to_string();
```

sax:
@wdv4758h Yeah I was going to try something like that, but realized that the private parsed field would not match up with the SDP, and that the SDP would not be evaluated for correctness without adding extra checks everywhere it was instantiated.

PR #170, opened 2022-03-21
always start sctp

melekes:
Fixes a bug where if you set
`disable_certificate_fingerprint_verification` to true via
`SettingEngine`, even though DTLS handshake is completed, SCTP does not
start. As you can see, this was due to the code returning `Ok(())`
without changing the state and starting the sctp.


melekes:
Refs https://github.com/paritytech/smoldot/pull/2135

rainliu:
Good catch!

PR #171, opened 2022-03-21
close DTLS conn upon err

melekes:
It looks like we're not closing the connection upon
ErrNoSRTPProtectionProfile / ErrNoRemoteCertificate errors.
Plus we're logging the wrong error when fingerprint is bad.


melekes:
Refs https://github.com/paritytech/smoldot/pull/2135

Issue #175, opened 2022-03-29
Cannot run example on Armv7 panicked at 'overflow when adding duration to instant'


frjol:
Based on the discussion from the issue in the example here https://github.com/webrtc-rs/examples/issues/5 

I have an issue running the example on a Raspberry Pi Zero 2 W armv7l.

It seems related this issue https://github.com/chronotope/chrono/issues/343

```
$ cat test.sdp | RUST_BACKTRACE=full ./play-from-disk-h264 --video output.h264
thread 'main' panicked at 'overflow when adding duration to instant', library/std/src/time.rs:550:31
stack backtrace:
   0:   0xab3708 - std::backtrace_rs::backtrace::libunwind::trace::h4bf3e5497ce7d0d1
                       at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:   0xab3708 - std::backtrace_rs::backtrace::trace_unsynchronized::h6b664ef1002045f8
                       at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:   0xab3708 - std::sys_common::backtrace::_print_fmt::h077a74dd0fd95079
                       at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:67:5
   3:   0xab3708 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::ha65a6b17248d4482
                       at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:46:22
   4:   0xad925c - core::fmt::write::h682b611af217d220
                       at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/fmt/mod.rs:1150:17
   5:   0xaaccec - std::io::Write::write_fmt::h8cc061d257db1323
                       at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/io/mod.rs:1667:15
   6:   0xab55d4 - std::sys_common::backtrace::_print::h8ea36d0a4c8c8101
                       at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:49:5
[...]

PR #176, opened 2022-03-29
Time overflow armv7 workaround

frjol:
Workaround for issue #175 

This is a workaround for the time overflow issue in chrono https://github.com/chronotope/chrono/issues/343 on armv7.

I am very open for discussion to find a more appropriate way to fix this.

Thanks!


src/peer_connection/certificate.rs
```diff
@@ -71,7 +72,10 @@ impl RTCCertificate {
         };
         params.key_pair = Some(key_pair);
 
-        let expires = params.not_after.into();
+        //let expires = params.not_after.into();
+        // Workaround for issue overflow when adding duration to instant on armv7
+        // https://github.com/webrtc-rs/examples/issues/5 https://github.com/chronotope/chrono/issues/343
+        let expires = SystemTime::now().add(Duration::from_secs(172800)); //60*60*48 or 2 days
```
rainliu, line 78: let’s use conditional compiler https://doc.rust-lang.org/reference/conditional-compilation.html  to handle arm special case, other arch still use line 75

rainliu:
Please also run cargo fmt and clippy

src/peer_connection/certificate.rs
```diff
@@ -71,7 +72,14 @@ impl RTCCertificate {
         };
         params.key_pair = Some(key_pair);
 
-        let expires = params.not_after.into();
+        let expires = if cfg!(armv7) {
```
rainliu, line 75: are you sure cfg!(armv7) is valid target_arch?

from https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch, only the following arch are supported:

"x86"
"x86_64"
"mips"
"powerpc"
"powerpc64"
"arm"
"aarch64"

rainliu: oh, just found this link: https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-with-host-tools

rainliu: Before accept this PR, I just want to confirm it: have you build and test this change in armv7 system?

rainliu: Thanks for verifying.

PR #180, opened 2022-04-06
Get stats from peer connection

sax:
This is obviously not finished, but I was hoping to get some feedback on a few points.

* I don't see a way of getting `bytes_received` or `bytes_sent` out of things that implement `Conn`. I was hoping that I was just missing some connecting thing.
* Do you have any hints on getting `base64_certificate` and `issuer_certificate_id` from an `RTCCertificate`?
* Serialization and casting... I was thinking that it might make sense to keep things close to the raw types (enums, Instants, etc), and delay transformations up until someone serialized types to something else like JSON. Any thoughts?
* `StatsReport`... in golang, this is a map. I was assuming this would be a struct that could be serialized, but wanted to ask if there was a preference.

I've left the existing commented-out golang functions for now, for references, but was going to delete them later.

### DIFF
--- a/src/ice_transport/mod.rs
+++ b/src/ice_transport/mod.rs
@@ -334,7 +334,7 @@ impl RTCIceTransport {
         let mut internal = self.internal.lock().await;
         if let Some(_conn) = internal.conn.take() {
             let collector = collector.clone();
-            let stats = ICETransportStats::new();
+            let stats = ICETransportStats::new("ice_transport".to_owned());
             // TODO: get bytes out of Conn.
             // bytes_received: conn.bytes_received,
             // bytes_sent: conn.bytes_sent,
--- a/src/ice_transport/mod.rs
+++ b/src/ice_transport/mod.rs
@@ -6,7 +6,6 @@ use std::sync::Arc;
 use ice::candidate::Candidate;
 use ice::state::ConnectionState;
 use tokio::sync::{mpsc, Mutex};
-use tokio::time::Instant;
 use util::Conn;
 
 use ice_candidate::RTCIceCandidate;
@@ -335,13 +334,10 @@ impl RTCIceTransport {
         let mut internal = self.internal.lock().await;
         if let Some(_conn) = internal.conn.take() {
             let collector = collector.clone();
-            let stats = ICETransportStats {
-                timestamp: Instant::now(),
-                id: "ice_transport".to_owned(),
-                // TODO: get bytes out of Conn.
-                // bytes_received: conn.bytes_received,
-                // bytes_sent: conn.bytes_sent,
-            };
+            let stats = ICETransportStats::new();
+            // TODO: get bytes out of Conn.
+            // bytes_received: conn.bytes_received,
+            // bytes_sent: conn.bytes_sent,
 
             let mut lock = collector.try_lock().unwrap();
             lock.push(Transport(stats));
--- a/src/ice_transport/mod.rs
+++ b/src/ice_transport/mod.rs
@@ -6,12 +6,14 @@ use std::sync::Arc;
 use ice::candidate::Candidate;
 use ice::state::ConnectionState;
 use tokio::sync::{mpsc, Mutex};
+use tokio::time::Instant;
 use util::Conn;
 
 use ice_candidate::RTCIceCandidate;
 use ice_candidate_pair::RTCIceCandidatePair;
 use ice_gatherer::RTCIceGatherer;
 use ice_role::RTCIceRole;
+use waitgroup::Worker;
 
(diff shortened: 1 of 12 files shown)


sax:
Also, obviously this would not be complete with some tests.

src/ice_transport/mod.rs
```diff
@@ -322,6 +326,25 @@ impl RTCIceTransport {
...
+        worker: Worker,
+    ) {
+        let mut internal = self.internal.lock().await;
+        if let Some(_conn) = internal.conn.take() {
+            let collector = collector.clone();
+            let stats = ICETransportStats::new("ice_transport".to_owned());
+            // TODO: get bytes out of Conn.
+            // bytes_received: conn.bytes_received,
+            // bytes_sent: conn.bytes_sent,
+
```
sax, line 341: I don't see any mechanism for getting `bytes_received` or `bytes_sent` from the conn.

src/sctp_transport/mod.rs
```diff
@@ -333,6 +338,51 @@ impl RTCSctpTransport {
...
+        collector: &Arc<Mutex<StatsCollector>>,
+        worker: Worker,
+        peer_connection_id: String,
+    ) {
+        let dtls_transport = self.transport();
+
+        // TODO: should this be collected?
+        dtls_transport
+            .collect_stats(collector, worker.clone())
+            .await;
```
sax, line 352: I don't see this specific metric being tracked in Pion, but since we're getting `dtls_transport` stats elsewhere it's easy to do.

src/sctp_transport/mod.rs
```diff
@@ -333,6 +338,51 @@ impl RTCSctpTransport {
...
+        )));
+
+        // conn
+        if let Some(_net_conn) = dtls_transport.conn().await {
+            let stats = ICETransportStats::new("sctp_transport".to_owned());
+            // TODO: get bytes out of Conn.
+            // bytes_received: conn.bytes_received,
+            // bytes_sent: conn.bytes_sent,
+            reports.push(SCTPTransport(stats));
+        }
```
sax, line 380: I don't see a way to get `bytes_received` or `bytes_sent`.

src/peer_connection/certificate.rs
```diff
@@ -168,7 +173,24 @@ impl RTCCertificate {
...
+        worker: Worker,
+    ) {
+        let fingerprints = self.get_fingerprints().unwrap();
+        if let Some(fingerprint) = fingerprints.into_iter().nth(0) {
+            let stats = CertificateStats::new(self, fingerprint);
+            let mut lock = collector.try_lock().unwrap();
+            lock.push(StatsReportType::CertificateStats(stats));
+
+            drop(worker);
+        }
```
sax, line 190: Pion gets `base64_certificate` and `issuer_certificate_id`, with some serialization. I don't see the actual source data for those in webrtc.rs.

src/stats/mod.rs
```diff
@@ -1 +1,300 @@
...
+            }
+        }
+    }
+}
+
+// TODO: should this be some form of String-indexed HashMap?
+#[derive(Debug)]
+pub struct StatsReport {
+    pub(crate) reports: Vec<StatsReportType>,
+}
```
sax, line 55: I think that the Pion `StatsReport` is actually a type alias to something like a Rust `std::collection::HashMap`. I thought that it might make sense to have a struct, and then convert types when serializing to JSON (which I'm not completely sure how to do).

Does that make sense? Maybe rather than reports being a `Vec<StatsReportType>` it should be a `HashMap<String, StatsReportType>`?

sax (commented):
I think this review covers the main questions I have to fill out the remaining stats fields that Pion generates.

src/sctp_transport/mod.rs
sax: I see that `DTLSConn` in the `webrtc-dtls` crate has a private `conn`. Perhaps if that's made public and released, `bytes_received` and `bytes_sent` can be gotten out of that?

Another thought is `fn bytes_received() -> usize` and `fn bytes_sent() -> usize` can be added to the required implementation of `Conn` (assuming it's possible for all implementers to track). That seems like it would require deeper changes, and perhaps multiple sub-crates to be released before it could be added to the stats code.

sax:
@rainliu would it be possible for you to take a look at this? Sorry to tag you unnecessarily if there's a better person or method for asking for reviews.

rainliu:
> @rainliu would it be possible for you to take a look at this? Sorry to tag you unnecessarily if there's a better person or method for asking for reviews.

Sure, will take a look soon.

PR #186, opened 2022-04-30
Support completely unreliable datachannels

connorcarpenter:
It appears that DataChannel parameters: `id` `maxRetransmits` & `maxPacketLifetime` were not translated completely from https://github.com/pion/webrtc/blob/master/datachannel.go#L26

These parameters are able to have a value of `nil` in Pion, which allows for setup of datachannels which are completely unordered & unreliable. There's a big difference between an `id` or `maxRetransmits` of `nil` vs `0`, which **webrtc-rs** currently conflates.

This will allow interop with https://github.com/triplehex/webrtc-unreliable .

This PR requires the usage of the following PRs to function correctly:
https://github.com/webrtc-rs/sctp/pull/10
https://github.com/webrtc-rs/data/pull/5

### DIFF
--- a/src/data_channel/data_channel_parameters.rs
+++ b/src/data_channel/data_channel_parameters.rs
@@ -5,9 +5,9 @@ use serde::{Deserialize, Serialize};
 pub struct DataChannelParameters {
     pub label: String,
     pub protocol: String,
-    pub id: u16,
+    pub id: Option<u16>,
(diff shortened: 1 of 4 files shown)


rainliu:
@connorcarpenter, similar issue in https://github.com/webrtc-rs/data/pull/5.

Issue #144, opened 2021-12-16
Firefox SDP parsing fails due to missing C line on media level


k0nserv:
Hey,

I noticed that, at least in my [modified reflect](https://github.com/webrtc-rs/examples/pull/7) example, Firefox throws a DOM exception when applying the answer SDP.

<details>
<summary>Example SDP</summary>

```
v=0
o=- 4985367504486208344 356629000 IN IP4 0.0.0.0
s=-
t=0 0
a=fingerprint:sha-256 D5:67:46:98:F8:D8:05:4B:D5:42:4E:FD:04:C7:03:93:B9:90:C8:AD:A2:82:3E:AE:24:E3:4A:B8:54:2D:76:31
a=group:BUNDLE 1 2
m=audio 0 UDP/TLS/RTP/SAVPF 0
a=candidate:167090039 1 udp 2130706431 :: 50760 typ host
a=candidate:167090039 2 udp 2130706431 :: 50760 typ host
a=candidate:3498147984 1 udp 2130706431 192.168.1.76 59816 typ host
a=candidate:3498147984 2 udp 2130706431 192.168.1.76 59816 typ host
a=candidate:3122685691 1 udp 2130706431 192.168.1.33 57564 typ host
a=candidate:3122685691 2 udp 2130706431 192.168.1.33 57564 typ host
a=end-of-candidates
m=video 9 UDP/TLS/RTP/SAVPF 120
c=IN IP4 0.0.0.0
a=setup:active
a=mid:1
a=ice-ufrag:nzCUobTruNCKeWEl
a=ice-pwd:pSowjNuaNylpeKBFgTsFUNwQRdrzDQxe
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:120 VP8/90000
a=fmtp:120 max-fs=12288;max-fr=60
a=rtcp-fb:120 nack
a=rtcp-fb:120 nack pli
a=rtcp-fb:120 ccm fir
a=rtcp-fb:120 goog-remb
a=rtcp-fb:120 transport-cc
a=extmap:7 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=recvonly
m=application 9 UDP/DTLS/SCTP webrtc-datachannel
c=IN IP4 0.0.0.0
a=setup:active
a=mid:2
a=sendrecv
a=sctp-port:5000
a=ice-ufrag:nzCUobTruNCKeWEl
[...]

PR #187, opened 2022-05-09
Fix SDP for rejected tracks to conform to RFC

k0nserv:
We were not including the connection line(`c=`) when rejecting track,
but the specification requires this. Firefox would reject SDPs without
this.


k0nserv:
This fixes https://github.com/webrtc-rs/webrtc/issues/144 and is ported from Pion's fix for the same issue.

PR #201, opened 2022-06-16
Handle inactive transceivers more correctly

k0nserv:
Instead of stopping transceivers when a media section transitions to
inactive we keep them active, then if they become active again we resume receiving.

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -1346,7 +1346,7 @@ impl RTCPeerConnection {
                                         t.set_mid(mid_value.to_owned()).await?;
                                     }
 
-                                    t.process_new_direection(previous_direction).await?;
+                                    t.process_new_direction(previous_direction).await?;
                                 } else {
                                     let receiver = Arc::new(RTCRtpReceiver::new(
                                         self.internal.setting_engine.get_receive_mtu(),
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -1323,29 +1323,18 @@ impl RTCPeerConnection {
 
                                 if let Some(t) = t {
                                     let previous_direction = t.direction();
-                                    if direction == RTCRtpTransceiverDirection::Recvonly {
-                                        if t.direction() == RTCRtpTransceiverDirection::Sendrecv {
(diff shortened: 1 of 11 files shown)


src/rtp_transceiver/rtp_receiver/mod.rs
```diff
@@ -37,28 +137,25 @@ pub struct RTPReceiverInternal {
...
-                _ = received_rx.recv() =>{
-                    // Drop into the below code and don't hold this lock any longer
-                }
-                _ = self.closed_rx.notified() => {
-                    return Err(Error::ErrClosedPipe);
-                }
-            }
-        }
+        let mut state_watch_rx = self.state_rx.clone();
+        // Ensure we are running or paused. When paused we still recevie RTCP even if RTP traffic
```
melekes, line 141:
```suggestion
        // Ensure we are running or paused. When paused we still receive RTCP even if RTP traffic
```

src/rtp_transceiver/rtp_receiver/mod.rs
```diff
@@ -18,14 +18,114 @@ use crate::track::{TrackStream, TrackStreams};
...
+                value
+            ),
+        }
+    }
+}
+
+impl fmt::Display for State {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            State::Initial => write!(f, "Initial"),
```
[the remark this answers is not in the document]
k0nserv: Nope, it's `fmt::Display` which cannot be derived.

src/peer_connection/mod.rs
```diff
@@ -1325,19 +1322,20 @@ impl RTCPeerConnection {
                                 };
```
algesten, line 1322: the `if let Some(t) = ...` unpacking here seems a bit strange when we just repack it `Some(t)`. Maybe use `or_else`?

algesten (commented):
LGTM!

src/peer_connection/mod.rs
k0nserv: I've been thinking the same thing, but it's not really in scope for this PR to change it. There's room for a PR that tweaks this pattern across the codebase, personally I think match and an early return to keep nesting levels in check is a good choice

PR #204, opened 2022-06-21
[peer_connection] allow persistent certificates

### LINKED ISSUE #168: [DTLS] allow providing custom certificates
Hey! I'm building a webrtc server for Parity's light client called smoldot (https://github.com/paritytech/smoldot). The server's identity is fixed, meaning server's certificate does not change.

In the current API, one can provide a certificate, BUT there seems to be no way to actually "load" existing certificate (from pem) because

```
pub struct RTCCertificate {
    pub(crate) certificate: dtls::crypto::Certificate,
    pub(crate) stats_id: String,
    pub(crate) x509_cert: rcgen::Certificate,
    pub(crate) expires: SystemTime,
}
```

rcgen::Certificate can only be used for generating new certificates (no way to construct an existing one). Am I missing something? Thanks

Maybe we can replace `x509_cert` `rcgen::Certificate` with `pem` string? I don't think it's used for anything other than serialising certificate to pem. See https://github.com/melekes/webrtc/commit/04acecb68ccb813ca7caf4e30fc3affb6314e834 What do you think?

melekes:
Closes #168 

This PR adds `RTCCertificate::from_existing` method, which constructs `RTCCertificate` from an existing  DTLS certificate. An existing certificate might be needed in cases like [this](https://github.com/libp2p/rust-libp2p/pull/2622) where you need DTLS identity to be fixed for some period of time (whole duration of the certificate or some part of it).

### DIFF
--- a/src/peer_connection/certificate.rs
+++ b/src/peer_connection/certificate.rs
@@ -107,10 +107,15 @@ impl RTCCertificate {
         })
     }
 
-    /// new constructs a `RTCCertificate` from an existing certificate. Use this method when you
-    /// have a persistent certificate (i.e. you don't want to generate a new one for each DTLS
-    /// connection).
-    pub fn new(certificate: dtls::crypto::Certificate, pem: &str, expires: SystemTime) -> Self {
+    /// Constructs a `RTCCertificate` from an existing certificate.
+    ///
+    /// Use this method when you have a persistent certificate (i.e. you don't want to generate a
+    /// new one for each DTLS connection).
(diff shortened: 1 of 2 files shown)


k0nserv (approved):
LGTM

melekes:
NOTE: this method does not exist in Go's Pion, but I believe adding it is reasonable. 

@rainliu any thoughts?

melekes:
Merged for now. We can always revert if needed since it's not breaking anything.

PR #205, opened 2022-06-23
data_channel: handle stream EOF

melekes:
Refs https://github.com/webrtc-rs/sctp/pull/14


src/data_channel/mod.rs
```diff
@@ -305,18 +305,31 @@ impl RTCDataChannel {
                 _ = notify_rx.notified() => break,
                 result = data_channel.read_data_channel(&mut buffer) => {
                     match result{
+                        Ok((0, _)) =>
```
k0nserv, line 308: Hmmmm 🤔 This gets me thinking: is there a convention in the project to return EOF as an error that we are violating by changing it to `Ok(0)`?

k0nserv (approved):

PR #208, opened 2022-06-28
update deps + loosen some requirements

melekes:
rules:
- for crates that are below v1.0 -> specify the precise version
- If the code uses a feature that was added for example in X 0.3.17,
  then you should specify 0.3.17, which actually means "0.3.y where y >=
  17"
- for crates the are above or equal v1.0 -> specify only major version
  if the crate's API is minimal and won't change between minor versions
    OR specify major&minor versions otherwise

tested with https://github.com/taiki-e/cargo-minimal-versions


melekes:
This will have to wait until new versions of "webrtc-*" are released.

k0nserv (approved):
Very nice 🎉

melekes:
NOTE: compilation with minimal versions still fails because it needs newer versions of webrtc-rs crates (I've updated them all, but we haven't released new versions yet except for util crate), but I'd prefer to merge it anyway.

PR #209, opened 2022-06-28
MID generator feature

algesten:


src/api/setting_engine/mod.rs
```diff
@@ -74,6 +74,7 @@ pub struct SettingEngine {
     pub(crate) disable_media_engine_copy: bool,
     pub(crate) srtp_protection_profiles: Vec<SrtpProtectionProfile>,
     pub(crate) receive_mtu: usize,
+    pub(crate) mid_generator: Option<Arc<Box<dyn Fn() -> String + Send + Sync>>>,
```
k0nserv, line 77: I don't think we need `Box` here? `Arc<dyn Fn() -> Strign + Send + Sync>` ought to be enough.

src/api/setting_engine/mod.rs
```diff
@@ -294,4 +295,11 @@ impl SettingEngine {
     pub fn set_receive_mtu(&mut self, receive_mtu: usize) {
         self.receive_mtu = receive_mtu;
     }
+
+    /// Sets a callback used to generate mid for transceivers create by this side of the RTCPeerconnection.
+    /// By having separate "naming schemes" for mids generated by either side of a connection, it's
+    /// possible to reduce complexity when handling SDP offers/answers clashing.
+    pub fn set_mid_generator(&mut self, f: impl Fn() -> String + Send + Sync + 'static) {
```
k0nserv, line 302: I think the callback should take an isize argument which is the value of greater_mid value at the time of calling it

lolgesten: sure

src/api/setting_engine/mod.rs
k0nserv: I think let's leave it as sync until someone complains?

algesten: alright

algesten: done

src/peer_connection/mod.rs
```diff
@@ -765,8 +765,28 @@ impl RTCPeerConnection {
...
+
+                        // If it's possible to parse the returned mid as numeric, we will update the greater_mid field.
+                        let numeric_mid = match mid.parse::<isize>() {
+                            Ok(n) => n,
+                            Err(_) => continue,
+                        };
+
+                        t.set_mid(mid).await?;
+
+                        if numeric_mid > self.internal.greater_mid.load(Ordering::SeqCst) {
```
k0nserv, line 781: This isn't done in the other case why do we need to do it here 🤔

lolgesten: See above for incoming mids

k0nserv (approved):

src/api/setting_engine/mod.rs
```diff
@@ -294,4 +295,18 @@ impl SettingEngine {
     pub fn set_receive_mtu(&mut self, receive_mtu: usize) {
         self.receive_mtu = receive_mtu;
     }
+
+    /// Sets a callback used to generate mid for transceivers create by this side of the RTCPeerconnection.
```
k0nserv, line 299: `create` –> `created`

PR #210, opened 2022-06-29
Rework transceiver direction support further

k0nserv:
The previous change to transceiver direction behaviour(#201) was a step in the
right direction(lol) but it wasn't correct. In particular we were making
changes when applying a remote offer that should happen when applying a
remote or local answer. The state machine for receivers was also
incorrect.

Here's the full state machine for RTP Receivers

image

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -1241,42 +1241,42 @@ impl RTCPeerConnection {
         let remote_description = self.remote_description().await;
         let mut local_transceivers = self.get_transceivers().await;
         if we_answer {
-            if let Some(remote_desc) = remote_description {
-                if let Some(parsed) = &remote_desc.parsed {
-                    // WebRTC Spec 1.0 https://www.w3.org/TR/webrtc/
-                    // Section 4.4.1.5
-                    for media in &parsed.media_descriptions {
-                        let mid_value = match get_mid_value(media) {
-                            Some(mid) => mid,
-                            None => continue,
-                        };
+            if let Some(parsed) = desc.parsed {
+                // WebRTC Spec 1.0 https://www.w3.org/TR/webrtc/
+                // Section 4.4.1.5
+                for media in &parsed.media_descriptions {
+                    let mid_value = match get_mid_value(media) {
+                        Some(mid) => mid,
+                        None => continue,
+                    };
 
-                        if media.media_name.media == MEDIA_SECTION_APPLICATION {
-                            continue;
-                        }
-
-                        let kind = RTPCodecType::from(media.media_name.media.as_str());
-                        let direction = get_peer_direction(media);
-                        if kind == RTPCodecType::Unspecified
-                            || direction == RTCRtpTransceiverDirection::Unspecified
-                        {
-                            continue;
-                        }
-
-                        let t = match find_by_mid(mid_value, &mut local_transceivers).await {
-                            Some(t) => t,
-                            None => continue,
-                        };
-                        let previous_direction = t.current_direction();
-                        // 4.9.1.7.3 applying a local answer or pranswer
-                        // Set transceiver.[[CurrentDirection]] and transceiver.[[FiredDirection]] to direction.
-
-                        // TODO: Also set FiredDirection here.
-                        t.set_current_direction(direction);
-                        t.process_new_current_direction(previous_direction).await?;
+                    if media.media_name.media == MEDIA_SECTION_APPLICATION {
+                        continue;
                     }
-                }
 
+                    let kind = RTPCodecType::from(media.media_name.media.as_str());
+                    let direction = get_peer_direction(media);
+                    if kind == RTPCodecType::Unspecified
+                        || direction == RTCRtpTransceiverDirection::Unspecified
+                    {
+                        continue;
+                    }
+
+                    let t = match find_by_mid(mid_value, &mut local_transceivers).await {
+                        Some(t) => t,
+                        None => continue,
+                    };
+                    let previous_direction = t.current_direction();
+                    // 4.9.1.7.3 applying a local answer or pranswer
+                    // Set transceiver.[[CurrentDirection]] and transceiver.[[FiredDirection]] to direction.
+
+                    // TODO: Also set FiredDirection here.
+                    t.set_current_direction(direction);
+                    t.process_new_current_direction(previous_direction).await?;
+                }
+            }
+
+            if let Some(remote_desc) = remote_description {
                 self.start_rtp_senders().await?;
 
                 let pci = Arc::clone(&self.internal);
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -1426,14 +1426,6 @@ impl RTCPeerConnection {
                             continue;
                         }
 
-                        let kind = RTPCodecType::from(media.media_name.media.as_str());
-                        let direction = get_peer_direction(media);
-                        if kind == RTPCodecType::Unspecified
-                            || direction == RTCRtpTransceiverDirection::Unspecified
-                        {
-                            continue;
-                        }
-
                         if let Some(t) = find_by_mid(mid_value, &mut local_transceivers).await {
                             let previous_direction = t.direction();
 
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -1324,85 +1324,84 @@ impl RTCPeerConnection {
             let we_offer = desc.sdp_type == RTCSdpType::Answer;
 
             if !we_offer && !detected_plan_b {
-                if let Some(remote_desc) = &remote_description {
-                    if let Some(parsed) = &remote_desc.parsed {
-                        for media in &parsed.media_descriptions {
-                            if let Some(mid_value) = get_mid_value(media) {
-                                if mid_value.is_empty() {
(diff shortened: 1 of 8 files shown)


k0nserv:
Notes on these changes:


## Create offer

Procedure described in [RFC8829 section 4.1.8](https://datatracker.ietf.org/doc/html/rfc8829#section-4.1.8) which in turn 
references [section 5.2.1](https://datatracker.ietf.org/doc/html/rfc8829#section-5.2.1) for initial offers and 
[section 5.2.2](https://datatracker.ietf.org/doc/html/rfc8829#section-5.2.2) for subsequent offers.

For initial offers the direction of any `m=` lines should be:

> A direction attribute that is the same as that of the associated transceiver.

[Section 4.2.3](https://datatracker.ietf.org/doc/html/rfc8829#section-4.2.3)

> When creating offers, the transceiver direction is directly reflected
> in the output, even for re-offers. 

## Create answer

Procedure described in [RFC8829 section 4.1.9](https://datatracker.ietf.org/doc/html/rfc8829#section-4.1.9) which references
[section 5.3](https://datatracker.ietf.org/doc/html/rfc8829#section-5.3). Similarly to offer there's a split between
initial answers([section 5.3.1](https://datatracker.ietf.org/doc/html/rfc8829#section-5.3.1)) and subsequent 
answers([section 5.3.1](https://datatracker.ietf.org/doc/html/rfc8829#section-5.3.2)).

When it comes to generating `m=` lines RFC8829 points to [RFC3264 Section 6.1](https://datatracker.ietf.org/doc/html/rfc3264#section-6.1).

About direction that section says:

> If a stream is offered as sendonly, the corresponding stream MUST be
> marked as recvonly or inactive in the answer.  If a media stream is
> listed as recvonly in the offer, the answer MUST be marked as
> sendonly or inactive in the answer.  If an offered media stream is
> listed as sendrecv (or if there is no direction attribute at the
> media or session level, in which case the stream is sendrecv by
> default), the corresponding stream in the answer MAY be marked as
> sendonly, recvonly, sendrecv, or inactive.  If an offered media
> stream is listed as inactive, it MUST be marked as inactive in the
> answer.

## Updating a Transceivers Direction

The direction can be set manually and indicates the desired, but not necessarily used, direction.

[RFC8829 section 4.2.3](https://datatracker.ietf.org/doc/html/rfc8829#section-4.2.3)

> The setDirection method sets the direction of a transceiver, which
> affects the direction property of the associated "m=" section on
> future calls to createOffer and createAnswer.

The actual direction in used is stored in `currentDirection` 

[RFC8829 section 4.2.5](https://datatracker.ietf.org/doc/html/rfc8829#section-4.2.3)

> The currentDirection property indicates the last negotiated direction
> for the transceiver's associated "m=" section.  More specifically, it
> indicates the direction attribute [RFC3264] of the associated "m="
> section in the last applied answer (including provisional answers),
> with "send" and "recv" directions reversed if it was a remote answer.
> For example, if the direction attribute for the associated "m="
> section in a remote answer is "recvonly", currentDirection is set to
> "sendonly".

> If an answer that references this transceiver has not yet been
> applied or if the transceiver is stopped, currentDirection is set to
> "null".


Both `direction` and `currentDirection` are updated when applying a remote answer i.e. via `set_remote_description`. This is step 4.9.2.13.2 of [the set description process](https://www.w3.org/TR/webrtc/#set-description).

`currentDirection` is updated when setting applying local answer i.e. via `set_local_description`. This is step 4.9.1.7.3 of [the set description process](https://www.w3.org/TR/webrtc/#set-description).

When `currentDirection` changes we potentially pause/unpause rtp receivers.

src/peer_connection/mod.rs
```diff
@@ -1288,87 +1324,131 @@ impl RTCPeerConnection {
...
+                        }
+                        let kind = RTPCodecType::from(media.media_name.media.as_str());
+                        let direction = get_peer_direction(media);
+                        if kind == RTPCodecType::Unspecified
+                            || direction == RTCRtpTransceiverDirection::Unspecified
+                        {
+                            continue;
+                        }
+
+                        let kind = RTPCodecType::from(media.media_name.media.as_str());
```
k0nserv, line 1429: Woops, this is duped

src/rtp_transceiver/mod.rs
```diff
@@ -304,28 +310,61 @@ impl RTCRtpTransceiver {
...
     }
 
-    /// direction returns the RTPTransceiver's current direction
+    /// direction returns the RTPTransceiver's desired direction.
     pub fn direction(&self) -> RTCRtpTransceiverDirection {
         self.direction.load(Ordering::SeqCst).into()
     }
 
     pub(crate) fn set_direction(&self, d: RTCRtpTransceiverDirection) {
+        let previous: RTCRtpTransceiverDirection = self.direction.load(Ordering::SeqCst).into();
```
melekes, line 319: note: you can use https://rust.group/core/sync/atomic/struct.AtomicU8.html#method.swap to get previous value and set new one in one hop

src/rtp_transceiver/rtp_receiver/mod.rs
```diff
@@ -27,19 +27,29 @@ use tokio::sync::{watch, Mutex, RwLock};
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 #[repr(u8)]
 pub enum State {
-    Initial = 0,
-    Running = 1,
-    Paused = 2,
-    Closed = 3,
+    /// We haven't started yet.
```
melekes, line 30:
```suggestion
    /// The transceiver hasn't started yet.
```

src/rtp_transceiver/rtp_receiver/mod.rs
```diff
@@ -27,19 +27,29 @@ use tokio::sync::{watch, Mutex, RwLock};
...
+    Unstarted = 0,
+    /// We haven't started yet and additionally we've been paused.
```
melekes, line 32:
```suggestion
    /// The transceiver hasn't started yet and additionally been paused.
```

src/rtp_transceiver/rtp_receiver/mod.rs
```diff
@@ -27,19 +27,29 @@ use tokio::sync::{watch, Mutex, RwLock};
...
+    UnstartedPaused = 1,
+
+    /// We have started and are running.
```
melekes, line 35:
```suggestion
    /// The transceiver has started and is running.
```

src/rtp_transceiver/rtp_receiver/mod.rs
```diff
@@ -27,19 +27,29 @@ use tokio::sync::{watch, Mutex, RwLock};
...
+    Started = 2,
+
+    /// We have been paused after starting.
```
melekes, line 38:
```suggestion
    /// The transceiver has been paused after starting.
```

src/rtp_transceiver/rtp_receiver/mod.rs
```diff
@@ -27,19 +27,29 @@ use tokio::sync::{watch, Mutex, RwLock};
...
+    Paused = 3,
+
+    /// We have been stopped.
```
melekes, line 41:
```suggestion
    /// The transceiver has been stopped.
```

src/peer_connection/mod.rs
```diff
@@ -1217,7 +1239,43 @@ impl RTCPeerConnection {
...
         let we_answer = desc.sdp_type == RTCSdpType::Answer;
         let remote_description = self.remote_description().await;
+        let mut local_transceivers = self.get_transceivers().await;
         if we_answer {
+            if let Some(parsed) = desc.parsed {
+                // WebRTC Spec 1.0 https://www.w3.org/TR/webrtc/
+                // Section 4.4.1.5
+                for media in &parsed.media_descriptions {
+                    let mid_value = match get_mid_value(media) {
+                        Some(mid) => mid,
```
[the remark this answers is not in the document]
k0nserv: It will fail silently later(`find_by_mid` will return None). I'm not quite sure if this should be an error here, but for now silent failure seems good enough.

src/rtp_transceiver/mod.rs
k0nserv: Good catch, this was supposed to be throw away code, but I think having the logging might be useful so I'll use `swap` instead.

src/rtp_transceiver/rtp_receiver/mod.rs
k0nserv: This state isn't transceiver specific, it's receiver specific. Further, I could see us using this struct for RtpSender too down the line

k0nserv: I'm thinking of "we" as which ever object is using this state enum to track its current state. Happy to hear other suggestions or perhaps I can reword it to no longer include the word "we".

src/peer_connection/mod.rs
```diff
@@ -1262,6 +1257,17 @@ impl RTCPeerConnection {
                         continue;
                     }
 
+                    let mid_value = match get_mid_value(media) {
+                        Some(mid) => {
```
melekes, line 1261:
```suggestion
                        Some(mid) if !mid.empty() => { mid }
                        _ => continue
```

k0nserv: I feel like there was a reason I didn't do this, but I can't remember what it was anymore :D

melekes (approved):

PR #212, opened 2022-07-04
Public RTCRtpTransceiver::set_direction

algesten:
This should be public API, and this PR makes it so.


k0nserv:
Could we explore a test for this too i.e. when setting direction via `set_direction` the Peer Connection's negotiation needed fires?

algesten:
@k0nserv done

src/rtp_transceiver/rtp_transceiver_test.rs
```diff
@@ -240,7 +243,7 @@ async fn test_rtp_transceiver_direction_change() -> Result<()> {
     answer_pc.set_local_description(answer.clone()).await?;
     offer_pc.set_remote_description(answer).await?;
 
-    offer_transceiver.set_direction(RTCRtpTransceiverDirection::Inactive);
+    offer_transceiver.set_direction_internal(RTCRtpTransceiverDirection::Inactive);
```
k0nserv, line 246: This should use the public version

k0nserv (approved):
LGTM

PR #217, opened 2022-07-07
Correct msid handling for RtpSender

k0nserv:
The previous logic would put negotiation into an endless loop after
running `remove_track` due to incorrectly implementing the negotiation
check and handling of stream ids.

Upon calling `remove_track` the following steps were carried out:

1. Find the transceiver the sender passed to `remove_track` is associated with
2. Stop the sender and set its track to `None`.

There were three problems with this:

1. The `remove_track` implementation is mostly correct, but it didn't always update the transceivers direction.
2. When checking if negotiation is needed the logic wasn't quite correct.
3. When offering the offered msid attribute didn't comply with the specification, in particular the msid line(s) should stay the same regardless of changes to the track or the transceivers direction. 

In total this meant that when removing a track, at least some times, we'd end up in an endless negotiation loop where the check if negotiation was needed was always returning true, even after negotiating with the peer.

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -1767,11 +1767,11 @@ impl RTCPeerConnection {
         }
 
         if let Some(t) = transceiver {
-            if sender.stop().await.is_ok() && t.set_sending_track(None).await.is_ok() {
-                t.set_direction_internal(RTCRtpTransceiverDirection::from_send_recv(
-                    false,
-                    t.direction().has_recv(),
-                ));
+            let sender_result = sender.stop().await;
+            // This also updates direction
+            let sending_track_result = t.set_sending_track(None).await;
+
+            if sender_result.is_ok() && sending_track_result.is_ok() {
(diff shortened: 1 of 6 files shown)


src/peer_connection/mod.rs
```diff
@@ -447,27 +447,48 @@ impl RTCPeerConnection {
...
-                let m = get_by_mid(t.mid().await.as_str(), local_desc);
+                let mid = t.mid().await;
+                let m = get_by_mid(&mid, local_desc);
                 // Step 5.2
                 if !t.stopped.load(Ordering::SeqCst) && m.is_none() {
                     return true;
                 }
                 if !t.stopped.load(Ordering::SeqCst) {
                     if let Some(m) = m {
                         // Step 5.3.1
```
k0nserv, line 458: See https://www.w3.org/TR/webrtc/#ref-for-dfn-check-if-negotiation-is-needed-1 for the steps outlined by the specification

PR #14, opened 2020-11-19
RTCP: Add bytes crate support

metaclips:

PR #219, opened 2022-07-12
update call to DataChannel::accept as per data pr #14

stuqdog:
Allows for acceptance of already created data channels. Duplicates logic already existing in pion at https://github.com/pion/webrtc/blob/master/sctptransport.go#L162. 

Note that this is a sibling(niece?) pr of https://github.com/webrtc-rs/data/pull/14; it relies on that pr and won't compile until that pr is merged.

### DIFF
--- a/src/sctp_transport/mod.rs
+++ b/src/sctp_transport/mod.rs
@@ -212,12 +212,22 @@ impl RTCSctpTransport {
     }
 
     async fn accept_data_channels(param: AcceptDataChannelParams) {
+        let dcs = param.data_channels.lock().await;
+        let mut existing_data_channels = Vec::new();
+        for dc in dcs.iter() {
+            if let Some(dc) = dc.data_channel.lock().await.clone() {
+                let dc2 = dc.as_ref().clone();
+                existing_data_channels.push(dc2);
+            }
+        }
+
         loop {
             let dc = tokio::select! {
                 _ = param.notify_rx.notified() => break,
                 result = DataChannel::accept(
                     &param.sctp_association,
                     data::data_channel::Config::default(),
+                    existing_data_channels,
                 ) => {
                     match result {
                         Ok(dc) => dc,


src/sctp_transport/mod.rs
```diff
@@ -212,12 +212,22 @@ impl RTCSctpTransport {
...
+            }
+        }
+
         loop {
             let dc = tokio::select! {
                 _ = param.notify_rx.notified() => break,
                 result = DataChannel::accept(
                     &param.sctp_association,
                     data::data_channel::Config::default(),
+                    existing_data_channels.to_owned(),
```
melekes, line 230: why do we need `to_owned`?

src/sctp_transport/mod.rs
```diff
@@ -212,12 +212,22 @@ impl RTCSctpTransport {
     }
 
     async fn accept_data_channels(param: AcceptDataChannelParams) {
+        let dcs = param.data_channels.lock().await;
+        let mut existing_data_channels = Vec::new();
+        for dc in dcs.iter() {
+            if let Some(dc) = dc.data_channel.lock().await.clone() {
+                let dc2 = dc.as_ref().clone();
```
melekes, line 219: why the second `clone`? if it's an already cloned object

src/sctp_transport/mod.rs
```diff
@@ -212,12 +212,22 @@ impl RTCSctpTransport {
     }
 
     async fn accept_data_channels(param: AcceptDataChannelParams) {
+        let dcs = param.data_channels.lock().await;
+        let mut existing_data_channels = Vec::new();
+        for dc in dcs.iter() {
```
melekes, line 217: Also one idea I just had is to modify `DataChannel::accept` to accept a reference to a list of existing data channels. That way we won't need to clone all of them, just the one with matching identifier.

src/sctp_transport/mod.rs
stuqdog: (sorry for slow response!) 

without `to_owned()` we get an error of the `existing_data_channels` value moving in a previous loop iteration. I'll readily admit that I am no Rust expert though, so if there's a better solution to this then I'm all ears :)

src/sctp_transport/mod.rs
melekes: Opened https://github.com/webrtc-rs/data/pull/15

PR #220, opened 2022-07-14
Make stats types pub so they can be used directly

k0nserv:
Previously the only way to access the values was serialising them and
then relying on stringly typed access.

### DIFF
--- a/src/stats/mod.rs
+++ b/src/stats/mod.rs
@@ -104,7 +104,7 @@ impl Serialize for StatsReportType {
 
 #[derive(Debug)]
 pub struct StatsReport {
-    pub(crate) reports: HashMap<String, StatsReportType>,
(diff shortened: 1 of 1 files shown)


k0nserv:
We could use a macro like 


```rust
macro_rules! define_stats_struct {
    (
        pub struct $struct:ident {
            $(
                $tt:tt
            )*
        }) => {
        #[derive(Debug, Serialize)]
        pub struct $struct {
        // RTCStats
            #[serde(with = "serialize::instant_to_epoch_seconds")]
            pub timestamp: Instant,
            #[serde(rename = "type")]
            pub stats_type: RTCStatsType,
            pub id: String,

            $( $tt )*
        }
    }
}
```

to reduce the repetition of the `RTCStats` fields

melekes (approved):
👍

PR #221, opened 2022-07-14
add RTCPeerConnection::get_remote_dtls_certificate

melekes:
Previously it was not possible to get DTLS certificate of the remote peer from `RTCPeerConnection`. Neither it's possible in Pion. _(NOTE: I'm going to raise a similar issue there)_

Also, I had to always require / set remote certificate (even if certificate verification is disabled via settings). It's in line with Pion https://github.com/pion/webrtc/blob/7feac502862a6a8804179dbadb816654138f34e0/dtlstransport.go#L367-L373.

Refs https://github.com/libp2p/specs/pull/412#discussion_r912612288

### DIFF
--- a/.github/workflows/webrtc.yml
+++ b/.github/workflows/webrtc.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.56.0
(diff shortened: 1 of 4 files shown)


melekes:
As pointed out in https://github.com/pion/webrtc/pull/2284, there's a way already: `sctp().transport().get_remote_certificate()`

PR #225, opened 2022-07-20
Add RTP Stats to stats report

k0nserv:
This PR implements inbound and outbound RTP stats according to the 
[specification](https://www.w3.org/TR/webrtc-stats/), with some 
limitations. 

## Todo

- [x] Inbound RTP
- [x] Outbound RTP
- [x] Tests

### DIFF
--- a/interceptor/src/stats/interceptor.rs
+++ b/interceptor/src/stats/interceptor.rs
@@ -1,10 +1,20 @@
 use std::collections::HashMap;
 use std::fmt;
+use std::hash::Hash;
 use std::sync::Arc;
+use std::time::SystemTime;
 
-use super::{RTPStats, RTPStatsReader};
+use super::{StatsContainer, StatsSnapshot, StreamStats};
 use async_trait::async_trait;
-use util::sync::Mutex;
+use rtcp::payload_feedbacks::full_intra_request::FullIntraRequest;
+use rtcp::payload_feedbacks::picture_loss_indication::PictureLossIndication;
+use rtcp::receiver_report::ReceiverReport;
+use rtcp::transport_feedbacks::transport_layer_nack::TransportLayerNack;
+use rtp::extension::abs_send_time_extension::unix2ntp;
+use tokio::sync::{mpsc, oneshot};
+use tokio::time::Duration;
+
+use util::sync::{Mutex, RwLock};
 use util::{MarshalSize, Unmarshal};
 
 use crate::error::Result;
@@ -12,47 +22,196 @@ use crate::stream_info::StreamInfo;
 use crate::{Attributes, Interceptor, RTCPReader, RTCPWriter, RTPReader, RTPWriter};
 
 #[derive(Debug)]
+enum Message {
+    StatUpdate {
+        ssrc: u32,
+        update: StatsUpdate,
+    },
+    RequestSnapshot {
+        ssrcs: Vec<u32>,
+        chan: oneshot::Sender<Vec<Option<StatsSnapshot>>>,
+    },
+}
+
+#[derive(Debug)]
+enum StatsUpdate {
+    RecvRTP {
+        packets: u64,
+        header_bytes: u64,
+        payload_bytes: u64,
+        last_packet_timestamp: SystemTime,
+    },
+    WriteRTP {
+        packets: u64,
+        header_bytes: u64,
+        payload_bytes: u64,
+        last_packet_timestamp: SystemTime,
+    },
+    RecvRTCP {
+        rtt_ms: Option<f64>,
+        loss: Option<u8>,
+        fir_count: Option<u64>,
+        pli_count: Option<u64>,
+        nack_count: Option<u64>,
+    },
+    WriteRTCP {
+        rtt_ms: Option<f64>,
+        loss: Option<u8>,
+        fir_count: Option<u64>,
+        pli_count: Option<u64>,
+        nack_count: Option<u64>,
+    },
+}
+
 pub struct StatsInterceptor {
+    // Wrapped RTP streams
     recv_streams: Mutex<HashMap<u32, Arc<RTPReadRecorder>>>,
     send_streams: Mutex<HashMap<u32, Arc<RTPWriteRecorder>>>,
+
+    tx: mpsc::Sender<Message>,
+
     id: String,
+    now_gen: Option<Arc<dyn Fn() -> SystemTime + Send + Sync>>,
 }
 
 impl StatsInterceptor {
     pub fn new(id: String) -> Self {
+        let (tx, rx) = mpsc::channel(100);
+
+        tokio::spawn(run_stats_reducer(rx));
+
         Self {
             id,
             recv_streams: Default::default(),
             send_streams: Default::default(),
+            tx,
+            now_gen: None,
         }
     }
 
-    pub fn recv_stats_reader(&self, ssrc: u32) -> Option<RTPStatsReader> {
-        self.recv_stats_readers([ssrc].iter().copied())
-            .into_iter()
-            .next()
+    fn with_time_gen<F>(id: String, now_gen: F) -> Self
+    where
+        F: Fn() -> SystemTime + Send + Sync + 'static,
+    {
+        let (tx, rx) = mpsc::channel(100);
+        tokio::spawn(run_stats_reducer(rx));
+
+        Self {
+            id,
+            recv_streams: Default::default(),
+            send_streams: Default::default(),
+            tx,
+            now_gen: Some(Arc::new(now_gen)),
+        }
     }
 
-    pub fn recv_stats_readers(&self, ssrcs: impl Iterator<Item = u32>) -> Vec<RTPStatsReader> {
-        let lock = self.recv_streams.lock();
+    pub async fn fetch_stats(&self, ssrcs: Vec<u32>) -> Vec<Option<StatsSnapshot>> {
+        let (tx, rx) = oneshot::channel();
 
-        ssrcs
-            .filter_map(|ssrc| lock.get(&ssrc).map(|r| r.reader()))
-            .collect()
+        if let Err(e) = self
+            .tx
+            .send(Message::RequestSnapshot { ssrcs, chan: tx })
+            .await
+        {
+            log::debug!(
+                "Failed to fetch RTP stream stats from stats task with error: {}",
+                e
+            );
+
+            return vec![];
+        }
+
+        rx.await.unwrap_or_default()
     }
+}
 
-    pub fn send_stats_reader(&self, ssrc: u32) -> Option<RTPStatsReader> {
-        self.send_stats_readers([ssrc].iter().copied())
-            .into_iter()
-            .next()
-    }
+async fn run_stats_reducer(mut rx: mpsc::Receiver<Message>) {
+    let mut ssrc_stats: StatsContainer = Default::default();
+    let mut cleanup_ticker = tokio::time::interval(Duration::from_secs(10));
 
-    pub fn send_stats_readers(&self, ssrcs: impl Iterator<Item = u32>) -> Vec<RTPStatsReader> {
-        let lock = self.send_streams.lock();
+    loop {
+        tokio::select! {
+            maybe_msg = rx.recv() => {
+                let msg = match maybe_msg {
+                    Some(m) => m,
+                    None => break,
+                };
+                match msg {
+                    Message::StatUpdate { ssrc, update } => {
+                        let stats = ssrc_stats.get_or_create_stream_stats(ssrc);
 
-        ssrcs
-            .filter_map(|ssrc| lock.get(&ssrc).map(|r| r.reader()))
-            .collect()
+                        match update {
+                            StatsUpdate::RecvRTP {
+                                packets,
+                                header_bytes,
+                                payload_bytes,
+                                last_packet_timestamp,
+                            }
+                            | StatsUpdate::WriteRTP {
+                                packets,
+                                header_bytes,
+                                payload_bytes,
+                                last_packet_timestamp,
+                            } => {
+                                if matches!(update, StatsUpdate::RecvRTP { .. }) {
+                                    stats.rtp_recv_stats.update(
+                                        header_bytes,
+                                        payload_bytes,
+                                        packets,
+                                        last_packet_timestamp,
+                                    );
+                                } else {
+                                    stats.rtp_write_stats.update(
+                                        header_bytes,
+                                        payload_bytes,
+                                        packets,
+                                        last_packet_timestamp,
+                                    );
+                                }
+                            }
+                            StatsUpdate::RecvRTCP {
+                                rtt_ms,
+                                loss,
+                                fir_count,
+                                pli_count,
+                                nack_count,
+                            }
+                            | StatsUpdate::WriteRTCP {
+                                rtt_ms,
+                                loss,
+                                fir_count,
+                                pli_count,
+                                nack_count,
+                            } => {
+                                if matches!(update, StatsUpdate::RecvRTCP { .. }) {
+                                    stats
+                                        .rtcp_recv_stats
+                                        .update(rtt_ms, loss, fir_count, pli_count, nack_count);
+                                } else {
+                                    stats
+                                        .rtcp_write_stats
+                                        .update(rtt_ms, loss, fir_count, pli_count, nack_count);
+                                }
+                            }
+                        }
+
+                        stats.mark_updated();
+                    }
+                    Message::RequestSnapshot { ssrcs, chan } => {
+                        let result = ssrcs
+                            .into_iter()
+                            .map(|ssrc| ssrc_stats.get(ssrc).map(StreamStats::snapshot))
+                            .collect();
+
+                        let _ = chan.send(result);
+                    }
+                }
+
+            }
+            _ = cleanup_ticker.tick() => {
+                ssrc_stats.remove_stale_entries();
+            }
+        }
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/peer_connection/peer_connection_internal.rs
```diff
@@ -1346,13 +1352,90 @@ impl PeerConnectionInternal {
...
+                        });
+                    }
+                }
+
+                let readers = self
+                    .stats_interceptor
+                    .recv_stats_readers(track_infos.iter().map(|t| t.ssrc));
+
+                for (reader, info) in readers.into_iter().zip(track_infos) {
+                    let id = format!("RTCInboundRTP{}Stream_{}", capitalize(info.kind), info.ssrc);
```
k0nserv, line 1403: This attempts to match Chrome's ids for stats ids. @sax Did you try to do this for the other ids or did you establish some other convention?

src/stats/mod.rs
```diff
@@ -422,3 +427,90 @@ impl PeerConnectionStats {
...
+    pub kind: &'static str, // Either "video" or "audio"
+    // TODO: Add tranportId
+    // TODO: Add codecId
+
+    // RTCSentRtpStreamStats
+    pub packets_sent: u64,
+    pub bytes_sent: u64,
+
+    // RTCOutboundRtpStreamStats
+    pub track_identifier: String, // NB: non-canon
```
k0nserv, line 499: I don't quite understand why this isn't in the specification. One of these is emitted per simulcast SSRC in the cast of simulcast, but it would still be useful to include the track's identifier imo.

src/peer_connection/peer_connection_internal.rs
sax: Where there was an id coming from a lower-level system I retained it, but where there was no id I tried to do the same thing.

Last I looked at different browsers was over a year ago, at which point it seemed like Safari was close (but not exactly the same) as Chrome, and Firefox did its own thing. So much has changed in the last year in terms of webrtc, however.

src/stats/mod.rs
sax: I haven't had a whole lot of time to read through the specification, but my guess is that pending a formal specification draft, the different vendors all guessed at what would be useful for themselves.

Since Pion predates the draft recommendation, I'm not surprised that there are fields that aren't documented in any spec.

I was starting to wonder if this might be a place where webrtc-rs moves out in front of Pion in terms of coding to the spec, or if the long-term intention is to move in concert with Pion.

src/peer_connection/peer_connection_internal.rs
k0nserv: Hmm, I guess I should review what the others do and establish a convention then

src/stats/mod.rs
k0nserv: This isn't actually ported from Pion, this code doesn't exist in Pion. FWIW I found the answer to why it's not in the specification, will push a commit that documents this.

interceptor/src/stats/interceptor.rs
```diff
@@ -1,58 +1,217 @@
...
+
+    loop {
+        tokio::select! {
+            maybe_msg = rx.recv() => {
+                let msg = match maybe_msg {
+                    Some(m) => m,
+                    None => break,
+                };
+                match msg {
+                    Message::StatUpdate { ssrc, update } => {
```
k0nserv, line 140: Maybe we move this whole thing to a function since formatting and highlighting breaks within `tokio::select!` because it's a macro

interceptor/src/stats/interceptor.rs
```diff
@@ -124,25 +291,206 @@ impl Interceptor for StatsInterceptor {
...
+                    let count = nack.nacks.iter().flat_map(|p| p.into_iter()).count() as u64;
+
+                    let e = acc.entry(nack.media_ssrc).or_default();
+                    e.nack_count = e.nack_count.map(|v| v + count).or(Some(count));
+                }
+
+                acc
+            });
+
+        dbg!(&updates);
```
k0nserv, line 449: Needs to go

interceptor/src/stats/interceptor.rs
```diff
@@ -155,39 +503,37 @@ impl RTPReader for RTPReadRecorder {
         let mut b = &buf[..bytes_read];
         let packet = rtp::packet::Packet::unmarshal(&mut b)?;
 
-        self.stats.update(
```
k0nserv, line 158: This used to be based on a collection of `Arc<Atomic*>`s, but this method is brittle because there is not synchronisation between each value. With the new approach a spawned task(one per PC) is in charge of reducing stats and no incomplete updates can be observed.

interceptor/src/stats/interceptor.rs
```diff
@@ -197,29 +543,92 @@ impl RTPWriter for RTPWriteRecorder {
...
         Ok(n)
     }
 }
 
-impl fmt::Debug for RTPWriteRecorder {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("RTPWriteRecorder")
-            .field("stats", &self.stats)
-            .finish()
+trait GetOrCreateAtomic<T, U> {
```
k0nserv, line 563: This isn't used anymore, delete

webrtc/src/peer_connection/peer_connection_test.rs
```diff
@@ -280,22 +293,95 @@ async fn test_get_stats() -> Result<()> {
...
-    match stats.reports.get("ice_transport") {
+    match offer_stats.reports.get("ice_transport") {
         Some(StatsReportType::Transport(ice_transport_stats)) => {
             assert!(ice_transport_stats.bytes_received > 0);
             assert!(ice_transport_stats.bytes_sent > 0);
         }
         Some(_other) => panic!("found the wrong type"),
         None => panic!("missed it"),
     }
+    let outbound_stats = offer_stats
```
k0nserv, line 359: I'll explore extending this test to also have some RTCP components to it

k0nserv:
Will tackle Remote{Inbound,Outbound}RTP in another PR

interceptor/src/stats/mod.rs
```diff
@@ -10,83 +12,180 @@ pub fn make_stats_interceptor(id: &str) -> Arc<StatsInterceptor> {
...
+    }
+
+    fn get(&self, ssrc: u32) -> Option<&StreamStats> {
+        self.stream_stats.get(&ssrc)
+    }
+
+    fn remove_stale_entries(&mut self) {
+        self.stream_stats
+            .retain(|_, s| s.duration_since_last_update() < Duration::from_secs(60))
+    }
```
lolgesten, line 87: Where is this called from?

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
```diff
@@ -27,40 +27,69 @@ pub struct NackPair {
...
+
+                while self.bitfield != 0 {
+                    if (self.bitfield & (1 << i)) != 0 {
+                        self.bitfield &= !(1 << i);
+                        self.index = Some(i + 1);
+
+                        return Some(self.packet_id.wrapping_add(i + 1));
+                    }
+
+                    i += 1;
```
lolgesten, line 57: This algo is a bit hard to read because what's technically one field have 3 different names: `self.index`, `id` and `i`. Would be good to be consistent.

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
```diff
@@ -27,40 +27,69 @@ pub struct NackPair {
...
-                out.push(self.packet_id + i + 1);
+                Some(self.packet_id)
             }
-            i += 1;
+            Some(id) if self.bitfield != 0 => {
+                let mut i = id;
+
+                while self.bitfield != 0 {
+                    if (self.bitfield & (1 << i)) != 0 {
+                        self.bitfield &= !(1 << i);
```
lolgesten, line 51: This looks strange. If we have this:

```
packet_id: 123
bitfield: `00000000 00000101`
```

We expect the iterator to produce `123`, `124`, and `126`. 

But it looks like `self.bitfield` is "destroyed" on the first iteration which means the result would be `123`, `124` – and missing the last value.

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
```diff
@@ -27,40 +27,69 @@ pub struct NackPair {
...
+impl IntoIterator for NackPair {
+    type Item = u16;
+
+    type IntoIter = NackIterator;
+
+    fn into_iter(self) -> Self::IntoIter {
+        NackIterator {
+            packet_id: self.packet_id,
+            bitfield: self.lost_packets,
+            index: None,
```
lolgesten, line 92: Wouldn't it be better to have `0` here? And increase by 1 for each iteration. It looks like this currently can oscillate, so you get:

None => 0 => None => 0

I.e. `iter.next()` would produce values again if you just "continue" after it produced the first `None`.

lolgesten (commented):
Some questions

interceptor/src/stats/interceptor.rs
```diff
@@ -1,58 +1,217 @@
...
+        packets: u64,
+        header_bytes: u64,
+        payload_bytes: u64,
+        last_packet_timestamp: SystemTime,
+    },
+    WriteRTP {
+        packets: u64,
+        header_bytes: u64,
+        payload_bytes: u64,
+        last_packet_timestamp: SystemTime,
```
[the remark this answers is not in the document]
k0nserv: I think `Instant` as it has been used in the stats API is dubious. Instant is about measuring absolute elapsed time, but this is about a specific point in time. Ultimately this will be provided to the consumer as unix epoch seconds so `SystemTime` feels appropriate. There's a [hack](https://github.com/webrtc-rs/webrtc/blob/59be92b2dfd5400d56a3d32e6f0ed62d7e8dfd7c/webrtc/src/stats/serialize.rs#L12-L26) to convert an `Instant` to `SystemTime` which we currently use, but that feels dubious.

interceptor/src/stats/interceptor.rs
```diff
@@ -1,58 +1,217 @@
...
+
 pub struct StatsInterceptor {
+    // Wrapped RTP streams
     recv_streams: Mutex<HashMap<u32, Arc<RTPReadRecorder>>>,
     send_streams: Mutex<HashMap<u32, Arc<RTPWriteRecorder>>>,
+
+    tx: mpsc::Sender<Message>,
+
     id: String,
+    now_gen: Option<Arc<dyn Fn() -> SystemTime + Send + Sync>>,
```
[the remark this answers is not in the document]
k0nserv: I suppose yeah

interceptor/src/stats/interceptor.rs
```diff
@@ -1,58 +1,217 @@
...
+            log::debug!(
+                "Failed to fetch RTP stream stats from stats task with error: {}",
+                e
+            );
 
-    pub fn send_stats_reader(&self, ssrc: u32) -> Option<RTPStatsReader> {
-        self.send_stats_readers([ssrc].iter().copied())
-            .into_iter()
-            .next()
-    }
```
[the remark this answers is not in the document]
k0nserv: I guess so, but I'm not sure if we have had a release since this was introduced.

interceptor/src/stats/interceptor.rs
```diff
@@ -1,58 +1,217 @@
...
+                                last_packet_timestamp,
+                            }
+                            | StatsUpdate::WriteRTP {
+                                packets,
+                                header_bytes,
+                                payload_bytes,
+                                last_packet_timestamp,
+                            } => {
+                                if matches!(update, StatsUpdate::RecvRTP { .. }) {
+                                    stats.rtp_recv_stats.update(
```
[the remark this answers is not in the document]
k0nserv: Maybe yes, I'll play around.

interceptor/src/stats/mod.rs
k0nserv: From the stats reducer task. It was the simplest way I could think of to not have unbounded memory growth. It would be better if we knew exactly when a given RTP stream went away so we could clean it up, but that seemed much more complex

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
k0nserv: This specific line zeroes out the bit in question which ensure we eventually hit the loop condition.

Keep in mind that we are within a `next` method here, as such we only run until we can produce a value, subsequent calls to `next` should produce another value if we do indeed have more values to produce.

I'll write a test with a gap like you pointed out to verify

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
k0nserv: I'll fix

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
k0nserv: `None`is different from `0`. 

For example, with the above values you used:

```
packet_id: 123
bitfield: `00000000 00000101`
```

The first call to `next` should produce `123` and also set `index` to `Some(0)` indicating that the next call to `next` should return a value based on the bitfield if any.

However, since each call to next that moves forward modifies `bitfield` I think we could get by with a `has_yielded_packet_id: bool` and always start the indexing from `0`, slightly less efficient but clearer.

k0nserv (commented):
Thanks for the review!

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
k0nserv: This is said alternative approach

```diff
diff --git a/rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs b/rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
index 8761a433..43db9e9f 100644
--- a/rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
+++ b/rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
@@ -30,36 +30,30 @@ pub type RangeFn =
 pub struct NackIterator {
     packet_id: u16,
     bitfield: PacketBitmap,
-    index: Option<u16>,
+    has_yielded_packet_id: bool,
 }
 
 impl Iterator for NackIterator {
     type Item = u16;
 
     fn next(&mut self) -> Option<Self::Item> {
-        match self.index {
-            None => {
-                self.index = Some(0);
+        if !self.has_yielded_packet_id {
+            self.has_yielded_packet_id = true;
 
-                Some(self.packet_id)
-            }
-            Some(id) if self.bitfield != 0 => {
-                let mut i = id;
-
-                while self.bitfield != 0 {
-                    if (self.bitfield & (1 << i)) != 0 {
-                        self.bitfield &= !(1 << i);
-                        self.index = Some(i + 1);
+            Some(self.packet_id)
+        } else {
+            let mut i = 0;
 
-                        return Some(self.packet_id.wrapping_add(i + 1));
-                    }
+            while self.bitfield != 0 {
+                if (self.bitfield & (1 << i)) != 0 {
+                    self.bitfield &= !(1 << i);
 
-                    i += 1;
+                    return Some(self.packet_id.wrapping_add(i + 1));
                 }
 
-                None
+                i += 1;
             }
-            _ => None,
+            None
         }
     }
 }
@@ -89,7 +83,7 @@ impl IntoIterator for NackPair {
         NackIterator {
             packet_id: self.packet_id,
             bitfield: self.lost_packets,
-            index: None,
+            has_yielded_packet_id: false,
         }
     }
 }
diff --git a/rtcp/src/transport_feedbacks/transport_layer_nack/transport_layer_nack_test.rs b/rtcp/src/transport_feedbacks/transport_layer_nack/transport_layer_nack_test.rs
index e83f1f7c..e8c6b76f 100644
--- a/rtcp/src/transport_feedbacks/transport_layer_nack/transport_layer_nack_test.rs
+++ b/rtcp/src/transport_feedbacks/transport_layer_nack/transport_layer_nack_test.rs
@@ -199,6 +199,15 @@ fn test_nack_pair() {
             lost_packets: 0b0000_0111,
         },
     );
+
+    // Gap
+    test_nack(
+        vec![123, 124, 126],
+        NackPair {
+            packet_id: 123,
+            lost_packets: 0b0000_0101,
+        },
+    );
 }
 
 #[tokio::test]
```

k0nserv: Let me know if you prefer this variant

interceptor/src/stats/interceptor.rs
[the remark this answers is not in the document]
algesten: Alright!

interceptor/src/stats/mod.rs
algesten: Ok

interceptor/src/stats/interceptor.rs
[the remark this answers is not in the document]
algesten: ok

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
algesten: That is better, yes.

algesten (approved):
I prefer the alternative impl for NackIter, but other than that. Good to go!

rtcp/src/transport_feedbacks/transport_layer_nack/mod.rs
k0nserv: Done

PR #226, opened 2022-07-20
base id updating on whether it's been finalized, not on its value

stuqdog:
Currently, a data channel id is updated on open if it's set to 0. This can be a problem if a data channel is created through a peer connection with the id intentionally set to 0. This pr adds a new field to `DataChannelParameters` and `RTCDataChannelInit` (`id_finalized`), which is set to `true` when a user intentionally provides an id, allowing us to respect that id choice on data channel open regardless of its value.

### DIFF
--- a/src/data_channel/data_channel_parameters.rs
+++ b/src/data_channel/data_channel_parameters.rs
@@ -10,4 +10,5 @@ pub struct DataChannelParameters {
     pub max_packet_life_time: u16,
     pub max_retransmits: u16,
     pub negotiated: bool,
+    pub(crate) id_finalized: bool,
 }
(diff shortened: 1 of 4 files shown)


k0nserv:
Is there [spec](https://w3c.github.io/webrtc-pc/) guidance on this, if so does this follow what the specification says to do?

k0nserv:
To answer, my own question: I had a quick look at the specification and I believe `RTCDataChannelInit.[[negotiated]]` is intended for this purposes.

stuqdog:
> Is there [spec](https://w3c.github.io/webrtc-pc/) guidance on this, if so does this follow what the specification says to do?

🤦 probably should have checked that! Thanks, appreciate you pointing this out :) I just updated to remove the `id_finalized` param and instead check on the `negotiated` field.

k0nserv (approved):
LGTM

melekes:
```
            .peer_connection
            .create_data_channel(
                "data",
                Some(RTCDataChannelInit {
                    id: Some(1),
                    negotiated: None,
                    ..RTCDataChannelInit::default()
                }),
            )
```

creates a data channel with `id=0` now, which is incorrect.

PR #229, opened 2022-08-02
Revert "base id updating on whether it's been negotiated, not on its …

melekes:
…value (#226)"

This reverts commit 823c3f80f13b40a7454efebc8f40ddbc04c24099.

As pointed out in [the original PR](https://github.com/webrtc-rs/webrtc/pull/226#issuecomment-1202028922), this commit broke the API guarantees we had before. I don't think I have time right now to fix this properly, but maybe @stuqdog does?

### DIFF
--- a/src/data_channel/mod.rs
+++ b/src/data_channel/mod.rs
@@ -157,7 +157,7 @@ impl RTCDataChannel {
                 negotiated: self.negotiated,
             };
 
-            if !self.negotiated {
+            if self.id.load(Ordering::SeqCst) == 0 {
                 self.id.store(
                     sctp_transport
                         .generate_and_set_data_channel_id(


stuqdog:
@melekes I'll start looking into this tomorrow.

melekes:
> @melekes I'll start looking into this tomorrow.

Great, thank you 🙏

stuqdog:
> > @melekes I'll start looking into this tomorrow.
> 
> Great, thank you 🙏

Apologies for the delay, some stuff came up! Per looking at the spec, I think we can actually get away with a pretty simple change here (https://github.com/melekes/webrtc/pull/1). This isn't _perfect_ (e.g., we're still sad if a user wants to set `id: 0` and `negotiated: false`), but I believe it addresses the problem cases we've been encountering. And, replacing the existing convention of `0 == null` while maintaining consistency with the spec is definitely more invasive (e.g., changing ID to `Option<AtomicU16>` means add mutability to lots of call sites). 

If you prefer the more invasive approach then let me know!

k0nserv (approved):

PR #236, opened 2022-08-22
Pause RTP writing if direction indicates it

algesten:
Currently if `RTCRtpTransceiver::direction` is set to a valueq which
doesn't allow sending data, we happily keep sending data if the user
does `write_sample` or `write_rtp`.

This changes so that any writes are just dropped as long as the
direction doesn't permit sending.

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -1718,6 +1718,7 @@ impl RTCPeerConnection {
                             Arc::clone(&self.internal.dtls_transport),
                             Arc::clone(&self.internal.media_engine),
                             Arc::clone(&self.interceptor),
+                            false, // adding a track sets a send direction.
                         )
                         .await,
                     );
--- a/src/rtp_transceiver/mod.rs
+++ b/src/rtp_transceiver/mod.rs
@@ -400,27 +400,25 @@ impl RTCRtpTransceiver {
                 previous_direction,
                 current_direction
             );
+        } else {
+            // no change.
+            return Ok(());
         }
 
-        match (previous_direction, current_direction) {
-            (a, b) if a == b => {
-                // No change, do nothing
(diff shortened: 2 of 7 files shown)


src/peer_connection/mod.rs
```diff
@@ -1718,6 +1718,7 @@ impl RTCPeerConnection {
                             Arc::clone(&self.internal.dtls_transport),
                             Arc::clone(&self.internal.media_engine),
                             Arc::clone(&self.interceptor),
+                            !t.direction().has_send(),
```
k0nserv, line 1721: This is wrong, the act of adding a track should make the transceiver's direction become send. Thus this should be `true` unconditionally.

src/rtp_transceiver/mod.rs
```diff
@@ -348,6 +348,11 @@ impl RTCRtpTransceiver {
                 previous,
                 d
             );
+
+            let sender = self.sender.lock().await;
+            if let Some(sender) = &*sender {
```
k0nserv, line 353: The `direction` of a Transceiver reflects the desired state, as such we shouldn't change anything about sending/receiving here. Instead it should be based on the `current_direction` which reflects the actually negotiated value. 

The appropriate place to set the sender as paused or not is in `process_new_current_direction`.

src/peer_connection/mod.rs
lolgesten: alright

algesten: alright

src/rtp_transceiver/mod.rs
k0nserv: There's a TODO for this on line 422

algesten: ah alright. wasn't aware of that one.

k0nserv (approved):
LGTM

Issue #235, opened 2022-08-22
Release 0.5.0


k0nserv:
We haven't had a release of the top level crate(`webrtc`) in some time and there are a significant number of useful changes that we should release, therefore it's time to cut 0.5.0.

The primary thing that needs doing for this to happen is creating the release notes. Especially I'd like the release notes to make it clear what breaking changes have happened since 0.4.0.

## Todo

+ [x] Create release notes
+ [x] Release all sub-crates @k0nserv 
+ [x] Bump versions of sub-crates to latest
+ [x] Merge [RTP stats PR](https://github.com/webrtc-rs/webrtc/pull/225) @k0nserv (_optional_)
+ [x] Properly fix the issue identified in https://github.com/webrtc-rs/webrtc/pull/229 @stuqdog (_optional_)

The two optional changes can be part of a later release

stuqdog:
I wrote a mostly*-fix for the id updating issue a couple weeks ago, but foolishly forgot to add a reviewer! 🤦

It's a one-liner [here](https://github.com/melekes/webrtc/pull/1). It was originally a PR on @melekes's revert PR, but I can easily move it over to the main branch.

Tl;dr of wall of text below: the solution I wrote is very simple, can be merged immediately, and handles all cases correctly except one (which is strictly better than we had before). A perfect solution is more invasive and adds a lot of mutability to the code base that could be problematic. If the quick and easy solution is good I'll create the PR right now, if not I'll start writing the more elaborate one but it will take longer to complete. @k0nserv please let me know what you think!

*I say "mostly" because there's still one case it doesn't handle well. To quote from [here](https://github.com/webrtc-rs/webrtc/pull/229#issuecomment-1208309753): This isn't _perfect_ (e.g., we're still sad if a user wants to set id: 0 and negotiated: false), but I believe it addresses the problem cases we've been encountering. And, replacing the existing convention of 0 == null while maintaining consistency with the spec is definitely more invasive (e.g., changing ID to Option<AtomicU16> means add mutability to lots of call sites).

[...]

PR #237, opened 2022-08-22
Fix datachannel id setting for 0.5.0 release

stuqdog:
Per comments [here](https://github.com/webrtc-rs/webrtc/issues/235), this prevents a `RTCDataChannel id` from updating on open unless a) the value was not negotiated, and b) the value is equal to zero (i.e., the `null` value by existing crate convention). This fixes most of the existing error cases and does not introduce new ones. It does, however, miss one specific case (where a user creates a non-negotiated data channel with `id = 0`). Fixing that case would be a sizeable refactor. There is a TODO comment for addressing it.


stuqdog:
@k0nserv @melekes fyi, I went ahead and created this small PR to address convo on [0.5.0 release](https://github.com/webrtc-rs/webrtc/issues/235). 

We could also create a new `RTCDataChannel` param to say whether or not the ID was set, which would handle all cases. This was rejected in an earlier PR because it added a value not discussed in the spec, but I think it's worth revisiting. The change from the spec is benign (it doesn't modify any of the existing values), it's a small and easy-to-reason-about change, and it would allow us to handle all cases correctly.

k0nserv:
@stuqdog it sounds like the primary problem here is that spec wants us to use `null` but since we are using an atomic value that's not available to us. Currently the sentinel value used to indicate `null` is `0`. If this was an `u64` value I would suggest the sentinel value `u64::MAX`, but since it's a `u16` we could just store it internally as `u32` and use `u32::MAX` to mean `null`. If I've understood things correctly this would allow us to be spec compliant and correct.

WDYT?

stuqdog:
> @stuqdog it sounds like the primary problem here is that spec wants us to use `null` but since we are using an atomic value that's not available to us. Currently the sentinel value used to indicate `null` is `0`. If this was an `u64` value I would suggest the sentinel value `u64::MAX`, but since it's a `u16` we could just store it internally as `u32` and use `u32::MAX` to mean `null`. If I've understood things correctly this would allow us to be spec compliant and correct.
> 
> WDYT?

@k0nserv Thanks for the suggestion! Unfortunately the spec wants us to use an unsigned short, i.e. u16. However, some [digging](https://w3c.github.io/webrtc-pc/#methods-11) shows that `u16::MAX (65535)` is actually not a valid data channel ID value. So I think we can use `u16::MAX` as our sentinel value. We should probably also add a validation check which prevents users from passing `u16::MAX` as ID when they create a new data channel to maintain compliance.

This is a pretty simple change so I'm gonna go ahead and start writing it up now, let he know if you have any concerns!

stuqdog:
update: @k0nserv @melekes I'm holding off on coding right now, but on additional digging into the spec I think the solution I wrote before was correct and we were wrong to revert it. See [here](https://w3c.github.io/webrtc-pc/#dom-peerconnection-createdatachannel) and specifically bullet point 13 and the accompanying note. Quoted below for convenience: 

> Initialize channel.[[[DataChannelId]]](https://w3c.github.io/webrtc-pc/#dfn-datachannelid) to the value of option.[id](https://w3c.github.io/webrtc-pc/#dom-rtcdatachannelinit-id), if it is present and [[[Negotiated]]](https://w3c.github.io/webrtc-pc/#dfn-negotiated) is true, otherwise null.
> Note
> This means the [id](https://w3c.github.io/webrtc-pc/#dom-rtcdatachannelinit-id) member will be ignored if the data channel is negotiated in-band; this is intentional. Data channels negotiated in-band should have IDs selected based on the DTLS role, as specified in [[RFC8832](https://w3c.github.io/webrtc-pc/#bib-rfc8832)].

Based on my reading of this, it's actually correct to update the ID value of a channel that is non-negotiated but has an ID. I think we should return to updating the ID if and only if negotiated is false, and additionally add the error handling for a case where a user tries to pass an ID value of `u16::MAX`. If this is breaking existing API guarantees in ways that are too disruptive right now, then I'm happy to just go with my previous solution, i.e. using `u16::MAX` as a sentinel value and updating ID based on that, but in that case I think we should add some sort of note that this is a compromise solution rather than maintaining total fidelity with the spec.

What do you two think?

k0nserv:
Awesome digging, I think it sounds like you are right. In fact, given that we are using Rust we could improve the api here maybe?

`RTCDataChannelInit` is currently a bit strange:

https://github.com/webrtc-rs/webrtc/blob/45e66fc3a9e3a66cf7584e018ced6512621f86c5/webrtc/src/data_channel/data_channel_init.rs#L4-L32

Why is `negotiated` an `Option<bool>`? To answer my own question, I guess because we want to have a spreadable `Default` implementation. 

I would suggest removing `id` from `RTCDataChannelInit` and changing the type of `negotiated` to `Option<u16>`, the same change would then be applied to `DataChannelParameters`. Only if `DataChannelParameters::negotiated` is `Some` would we use the user's provided id, otherwise we'd generate one.

@melekes, thoughts?

melekes:
> Data channels negotiated in-band should have IDs selected based on the DTLS role

what does that mean in terms of code?

> I would suggest removing id from RTCDataChannelInit and changing the type of negotiated to Option<u16>, the same change would then be applied to DataChannelParameters. Only if DataChannelParameters::negotiated is Some would we use the user's provided id, otherwise we'd generate one.

then it won't be possible to establish a data channel between two peers programmatically. e.g. inside libp2p transport I'm working on right now I open 1 data channel on the client w/ `negotiated: false` and id=1. On the server I also open a data channel w/ `negotiated: true` and id=1. If we remove ability to set id, how could I open an already negotiated channel on the server (i.e. id could be different each time)?

k0nserv:
@melekes if you have negotiated ids on both sides you'd want to use `negotiated: true, id: 1` on both sides, that's how out of band negotiated ids are supposed to work I believe.

melekes:
@k0nserv indeed, thanks! I'm in favor of https://github.com/webrtc-rs/webrtc/pull/237#issuecomment-1227035085 then 👍

stuqdog:
Agreed, I think updating the API to combine `id` and `negotiated` sounds great. I'll write up the fix today!

k0nserv (approved):
Let me fix CI, then you can rebase and it should pass

melekes (approved):
👍

k0nserv:
Aight, CI should be okay now, try rebasing on `master`

k0nserv:
I think something is broken with the `test_data_channel_ortc_e2e` because of these changes.

stuqdog:
> I think something is broken with the `test_data_channel_ortc_e2e` because of these changes.

Yeah, looks like it 😞 I'll dig.

stuqdog:
So it seems that `test_data_channel_ortc_e2e` was relying on a non-negotiated channel. I'm not sure where that requirement exists, but changing `negotiated` to `None` (consistent with what it was before the changes to `id`/`negotiated`) makes all tests pass successfully.

k0nserv:
Yeah this old code was weird, it specifies an id but it doesn't set `negotiated: true` and as such wasn't using a pre-negotiated channel. Using `negotiated: None` seems like the right change.

k0nserv:
Excellent work @stuqdog, thanks so much for seeing this through 🌟

PR #239, opened 2022-08-23
One repo (again)

algesten:
1. Used script here to rewrite history https://stackoverflow.com/a/43554035/437679
2. Moved all crates/* to top level
3. Moved src, Cargo.toml to webrtc/
4. Rewrite all Cargo.toml to use paths and add workspace on top level


k0nserv (approved):
Let's do it

PR #243, opened 2022-08-23
Fixes for cargo minimal-versions

algesten:


k0nserv:
This fails locally for me still

image

algesten:
Yeah, me too. I wanted to see if the CI had more luck cause I'm stuck.

algesten:
Ok, let's try.

k0nserv:
I dug some more, the problem is a combination of minimal versions being very aggressive(downgrading even transitive dependencies) and serde specifying incorrect dependencies(at least historically). 

For example,

If you fix `serde` to `1.0.51` then you must also fix `serde_derive` to `1.0.51` because `serde@1.0.51`, incorrectly, specifies `serder_derive^1.0`. If you fix that then you get errors further down because `serde_derive@1.0.51`, incorrectly, specifies `0.13.0` for `syn` but uses a method `make_where_clause` that didn't exist in said version of `syn`.

algesten:
is it worth keeping this check?

k0nserv:
> is it worth keeping this check?

I think the check is quite useful, but if it's going to be this painful, then no. @melekes WDYT?

algesten:
@k0nserv i couldn't get 104 to work locally, but 110 did.

k0nserv:
Note: I rebased this on master which includes the fix for SCTP that ought to make this work properly

k0nserv (approved):

melekes (commented):
👍

PR #247, opened 2022-08-23
Fix SCTP fuzz tests

k0nserv:


xnorpx:
should artifacts be in git-lfs if they are large?

k0nserv:
>should artifacts be in git-lfs if they are large?
Probably, but these are absolutely tiny

xnorpx:
should the artifacts still be in gitignore and then force add the binary files to keep it as it was before?

xnorpx:
other than that LGTM

algesten (approved):

k0nserv:
> should the artifacts still be in gitignore and then force add the binary files to keep it as it was before?

I don't think so, it's mostly just confusing and potentially problematic to do it this way

PR #250, opened 2022-08-23
Fix log at ^0.4.16 to make tests compile

k0nserv:


k0nserv:
Before this the tests were failing with the error detailed in this issue https://github.com/rust-lang/log/issues/500

algesten (approved):

PR #258, opened 2022-08-24
Fix test_receiver_interceptor_jitter on Ubuntu

k0nserv:
A [previous fix][0] for this issue has proven not to work. With this we
make time entirely deterministic in this test, a practice that we should
use in other tests if successful here.

[0]: https://github.com/webrtc-rs/interceptor/commit/0a2e942e0591e45bfeba6eec7f18960f96e3c5ba

### DIFF
--- a/interceptor/src/report/receiver/receiver_test.rs
+++ b/interceptor/src/report/receiver/receiver_test.rs
@@ -619,7 +619,7 @@ async fn test_receiver_interceptor_reordered_packets() -> Result<()> {
     Ok(())
 }
 
-#[tokio::test]
+#[tokio::test(start_paused = true)]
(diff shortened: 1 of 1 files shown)


k0nserv:
I initially introduced this problem in https://github.com/webrtc-rs/interceptor/pull/5

k0nserv:
Wanna see if CI passes first :)

k0nserv:
This seems like an improvement, but we have tests on Windows that will run forever due to similar issues. I've stopped the Windows builds to not hog those machines.

I also created an [issue](https://github.com/webrtc-rs/webrtc/issues/259) to tackle this in other parts of the project.

PR #261, opened 2022-08-24
Build windows on C:\ where github actions have more disk space

algesten:


.github/workflows/cargo.yml
```diff
@@ -42,6 +42,41 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-cargo-registry-
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.toolchain }}
+          profile: minimal
+          override: true
+      - name: Copy to C drive
+         run: cp D:\a C:\ -Recurse
+      # - name: 🏭 Cache dependencies
```
k0nserv, line 69: Would we still not want the dependency caching? Presumably D:\ has enough space for it

algesten: I want to see the build working without this complication first. I don't quite understand where the caching is doing its thing.

algesten: But then for some reason I only get the coverage report. Why can't I have all the github actions? :'(

k0nserv (approved):

Issue #259, opened 2022-08-24
Make tests that rely on time deterimnistic


k0nserv:
There are many tests that use `tokio::time::sleep` to advance time and observe the effect of timers/tickers running. This is brittle and flaky.

<summary>

`$ rg -g '**/*/*test.rs' sleep`

<details>
data/src/data_channel/data_channel_test.rs:        tokio::time::sleep(Duration::from_millis(10)).await;
data/src/data_channel/data_channel_test.rs:        let timer = tokio::time::sleep(Duration::from_millis(10));
data/src/data_channel/data_channel_test.rs:        let timer = tokio::time::sleep(Duration::from_millis(10));
data/src/data_channel/data_channel_test.rs:    tokio::time::sleep(Duration::from_millis(100)).await;
data/src/data_channel/data_channel_test.rs:        tokio::time::sleep(Duration::from_millis(100)).await;
data/src/data_channel/data_channel_test.rs:    tokio::time::sleep(Duration::from_millis(100)).await;
data/src/data_channel/data_channel_test.rs:        tokio::time::sleep(Duration::from_millis(10)).await;
sctp/src/timer/timer_test.rs:use tokio::time::{sleep, Duration};
sctp/src/timer/timer_test.rs:        sleep(ACK_INTERVAL + Duration::from_millis(50)).await;
sctp/src/timer/timer_test.rs:        sleep(Duration::from_millis(650)).await;
sctp/src/timer/timer_test.rs:        sleep(Duration::from_millis((interval * 3) / 2)).await;
sctp/src/timer/timer_test.rs:        sleep(Duration::from_millis((interval * 3) / 2)).await;
sctp/src/timer/timer_test.rs:        sleep(Duration::from_millis((interval * 3) / 2)).await;
sctp/src/timer/timer_test.rs:        sleep(Duration::from_millis(100)).await;
dtls/src/conn/conn_test.rs:    tokio::time::sleep(Duration::from_millis(1)).await;
dtls/src/conn/conn_test.rs:    tokio::time::sleep(Duration::from_millis(1)).await;
dtls/src/conn/conn_test.rs:    tokio::time::sleep(Duration::from_millis(1)).await;
dtls/src/conn/conn_test.rs:            let timer = tokio::time::sleep(Duration::from_millis(10));
dtls/src/conn/conn_test.rs:    //tokio::time::sleep(Duration::from_millis(300)).await;
[...]

PR #262, opened 2022-08-24
Make tests run clean on windows.

xnorpx:
Enable high resolution timer on windows but disable tests that could not run
reliable even with high resolution timers.

### DIFF
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -25,3 +25,7 @@ thiserror = "1.0"
 tokio-test = "0.4.0" # must match the min version of the `tokio` crate above
 env_logger = "0.9.0"
 chrono = "0.4.19"
+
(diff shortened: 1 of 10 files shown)


k0nserv:
Thanks for the PR.

To be honest this seems like a poor fix compared to the approach outlined in https://github.com/webrtc-rs/webrtc/issues/259 relying on `tokio::time::advance`. Is there something I'm missing here, i.e. does this address something different or meant more as a stopgap until the `tokio::time::advance` approach can be applied?

If it's meant as a stopgap and was deemed simpler than the `tokio::time::advance` approach I think this PR is still worth merging if it does alleviate the flakiness

xnorpx:
tokio advanced is a better solution, (note this is needed if you want high resolution timers in production code on windows)

PR #264, opened 2022-08-24
Disable some flaky tests on windows

algesten:


k0nserv (approved):
Only point I'd make is that we should perhaps leave a comment about why these are disabled

algesten:
ok i fix

PR #268, opened 2022-08-25
Fix current direction update when applying answer

k0nserv:
When applying an answer from the remote the value of
`current_direction`was correctly update, but the process to trigger any
side effect was incorrectly using `direction` instead of
`current_direction` to determine the previous direction.


algesten (approved):

k0nserv:
Just gotta rebase this

PR #275, opened 2022-08-30
mdns: get_interface_addr_for_ip once per query

melekes:
Closes https://github.com/webrtc-rs/mdns/issues/23

Port of https://github.com/webrtc-rs/mdns/pull/25

### DIFF
--- a/mdns/src/conn/mod.rs
+++ b/mdns/src/conn/mod.rs
@@ -314,9 +314,13 @@ async fn run(
 
         for local_name in local_names {
             if *local_name == q.name.data {
-                if interface_addr.is_none() {
-                    let addr = match get_interface_addr_for_ip(src).await {
-                        Ok(addr) => addr,
+                let interface_addr = match interface_addr {
+                    Some(addr) => addr,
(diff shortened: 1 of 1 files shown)


mdns/src/conn/mod.rs
```diff
@@ -313,17 +314,21 @@ async fn run(
...
-                    Ok(e) => e,
-                    Err(e) => {
-                        log::warn!(
-                            "Failed to get local interface to communicate with {}: {:?}",
-                            &src,
-                            e
-                        );
-                        continue;
-                    }
+                if interface_addr.is_none() {
```
k0nserv, line 317: We can avoid the unwrap with

```rust
let interface_addr = match interface_adr {
    Some(addr) => addr,
    None => {
        match get_interface_addr_for_ip(src).await {
            Ok(addr) => addr,
            Err(e) => {
                log::warn!(
                    "Failed to get local interface to communicate with {}: {:?}",
                    &src,
                    e
                );
                continue;
            }
        }
    }
};
```

Which seems desirable

k0nserv (approved):

PR #276, opened 2022-08-31
relax regex version requirement

melekes:
some projects like https://github.com/bheisler/criterion.rs depend on regex: `1.3`. I don't see why 1.6 have to be a requirement.

### DIFF
--- a/ice/Cargo.toml
+++ b/ice/Cargo.toml
@@ -31,7 +31,7 @@ waitgroup = "0.1.2"
 
 [dev-dependencies]
 tokio-test = "0.4.0" # must match the min version of the `tokio` crate above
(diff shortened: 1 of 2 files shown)


k0nserv:
There's a security fix in 1.5.5 that we might want, can you explore if we are affected by it? The relevant CVE is [CVE-2022-24713](https://nvd.nist.gov/vuln/detail/CVE-2022-24713)

melekes:
" This makes it possible to perform denial of service attacks by sending specially crafted regexes to services accepting user-controlled, untrusted regexes."

we do not accept user regexes, so we're unaffected.

k0nserv (approved):
This will need a rebase for CI

PR #278, opened 2022-08-31
Fix async-global-executor

k0nserv:
Build broke on 1.57.0 due to 2.3.0 being released

### DIFF
--- a/util/Cargo.toml
+++ b/util/Cargo.toml
@@ -53,6 +53,7 @@ tokio-test = "0.4.0" # must match the min version of the `tokio` crate above
 env_logger = "0.9.0"
 chrono = "0.4.19"
 criterion = { version = "0.3.5", features = ["stable"]}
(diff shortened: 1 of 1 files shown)


util/Cargo.toml
```diff
@@ -53,6 +53,7 @@ tokio-test = "0.4.0" # must match the min version of the `tokio` crate above
 env_logger = "0.9.0"
 chrono = "0.4.19"
 criterion = { version = "0.3.5", features = ["stable"]}
+async-global-executor = "=2.2.0" # Transitive dependency due to Criterion, fixed to 2.2.0 because 2.3.0 requires Rust 1.59.0
```
melekes, line 56:
```suggestion
async-global-executor = "<=2.2.0" # Transitive dependency due to Criterion, fixed to 2.2.0 because 2.3.0 requires Rust 1.59.0
```

?

k0nserv: Ah yes, I should use the same fix as for `time` https://github.com/webrtc-rs/webrtc/pull/266/files.

melekes (approved):

PR #288, opened 2022-09-12
Get allocations info

logist322:
Add possibility to get information about `Allocation`s with `Server::get_allocations_info` method. This method returns `HashMap<FiveTuple, AllocationInfo>`.
Now `AllocationInfo` provides `five_tuple`, `name` and `relayed_bytes`, but can be extended.


logist322:
@k0nserv a new PR is ready for your review :)

turn/src/allocation/allocation_manager.rs
```diff
@@ -41,6 +44,36 @@ impl Manager {
...
+        &self,
+        five_tuples: Option<Vec<FiveTuple>>,
+    ) -> HashMap<FiveTuple, AllocationInfo> {
+        let mut infos = HashMap::new();
+
+        let guarded = self.allocations.lock().await;
+
+        guarded.iter().for_each(|(five_tuple, alloc)| {
+            if five_tuples.is_none() || five_tuples.as_ref().unwrap().contains(five_tuple) {
+                drop(infos.insert(
```
k0nserv, line 59: Why this explicit call to `drop`? If the return value isn't bound to anything this already happens

turn/src/allocation/allocation_manager.rs
```diff
@@ -13,13 +13,15 @@ use util::Conn;
 // ManagerConfig a bag of config params for Manager.
 pub struct ManagerConfig {
     pub relay_addr_generator: Box<dyn RelayAddressGenerator + Send + Sync>,
+    pub gather_metrics: bool,
```
k0nserv, line 16: I think we should consider using a feature instead of a bool for this

turn/src/allocation/allocation_manager/allocation_manager_test.rs
```diff
@@ -375,3 +382,209 @@ async fn test_delete_allocation_by_username() -> Result<()> {
...
+    let alloc1 = m
+        .create_allocation(
+            five_tuple1.clone(),
+            Arc::clone(&turn_socket),
+            0,
+            DEFAULT_LIFETIME,
+            TextAttribute::new(ATTR_USERNAME, "user1".into()),
+        )
+        .await?;
+    alloc1.relayed_bytes.store(111, Ordering::SeqCst);
```
k0nserv, line 405: This way to test this seems kind of shallow, wouldn't it be better to actually send some data and verify it this way?

turn/src/server/mod.rs
```diff
@@ -92,6 +95,45 @@ impl Server {
...
+        &self,
+        five_tuples: Option<Vec<FiveTuple>>,
+    ) -> Result<HashMap<FiveTuple, AllocationInfo>> {
+        if let Some(five_tuples) = &five_tuples {
+            if five_tuples.is_empty() {
+                return Ok(HashMap::new());
+            }
+        }
+
+        let tx = self.command_tx.lock().await.clone();
```
k0nserv, line 119: Can we use a block expression to make the scope of this lock clearer please?

turn/src/server/mod.rs
```diff
@@ -117,6 +159,12 @@ impl Server {
                                 .await;
                             continue;
                         }
+                        Ok(Command::GetAllocationsInfo(five_tuples, tx)) => {
+                            let infos = allocation_manager.get_allocations_info(five_tuples).await;
+                            drop(tx.send(infos).await);
```
k0nserv, line 164: Ca we use `let _ = tx.send(infos).await` instead of `drop`?

turn/src/server/server_test.rs
```diff
@@ -93,15 +94,15 @@ async fn test_server_simple() -> Result<()> {
     Ok(())
 }
 
-struct VNet {
-    wan: Arc<Mutex<router::Router>>,
-    net0: Arc<net::Net>,
-    net1: Arc<net::Net>,
-    netl0: Arc<net::Net>,
-    server: Server,
+pub struct VNet {
```
k0nserv, line 97: Why these changes?

turn/src/allocation/allocation_manager.rs
logist322: You are right, it's redundant.

turn/src/allocation/allocation_manager.rs
logist322: I do not think so. [Comments say](https://github.com/logist322/webrtc/blob/98d7a5754e8022d71418d672bcfe85ac34b6f187/turn/src/server/config.rs#L32) different connections can have custom behavior: 

```
// conn_configs are a list of all the turn listeners
// Each listener can have custom behavior around the creation of Relays
pub conn_configs: Vec<ConnConfig>,
```

Do you think we should have the same settings for every **Allocation** `Manager` on the `Server` instance?

logist322: All right, so it will be a feature.

turn/src/server/mod.rs
logist322: Yes, we can. But as far as I know, it has the same behavior in this case. Can you explain this propose, please?

turn/src/allocation/allocation_manager/allocation_manager_test.rs
logist322: I think it would. I will add tests with sending real data.

turn/src/server/server_test.rs
logist322: It is not actual now, as the tests will be reworked.

turn/src/server/mod.rs
logist322: It makes sense, it will be `let _`.

tyranron: @logist322 @k0nserv there are two `clippy` lints controlling this:
- [`clippy::let_underscore_drop`](https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_drop)
- [`clippy::drop_non_drop`](https://rust-lang.github.io/rust-clippy/master/index.html#drop_non_drop)

The whole idea to squat around them exists only if the crate has `#![warn(unused_results)]` `rustc` lint, or we're dealing with `#[must_use]` function/type.

k0nserv: Right, but we don't care about the lifetime of the value returned by `tx.send(infos).await` hence `let _ = ..` to drop it immediately is fine?

turn/src/server/server_test.rs
k0nserv: Aight, please revert it if it's not gonna be used

logist322: Sure

turn/src/server/mod.rs
tyranron: @k0nserv from this prespective (borrow check) there should be no difference between `drop()`, `let _ = ..` or just `tx.send(infos).await;`. It's more a code style issue rather than semantics.

turn/src/allocation/allocation_manager.rs
```diff
@@ -41,6 +41,33 @@ impl Manager {
...
+    ) -> HashMap<FiveTuple, AllocationInfo> {
+        let mut infos = HashMap::new();
+
+        let guarded = self.allocations.lock().await;
+
+        guarded.iter().for_each(|(five_tuple, alloc)| {
+            if five_tuples.is_none() || five_tuples.as_ref().unwrap().contains(five_tuple) {
+                #[cfg(not(feature = "metrics"))]
+                let relayed_bytes: Option<usize> = None;
+                #[cfg(feature = "metrics")]
```
k0nserv, line 58: Why not just feature gate this whole method? There's no value in `AllocationInfo` if not using metrics, no?

turn/src/allocation/mod.rs
```diff
@@ -31,6 +32,32 @@ const RTP_MTU: usize = 1500;
...
+    /// [`FiveTuple`] of this [`Allocation`].
+    pub five_tuple: FiveTuple,
+
+    /// Username of this [`Allocation`].
+    pub username: String,
+
+    /// Relayed bytes with this [`Allocation`].
+    ///
+    /// If it is [`None`] it means the relayed bytes are not counted.
+    pub relayed_bytes: Option<usize>,
```
k0nserv, line 47: As above, is there are reason for this whole struct to exist if the metrics feature is not enabled? If there is I would suggest using

```rust
#[cfg(feature = "metrics")]
pub relayed_bytes: usize
```

instead of an `Option`

turn/src/server/mod.rs
```diff
@@ -78,7 +80,10 @@ impl Server {
 
     /// Deletes all existing [`crate::allocation::Allocation`]s by the provided `username`.
     pub async fn delete_allocations_by_username(&self, username: String) -> Result<()> {
-        let tx = self.command_tx.lock().await.clone();
+        let tx = {
```
k0nserv, line 83: Excellent 👌🏼

turn/src/server/mod.rs
```diff
@@ -92,6 +97,48 @@ impl Server {
...
+    /// If `five_tuples` is:
+    /// - [`None`]:               It returns information about the all
+    ///                           [`Allocation`]s.
+    /// - [`Some`] and not empty: It returns information about
+    ///                           the [`Allocation`]s associated with
+    ///                           the specified [`FiveTuples`].
+    /// - [`Some`], but empty:    It returns an empty [`HashMap`].
+    ///
+    /// [`Allocation`]: crate::allocation::Allocation
+    pub async fn get_allocations_info(
```
k0nserv, line 111: Should this whole method be featured gated per the logic above?

turn/src/allocation/mod.rs
logist322: Generally this struct exists not only for `relayed_bytes`. It also provides information about `five_tuple` and `name` which can be useful for `delete_allocations_by_username` method, for example. By the way, `AllocationInfo` can be extended to bring more information about. And `relayed_bytes` is one of many possible fields.

turn/src/allocation/allocation_manager.rs
logist322: No, it will return `AllocationInfo` with all the other information and `relayed_bytes: None` which indicates that metrics have not been gathered.

turn/src/server/mod.rs
logist322: I do think we should. I've tried to explain why in previous answers above.

turn/src/allocation/mod.rs
k0nserv: Aight that sounds fair. Instead of Option it would seem that feature gating `relayed_bytes` would be a nicer API both when using the feature and not, WDYT?

k0nserv (approved):

PR #290, opened 2022-09-12
Promote create_agent lock to top of function, to avoid race condition

efer-ms:
create_agent is called multiple times from multiple threads, however, it is not locked properly.

The code as written before:
1) Acquire lock
2) Check for early exit if agent is already in a usable state
3) Release lock
4) Build agent config
5) Acquire lock
6) Create agent and update value stored
7) Release lock

The issue is that two threads can get to step 4 at the same time, and hence the second one to complete step 6, obliterates the agent of the other.

Since building the agent config is cheap, it is cleaner to simply acquire the lock, and not release it until the function completes. So:

1) Acquire lock
2) Check for early exit if agent is already in a usable state
3) Build agent config
4) Create agent and update value stored
5) Release lock

---

An alternative is to do the early exit check once again after step 5, as:
1) Acquire lock
2) Check for early exit if agent is already in a usable state
3) Release lock
4) Build agent config
5) Acquire lock
6) Check for early exit if agent is already in a usable state
7) Create agent and update value stored
8) Release lock

But simply promoting the lock seems like a cleaner, less error prone choice.

---

We were seeing this race condition trigger about 0.1% of the time on certain setups.

### DIFF
--- a/ice/CHANGELOG.md
+++ b/ice/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Increased min verison of `log` dependency to `0.4.16`. [#250 Fix log at ^0.4.16 to make tests compile](https://github.com/webrtc-rs/webrtc/pull/250) by [@k0nserv](https://github.com/k0nserv).
 * Incresed serde's minimum version to 1.0.102 [#243 Fixes for cargo minimal-versions](https://github.com/webrtc-rs/webrtc/pull/243) contributed by [algesten](https://github.com/algesten)
+* Promote agent lock in ice_gather.rs create_agent() to top level of the function to avoid a race condition. [#290 Promote create_agent lock to top of function, to avoid race condition](https://github.com/webrtc-rs/webrtc/pull/290) contributed by [efer-ms](https://github.com/efer-ms)
 
 
 ## Prior to 0.8.0
--- a/ice/CHANGELOG.md
+++ b/ice/CHANGELOG.md
@@ -1,12 +1,12 @@
 # webrtc-ice changelog
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


efer-ms:
> 1. Can you please add an entry to `ice/CHANGELOG.md` under the "unreleased" section that explains this change You can look at prior entries to get an idea for the format
Done
> 2. Could we leave a comment somewhere in `create_agent` about this large lock scope to prevent us from introducing slow or complicated code in the critical section?
Done

Thanks for the quick turn-around.

ice/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * Increased min verison of `log` dependency to `0.4.16`. [#250 Fix log at ^0.4.16 to make tests compile](https://github.com/webrtc-rs/webrtc/pull/250) by [@k0nserv](https://github.com/k0nserv).
 * Incresed serde's minimum version to 1.0.102 [#243 Fixes for cargo minimal-versions](https://github.com/webrtc-rs/webrtc/pull/243) contributed by [algesten](https://github.com/algesten)
+* Promote agent lock in ice_gather.rs create_agent() to top level of the function to avoid a race condition. [#290 Promote create_agent lock to top of function, to avoid race condition](https://github.com/webrtc-rs/webrtc/pull/290) contributed by [efer-ms](https://github.com/efer-ms)
```
k0nserv, line 9: Move this up so it's under "unreleased" not "v.0.8.0" please :)

efer-ms: Gah. Sorry. Should it be "## v0.8.0" instead of ###? I think that's what threw me. Looked like 0.8.0 was under unreleased.

k0nserv (approved):

efer-ms:
@k0nserv , did I update the wrong CHANGELOG.md file? The file that changed is in `webrtc `(`webrtc/src/ice_transport/ice_gatherer.rs`), not under the top-level `ice` directory. So shouldn't the crate that needs to be updated be `webrtc` and not `webrtc-ice`?

Or am I missing something?

k0nserv:
Hmm, yes you are right. I'll fix it

PR #293, opened 2022-09-19
Simplify OnBufferedAmountLowFn callback API

llacqie:
With this minor changes, you can now provide `|| async {}` as OnBufferedAmountLowFn callback instead of 
`Box::new(|| Box::pin(async {}))`. Other callback APIs can be refactored like this one easily.


k0nserv (commented):
This looks really nice! Can you please add a changelog entry to `sctp/CHANGELOG.md`, under the `unreleased` heading?

k0nserv (approved):

k0nserv:
Actually, one more thing. We should  update `webrtc/CHANGELOG.md` since there's a change to the public API here. Can you sort that out @llacqie?

PR #294, opened 2022-09-19
Remove video/rtx codecs

k0nserv:
We don't actually support handling RTX on a distinct SSRC and declaring
these causes Chrome(and maybe others) to not sent retransmissions within
the existing RTP stream. This all means that even if we emit NACKs the
resulting retransmissions will simply be ignored.


melekes (approved):
👍

PR #296, opened 2022-09-19
Rework almost all callbacks

llacqie:
Almost all callbacks, whose name matches On*Fn pattern, reworked, except for those that were skipped with a comment why (check commits "Skip <callback_name>").
Callback representation reverted to `Box::new(|| async {})` because of `webrtc_ice::agent::agent_vnet_test::on_connected()` that returns boxed closure (we can't return non boxed closure).

### DIFF
--- a/webrtc/src/data_channel/mod.rs
+++ b/webrtc/src/data_channel/mod.rs
@@ -54,7 +54,7 @@ where
     }
 }
 
-// TODO: Rework
+// TODO: Decide what to do with FnOnce
 pub type OnOpenHdlrFn =
     Box<dyn (FnOnce() -> Pin<Box<dyn Future<Output = ()> + Send + 'static>>) + Send + Sync>;
 
--- a/webrtc/src/data_channel/mod.rs
+++ b/webrtc/src/data_channel/mod.rs
@@ -569,16 +569,13 @@ impl RTCDataChannel {
     /// on_buffered_amount_low sets an event handler which is invoked when
     /// the number of bytes of outgoing data becomes lower than the
     /// buffered_amount_low_threshold.
-    pub async fn on_buffered_amount_low<F>(&self, f: F)
-    where
-        F: OnBufferedAmountLowFn + 'static,
-    {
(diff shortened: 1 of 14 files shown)


webrtc/src/data_channel/mod.rs
```diff
@@ -535,16 +569,13 @@ impl RTCDataChannel {
     /// on_buffered_amount_low sets an event handler which is invoked when
     /// the number of bytes of outgoing data becomes lower than the
     /// buffered_amount_low_threshold.
-    pub async fn on_buffered_amount_low<F>(&self, f: F)
-    where
-        F: OnBufferedAmountLowFn + 'static,
-    {
+    pub async fn on_buffered_amount_low(&self, f: Box<dyn OnBufferedAmountLowFn>) {
```
k0nserv, line 572: Why did we have to revert back to the Box variant here?

k0nserv (commented):
Can you talk a bit more about the examples that can't be reworkeed, it would also be good to remove the commented out code.

I'm curious about still using Boxed arguments for functions that set these handlers, we can't make them generic?

k0nserv:
@KillingSpark, @MarinPostma, @pascualex, and @jeremy-prater please also weigh in about this change. Does it help the ergonomics do you think?

webrtc/src/data_channel/mod.rs
llacqie: Because I would like to switch to generics in all callback setters in one pull request, now it's not possible

llacqie:
> Can you talk a bit more about the examples that can't be reworkeed, it would also be good to remove the commented out code.

I will remove unnecessary comments before the merge into the master.

k0nserv:
Aight, let's merge to unblock you for now.

MarinPostma:
This looks promising! what's the rational for dynamic dispatch over static dispatch?

k0nserv:
Static dispatch infects everything and makes the code super unergonomic. All of these callbacks are called fairly infrequently so the cost of dynamic dispatch is negligible

melekes (commented):
lgtm

jeremy-prater:
Thanks :+1:

PR #300, opened 2022-09-23
[ice-restart example]: directory for index.html uses examples/examples

ericfont:
To fix a 404 not found error when running the example.


k0nserv:
Thanks, for fixing this. I found another case and fixed it on `master`.

PR #302, opened 2022-09-29
Remove reliance on part tail for sampler builder

k0nserv:
This reliance caused problems where Sample's would incorrectly report
dropped packets because Chrome sends empty packets with the same
timestamp after the packet with the RTP marker.

By only relying on timestamp we correctly merge these empty packets into
the Sample and don't report dropped packets when none were dropped. We
also guarantee that we can always calculate a duration for a Sample.


lolgesten (approved):

k0nserv:
This was a bad idea

PR #303, opened 2022-09-29
Handle empty RTP packets in Sample Builder

k0nserv:
libWebRTC, among others, will sometime send padding RTP packets to keep send rates steady. When they do the packets are sent after a complete sequence of RTP packets that form a Sample/GOP/Frame. Example:

* P0, Timestamp=1, Start
* P1, T=1
* P2, T=1, End(RTP Marker Bit set)
* P3, T=1, Padding
* P4, T=1, Padding
* P5, T=2, Start
* P6, T=2 End

Here, when building a sample using the packets P5 and P6 we'll have `prev_dropped_packets = 2`, however these dropped packets are inconsequential as they don't carry media. The stream of samples is unbroken, despite having dropped some packets. With the introduction of `prev_padding_packets` it's possible to distinguish between dropped packets that are likely to break the sample stream and these unimportant padding packets

### DIFF
--- a/media/src/io/sample_builder/mod.rs
+++ b/media/src/io/sample_builder/mod.rs
@@ -41,6 +41,10 @@ pub struct SampleBuilder<T: Depacketizer> {
 
     /// number of packets forced to be dropped
     dropped_packets: u16,
+
+    /// number of padding packets detected and dropped. This number will be a subset of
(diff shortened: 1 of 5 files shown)


lolgesten (approved):
Me gusta

algesten (approved):

PR #304, opened 2022-09-30
sctp: allow partial reads

### LINKED ISSUE #273: [SCTP] Stream::read: partial read OR returning the expected size of the buffer
The problem I'm facing right now is where I'm passing a buffer to Stream::read, which is not big enough (Error::ErrShortBuffer). Note there's no indication of the expected size => you're forced to guess the number.

Also note it's different from TcpStream https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read, which partially reads the data and does not put any restrictions on the size of the buffer (i.e. you, the caller, is in control of how fast you're consuming data).

I can see that [the code in Pion](https://github.com/pion/sctp/blob/d0b7cf3963e99337bdf09e11989cf8b0916e1179/reassembly_queue.go#L281-L290) is identical to one here, but I nonetheless think we need to change something.

Either

1. switch to "tcp stream" partial reading model
2. indicate the expected buffer size by changing ErrShortBuffer to be ErrShortBuffer { min: usize }

(1) is more "rust" therefore preferred.

There's also talk about auto-tuning buffer size, which is slightly related https://github.com/pion/sctp/issues/218#issuecomment-1080008490

---

https://github.com/webrtc-rs/sctp/blob/ed21dae1aa7dc5f37bd40fb344fa19746b932570/src/queue/reassembly_queue.rs#L282

also, why are we subtracting n_bytes before we've actually successfully written them to client's buffer?

---

Migrated from https://github.com/webrtc-rs/sctp/issues/28

melekes:
This PR modifies `read` behaviour to permit partial reads and no longer return `ErrShortBuffer` if the buffer is too small to fit a whole message. The buffer will be filled up to its size and the rest of the msg will be returned upon the next call to `read`.

Closes #273

### DIFF
--- a/sctp/CHANGELOG.md
+++ b/sctp/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Allow partial reads [\#304](https://github.com/webrtc-rs/webrtc/pull/304)
+  `read` no longer returns `ErrShortBuffer` if the buffer is too small to fit a
+  whole message. The buffer will be filled up to its size and the rest of the
+  msg will be returned upon the next call to `read`.
+
 ## v0.6.1
 
 * Increased min verison of `log` dependency to `0.4.16`. [#250 Fix log at ^0.4.16 to make tests compile](https://github.com/webrtc-rs/webrtc/pull/250) by [@k0nserv](https://github.com/k0nserv).
--- a/sctp/src/association/association_test.rs
+++ b/sctp/src/association/association_test.rs
@@ -735,16 +735,12 @@ async fn test_assoc_reliable_short_buffer() -> Result<()> {
 
     flush_buffers(&br, &a0, &a1).await;
 
+    // Verify partial reads are permitted.
     let mut buf = vec![0u8; 3];
-    let result = s1.read_sctp(&mut buf).await;
-    assert!(result.is_err(), "expected error to be io.ErrShortBuffer");
-    if let Err(err) = result {
-        assert_eq!(
-            Error::ErrShortBuffer,
-            err,
-            "expected error to be io.ErrShortBuffer"
-        );
-    }
+    let (n, ppi) = s1.read_sctp(&mut buf).await?;
+    assert_eq!(n, 3, "unexpected length of received data");
+    assert_eq!(&buf[..n], &MSG[..3], "unexpected length of received data");
+    assert_eq!(ppi, PayloadProtocolIdentifier::Binary, "unexpected ppi");
 
     {
         let q = s0.reassembly_queue.lock().await;
--- a/sctp/src/error.rs
+++ b/sctp/src/error.rs
@@ -209,8 +209,6 @@ pub enum Error {
     ErrOutboundPacketTooLarge,
     #[error("Stream closed")]
     ErrStreamClosed,
-    #[error("Short buffer to be filled")]
-    ErrShortBuffer,
     #[error("Io EOF")]
     ErrEof,
     #[error("Invalid SystemTime")]
--- a/sctp/src/queue/queue_test.rs
+++ b/sctp/src/queue/queue_test.rs
@@ -776,12 +776,20 @@ fn test_reassembly_queue_permits_partial_reads() -> Result<()> {
     assert!(complete, "the set should be complete");
     assert_eq!(10, rq.get_num_bytes(), "num bytes mismatch");
 
+    // partial read
     let mut buf = vec![0u8; 8]; // <- passing buffer too short
     let (n, ppi) = rq.read(&mut buf)?;
     assert_eq!(8, n, "should received 8 bytes");
     assert_eq!(2, rq.get_num_bytes(), "num bytes mismatch");
     assert_eq!(ppi, org_ppi, "should have valid ppi");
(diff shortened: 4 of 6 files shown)


melekes:
✅ tested with webrtc libp2p transport (all tests are passing)

k0nserv:
@melekes did you review the callsites where data channels use SCTP?

melekes:
> @melekes did you review the callsites where data channels use SCTP?

Yes

KallDrexx:
Maybe I am missing something, but this seems like a silently breaking change?

`read_data_channel()` calls `read_sctp()` and this change seems to imply that we may now get a partial read.  There's no indication that each each `read_sctp()` call is a partial or full message.

From what I can tell, WebRTC data channels are meant to be message based, not stream based, and so it seems like this like a single logical message will now trigger multiple `DataChannelMessage` events to propagate with the data split between them.

Am I missing something?

melekes:
> Maybe I am missing something, but this seems like a silently breaking change?

It is a breaking change (if the given buffer is not long enough to fit the message).

However, the size of the buffer is capped at `u16::MAX` and it's max for Chrome.

```
/// message size limit for Chromium
const DATA_CHANNEL_BUFFER_SIZE: u16 = u16::MAX;
```

So if you're using the default (i.e. [`read_loop`](https://github.com/webrtc-rs/webrtc/blob/30a8275f9837dc4c9d6a9989e815336f3fb62ff8/webrtc/src/data_channel/mod.rs#L304)), then you should be fine.

If you're calling `data_channel.read` with a smaller buffer, then the behavior is indeed different now.

>  and so it seems like this like a single logical message will now trigger multiple DataChannelMessage events to propagate with the data split between them.

It won't (because of u16::max).

> From what I can tell, WebRTC data channels are meant to be message based, not stream based,

Does this mean we can't make this change?

melekes:
After giving it a little bit of thought (and reading https://lgrahl.de/articles/demystifying-webrtc-dc-size-limit.html), I think from the Rust perspective it would still be correct to fill the buffer in case of `read` (because other Reader's do it and that's what every developer expects) BUT we should also indicate that message is not fully read somehow (maybe adding `Result<usize, bool>` boolean flag to indicate that). What do you think @k0nserv @KallDrexx ?

k0nserv:
It feels a bit dubious to mess with the message based nature of Data Channels. As it would be a departure from the specification. Is there a different way to work around the problem you have @melekes?

If you control both sides‚ message size can be known and buffers always appropriately sized

melekes:
> message size can be known

not always!

> As it would be a departure from the specification.

can you point me to the exact place?

>  Is there a different way to work around the problem you have @melekes?

yes. create a very big buffer, which is not ideal to say the least.

melekes:
I see two problems here:

- if you don't know the size of the biggest message, you will try to read it and fail (ErrShortBuffer). The message will be lost. Unless resubmitted by the remote.
- if you have two messages (1MB and 200MB in size), you are forced to always create a buffer of size 200MB, right?

With this PR, you'll never face both problems.

KallDrexx:
> With this PR, you'll never face both problems.

I could be mistaken, but while this PR as written fixes those two problems it creates new problems. 

Specifically, we send protobuf messages across data channels from the browser.  We had an issue where the message we were trying to send ended up being significantly large in some circumstances which caused the `ErrShortBuffer` error to propogate, which ended up in our code tearing down the connection and raising an error. This was good for us because we decided that the large data messages was a design flaw in our protocol and worked around it so we didn't need such large data messages (yes I understand this is not always feasible without manual splitting).

If your PR was in at the time, unless I'm missing something this would have been *significantly* harder to figure out what was going wrong, because partial reads would have triggered multiple data channel messages, which mean all code that expects a single data channel message to be a single deserializable self contained message would have been false.  We would have gotten no errors or warnings and just random circumstances of deserialization failures.  

Maybe I'm wrong and I'm just missing the code which would assemble partial reads into a single `DataChannelMessage` buffer though.

For what it's worth, the [data channels spec itself in section 6.6](https://www.rfc-editor.org/rfc/rfc8831#name-transferring-user-data-on-a) recommends a message size limit of 16K.  Specifically: 

>  As long as message interleaving is not supported, the sender SHOULD limit the maximum message size to 16 KB to avoid monopolization.

KallDrexx:
> if you have two messages (1MB and 200MB in size), you are forced to always create a buffer of size 200MB, right?

Maybe this is better solved by utilizing the `bytes` crate better.  Specifically, 

1. Make the buffer a `BytesMut`
2. set the size to whatever size we want to do each individual read at
3. Read into the `BytesMut`.  If we read to the full capacity and there's more, extend the buffer and do another read
4. If we've red less than the full capacity of the individual read buffer, consider the message complete, split and freeze the buffer and return that as the message, and retain the `BytesMut` for the next read attempt.

melekes:
> Maybe I'm wrong and I'm just missing the code which would assemble partial reads into a single DataChannelMessage buffer though.

What if do this? Would it be strictly better than throwing an error if msg is too big?

melekes:
> Read into the BytesMut. If we read to the full capacity and there's more, extend the buffer and do another read

huh? in the current implementation you'll just get `ErrShortBuffer` if you "read to the full capacity and there's more".

PR #307, opened 2022-10-03
Rework data-channels-flow-control example

llacqie:
All examples and tests should be reworked like this one to make it possible to rework inner callback representation.

Under "like this one" I mean reworking at least callbacks in examples and tests, so there is no logic outside of async blocks, only state management.

I want to merge this into master, and not into feature/rework-callbacks, because in this example old callback representation is used, it does not break anything and will not prevent the release in any way.

If I rework other examples, I will also use the old callback representation, since the examples are generally poorly written. It is too difficult to move them to a new representation at the moment, I will still have to rewrite them from scratch. So I don't want to rewrite and move them to a new representation at the same time.

### DIFF
--- a/examples/examples/data-channels-flow-control/data-channels-flow-control.rs
+++ b/examples/examples/data-channels-flow-control/data-channels-flow-control.rs
@@ -65,8 +65,6 @@ async fn create_requester() -> anyhow::Result<RTCPeerConnection> {
     let shared_dc = dc.clone();
     dc.on_open(Box::new(|| {
         Box::pin(async move {
-            println!("requester :: on_open");
-
             tokio::spawn(async move {
                 let buf = Bytes::from_static(&[0u8; 1024]);
 
@@ -111,8 +109,6 @@ async fn create_responder() -> anyhow::Result<RTCPeerConnection> {
             let get_total_bytes_received = total_bytes_received.clone();
             dc.on_open(Box::new(move || {
                 Box::pin(async {
-                    println!("responder :: on_open");
-
                     tokio::spawn(async move {
                         let start = SystemTime::now();
 
@@ -153,6 +149,8 @@ async fn create_responder() -> anyhow::Result<RTCPeerConnection> {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    env_logger::init();
+
     let requester = Arc::new(create_requester().await?);
     let responder = Arc::new(create_responder().await?);
 
@@ -196,12 +194,8 @@ async fn main() -> anyhow::Result<()> {
             let fault = fault.clone();
 
             Box::pin(async move {
-                println!("requester :: peer_connection_state_change :: {}", s);
-
                 if s == RTCPeerConnectionState::Failed {
-                    println!("{:?}", s);
-
-                    let _ = fault.try_send(());
+                    fault.send(()).await.unwrap();
                 }
             })
         }))
@@ -213,12 +207,8 @@ async fn main() -> anyhow::Result<()> {
             let fault = fault.clone();
 
             Box::pin(async move {
-                println!("responder :: peer_connection_state_change :: {}", s);
-
                 if s == RTCPeerConnectionState::Failed {
-                    println!("{:?}", s);
-
-                    let _ = fault.try_send(());
+                    fault.send(()).await.unwrap();
                 }
             })
         }))
@@ -235,24 +225,17 @@ async fn main() -> anyhow::Result<()> {
     requester.set_remote_description(resp).await?;
 
     tokio::select! {
-        _ = tokio::signal::ctrl_c() => {
-            println!("");
-        }
+        _ = tokio::signal::ctrl_c() => {}
         _ = reqs_fault.recv() => {
-            println!("reqs_fault");
+            log::error!("Requester's peer connection failed...")
         }
         _ = resp_fault.recv() => {
-            println!("resp_fault");
+            log::error!("Responder's peer connection failed...");
         }
     }
 
-    if let Err(err) = requester.close().await {
-        println!("{}", err);
-    }
-
-    if let Err(err) = responder.close().await {
-        println!("{}", err);
-    }
+    requester.close().await?;
+    responder.close().await?;
 
     println!("");
 
--- a/examples/examples/data-channels-flow-control/data-channels-flow-control.rs
+++ b/examples/examples/data-channels-flow-control/data-channels-flow-control.rs
@@ -24,15 +24,14 @@ use webrtc::{
 const BUFFERED_AMOUNT_LOW_THRESHOLD: usize = 512 * 1024; // 512 KB
 const MAX_BUFFERED_AMOUNT: usize = 1024 * 1024; // 1 MB
 
-async fn create_peer_connection() -> RTCPeerConnection {
+async fn create_peer_connection() -> anyhow::Result<RTCPeerConnection> {
     let mut media_engine = MediaEngine::default();
 
-    media_engine.register_default_codecs().unwrap();
+    media_engine.register_default_codecs()?;
 
     let mut interceptor_registry = Registry::new();
 
-    interceptor_registry =
-        register_default_interceptors(interceptor_registry, &mut media_engine).unwrap();
+    interceptor_registry = register_default_interceptors(interceptor_registry, &mut media_engine)?;
 
     let api = APIBuilder::new()
         .with_media_engine(media_engine)
@@ -48,11 +47,11 @@ async fn create_peer_connection() -> RTCPeerConnection {
         ..Default::default()
     };
 
-    api.new_peer_connection(config).await.unwrap()
+    Ok(api.new_peer_connection(config).await?)
 }
 
-async fn create_requester() -> RTCPeerConnection {
-    let pc = create_peer_connection().await;
+async fn create_requester() -> anyhow::Result<RTCPeerConnection> {
+    let pc = create_peer_connection().await?;
 
     let options = Some(RTCDataChannelInit {
         ordered: Some(false),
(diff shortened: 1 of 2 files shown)


examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
-
-async fn create_offerer() -> Result<Arc<RTCPeerConnection>> {
-    // Create a MediaEngine object to configure the supported codec
-    let mut m = MediaEngine::default();
+async fn create_peer_connection() -> RTCPeerConnection {
+    let mut media_engine = MediaEngine::default();
 
-    // Register default codecs
-    m.register_default_codecs()?;
+    media_engine.register_default_codecs().unwrap();
```
k0nserv, line 30: This is a regression, let's continue returning `Result` and using `?`

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
-        ordered,
-        max_retransmits,
+        ordered: Some(false),
+        max_retransmits: Some(0u16),
         ..Default::default()
     });
 
-    let (send_more_ch_tx, mut send_more_ch_rx) = tokio::sync::mpsc::channel::<()>(1);
-    let send_more_ch_tx = Arc::new(send_more_ch_tx);
+    let (more_can_be_sent, mut maybe_more_can_be_sent) = tokio::sync::mpsc::channel(1);
```
k0nserv, line 63: Can we use the `tx`/  `rx` convention to name these?

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
+    let dc = pc.create_data_channel("data", options).await.unwrap();
+
+    let shared_dc = dc.clone();
```
k0nserv, line 66: A neat trick when capturing values in closures/async blocks is to create a new block and shadow the captured variables with the same name, i.e.,


```rust
{
    let dc = dc.clone();
    dc.on_open(...
}
```

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
+                        let _ = maybe_more_can_be_sent.recv().await;
+                    }
                 }
-            }
-            println!("exit on_open");
-        });
-        Box::pin(async {})
+            });
+
+            task.await.unwrap();
```
k0nserv, line 87: This is not good, we don't want to block the code that calls that awaits this future on this tasks. The future returned from `on_open` must complete fairly quickly since it awaited within a core part of the code. Blocking for a long time is liable to break stuff

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
+                            println!(
+                                "Throughput is about {:.03} Mbps",
+                                bps / (1024 * 1024) as f64
+                            );
+                            tokio::time::sleep(Duration::from_secs(1)).await;
+                        }
+                    });
 
-                Box::pin(async {})
+                    task.await.unwrap();
```
k0nserv, line 139: Same point as above

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
+                            let total_bytes = get_total_bytes_received.load(Ordering::Relaxed);
 
-        Box::pin(async move {
-            // Register channel opening handling
-            let total_bytes_received2 = Arc::clone(&total_bytes_received);
-            dc.on_open(Box::new(move || {
-                println!("OnOpen: Start receiving data"); //, dc2.label(), dc2.id());
-                tokio::spawn(async move {
-                    let since = SystemTime::now();
+                            let elapsed = SystemTime::now().duration_since(start);
```
k0nserv, line 128: We should switch to using `Instant` instead of `SystemTime` to measure elapsed time. Doesn't have to happen in this PR, however

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
-            // Register the OnMessage to handle incoming messages
-            dc.on_message(Box::new(move |msg: DataChannelMessage| {
-                let n = msg.data.len();
-                total_bytes_received.fetch_add(n, Ordering::SeqCst);
-                Box::pin(async {})
+            dc.on_message(Box::new(move |msg| {
+                let total_bytes_received = total_bytes_received.clone();
+
+                Box::pin(async move {
+                    total_bytes_received.fetch_add(msg.data.len(), Ordering::Relaxed);
```
k0nserv, line 148: I'm not super clued in to the atomic Orderings, but I would like to continue using the strong ordering of SeqCst.

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
-            .on_ice_candidate(Box::new(move |c: Option<RTCIceCandidate>| {
-                let offer_pc3 = offer_pc2.clone();
-                Box::pin(async move {
-                    if let Some(c) = c {
-                        if let Ok(c) = c.to_json().await {
-                            if let Some(offer_pc3) = offer_pc3.upgrade() {
-                                let _ = offer_pc3.add_ice_candidate(c).await;
-                            }
+async fn main() {
+    let requester = Arc::new(create_requester().await);
```
k0nserv, line 161: I know the help menu here wasn't the most expansive, but the convention is that all examples have one. Please reintroduce it

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
+    let maybe_requester = Arc::downgrade(&requester);
+    responder
+        .on_ice_candidate(Box::new(move |candidate: Option<RTCIceCandidate>| {
+            let maybe_requester = maybe_requester.clone();
+
+            Box::pin(async move {
+                if let Some(candidate) = candidate {
+                    if let Ok(candidate) = candidate.to_json().await {
+                        if let Some(requester) = maybe_requester.upgrade() {
+                            requester.add_ice_candidate(candidate).await.unwrap();
```
k0nserv, line 173: Can we not use `unwrap` and instead log the error a the warn level?

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
+    let maybe_responder = Arc::downgrade(&responder);
+    requester
+        .on_ice_candidate(Box::new(move |candidate: Option<RTCIceCandidate>| {
+            let maybe_responder = maybe_responder.clone();
+
+            Box::pin(async move {
+                if let Some(candidate) = candidate {
+                    if let Ok(candidate) = candidate.to_json().await {
+                        if let Some(responder) = maybe_responder.upgrade() {
+                            responder.add_ice_candidate(candidate).await.unwrap();
```
k0nserv, line 190: Same as above

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
+            Box::pin(async move {
+                println!("requester :: peer_connection_state_change :: {}", s);
 
                 if s == RTCPeerConnectionState::Failed {
-                    // Wait until PeerConnection has had no network activity for 30 seconds or another failure. It may be reconnected using an ICE Restart.
-                    // Use webrtc.PeerConnectionStateDisconnected if you are interested in detecting faster timeout.
-                    // Note that the PeerConnection may come back from PeerConnectionStateDisconnected.
-                    println!("Peer Connection (offerer) has gone to failed exiting");
-                    let _ = offer_done_tx.try_send(());
+                    println!("{:?}", s);
```
k0nserv, line 207: This should use log if we aim to keep it

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,368 +1,263 @@
...
+    requester.set_remote_description(resp).await.unwrap();
 
-        if let Err(err) = answer_pc.close().await {
-            println!("cannot close answer_pc: {}", err);
+    tokio::select! {
+        _ = tokio::signal::ctrl_c() => {
+            println!("");
+        }
+        _ = reqs_fault.recv() => {
+            println!("reqs_fault");
```
k0nserv, line 247: These log lines need to be more informative

k0nserv (changes requested):
To be honest I'm not sure this actually changes much if you address my concern about not awaiting the spawned tasks in `on_open`. We really shouldn't be blocking in those futures I believe.

Other than this can we please remove all use of `unwrap` in favour of the question mark operator and re-introduce the comments that explain what's happening in each step?

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
llacqie: There is no need for this, it's just a counter that only increases, Ordering::Relaxed is the best Ordering in this case.

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
llacqie: It is written somewhere in the documentation that this callback should return future, that "must complete fairly quickly since it awaited within a core part of the code"? Maybe we should make this callback not async, to avoid unexpected blocking?

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
k0nserv: I didn't know that, thanks for teaching me :)

examples/examples/data-channels-flow-control/data-channels-flow-control.rs
```diff
@@ -1,200 +1,160 @@
...
     };
 
-    // Create a new RTCPeerConnection
-    let pc = Arc::new(api.new_peer_connection(config).await?);
+    Ok(api.new_peer_connection(config).await?)
+}
 
-    let ordered = Some(false);
-    let max_retransmits = Some(0u16);
+async fn create_requester() -> anyhow::Result<RTCPeerConnection> {
```
k0nserv, line 56: These names no longer matches the names in the `README.md`

k0nserv (approved):

PR #309, opened 2022-10-05
Improve Sample Builder padding handling for static streams

k0nserv:
Improve Sample Builder handling of padding packets

When sending video streams that are fairly static, such as a screen
share of an inactive screen, libWebRTC will insert padding RTP packets.
These were not correctly handled by sample builder, because of this
`SampleBuilder` would throw away packets carrying actual media.

There was also a bug where dropping packets because the sample head was
not at a partition start would double count the dropped packets(once in
`build_sample` and once in `purge_buffers`.

## Example

As an example libWebRTC will send streams like this:

```
p1(t=1, start=true)
p2(t=1)
p3(t=1)
p4(t=1, marker=true)
p5(t=1, padding=true)
p6(t=1, padding=true)
p7(t=1, padding=true)
p8(t=2, start=true, marker=true)
p9(t=2, padding=true)
p10(t=2, padding=true)
p11(t=2, padding=true)
```

Prior to this commit `SampleBuilder` would build a sample out of `p1..=p4`
then drop `p4..=p8`. This would mean the bit of media data in `p8` would
be incorrectly dropped.

The reason for dropping `p4..=p8` is that `p4` is not at a partition head
and `p8` is incorrectly identified as the end of the sample, the real
end is `p7`.

With this commit `SampleBuilder` will build sample out of `p1..=p4`,
identify `p4..=p7` as padding and drop these packets, and build a sample
for `p8`.

### DIFF
--- a/media/src/io/sample_builder/mod.rs
+++ b/media/src/io/sample_builder/mod.rs
@@ -98,7 +98,7 @@ impl<T: Depacketizer> SampleBuilder<T> {
             return false;
         }
 
-        let mut i = location.tail - 1;
+        let mut i = location.tail.wrapping_sub(1);
         while i != location.head {
             if let Some(ref packet) = self.buffer[i as usize] {
                 found_tail = Some(packet.header.timestamp);
--- a/media/src/io/sample_builder/mod.rs
+++ b/media/src/io/sample_builder/mod.rs
@@ -174,13 +174,18 @@ impl<T: Depacketizer> SampleBuilder<T> {
             if self.active.has_data() && (self.active.head == self.filled.head) {
                 // attempt to force the active packet to be consumed even though
                 // outstanding data may be pending arrival
-                if self.build_sample(true).is_some() {
-                    continue;
+                let err = match self.build_sample(true) {
+                    Ok(_) => continue,
+                    Err(e) => e,
+                };
+
+                if !matches!(err, BuildError::InvalidParition(_)) {
+                    // In the InvalidParition case `build_sample` will have already adjusted `droppped_packets`.
+                    self.dropped_packets += 1;
                 }
 
                 // could not build the sample so drop it
                 self.active.head = self.active.head.wrapping_add(1);
-                self.dropped_packets += 1;
             }
 
             self.release_packet(self.filled.head);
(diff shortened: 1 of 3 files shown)


k0nserv:
cc @niklaskorz and @jbis9051

k0nserv:
The following test case fails now: https://github.com/webrtc-rs/webrtc/blob/11b658310ca567b2c97118544ab1421c7560a36b/media/src/io/sample_builder/sample_builder_test.rs#L726-L824


I'm exploring if this is a problem or not

lolgesten (approved):
LGTM

media/src/io/sample_builder/sample_builder_test.rs
```diff
@@ -813,7 +813,7 @@ pub fn test_sample_builder() {
                 data: bytes!(4, 5),
                 duration: Duration::from_secs(2),
                 packet_timestamp: 4000,
-                prev_dropped_packets: 13,
```
k0nserv, line 816: This was actually wrong. The total number of dropped packets is `12`(5000 through 5012).

media/src/io/sample_builder/mod.rs
```diff
@@ -174,13 +174,19 @@ impl<T: Depacketizer> SampleBuilder<T> {
...
-                if self.build_sample(true).is_some() {
-                    continue;
+                let err = match self.build_sample(true) {
+                    Ok(_) => continue,
+                    Err(e) => e,
+                };
+
+                if !matches!(err, BuildError::InvalidParition(_)) {
+                    // In the InvalidParition case `build_sample` will have already adjusted `droppped_packets`.
+                    println!("Dropping packet {} due to being too old", self.active.head);
```
k0nserv, line 184: Oops :)

niklaskorz (commented):
Awesome, I will test this with my own project tomorrow :)

algesten (approved):

media/src/io/sample_builder/sample_builder_test.rs
k0nserv: Actually it was correct since it's 5000 _through_ 5012, not 5000 _to_ 5012.

I'll fix

PR #315, opened 2022-10-10
Pause track writers when direction doesn't allow sending.

algesten:
This makes sure we don't allocate sequence numbers for each write_sample/write_rtp when the track is in a direction that doesn't allow sending.

### DIFF
--- a/webrtc/src/track/track_local/track_local_static_rtp.rs
+++ b/webrtc/src/track/track_local/track_local_static_rtp.rs
@@ -11,6 +11,7 @@ pub struct TrackLocalStaticRTP {
     codec: RTCRtpCodecCapability,
     id: String,
     stream_id: String,
+    paused: Mutex<Option<Arc<AtomicBool>>>,
 }
 
 impl TrackLocalStaticRTP {
@@ -21,12 +22,20 @@ impl TrackLocalStaticRTP {
             bindings: Mutex::new(vec![]),
             id,
             stream_id,
+            paused: Mutex::new(None),
         }
     }
 
     /// codec gets the Codec of the track
     pub fn codec(&self) -> RTCRtpCodecCapability {
         self.codec.clone()
+    }
+
+    async fn is_paused(&self) -> bool {
+        let p = self.paused.lock().await;
+        p.as_ref()
+            .map(|p| p.load(Ordering::SeqCst))
+            .unwrap_or(false)
     }
 }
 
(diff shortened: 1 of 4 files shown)


webrtc/src/track/track_local/track_local_static_rtp.rs
```diff
@@ -11,6 +11,7 @@ pub struct TrackLocalStaticRTP {
     codec: RTCRtpCodecCapability,
     id: String,
     stream_id: String,
+    paused: Mutex<Option<Arc<AtomicBool>>>,
```
k0nserv, line 14: This should live inside `TrackBinding` I think. One `TrackLocalStaticRTP` can be bound to several senders(which are the things actually being paused), hence you can have a case where one sender is paused, but the other isn't it

webrtc/src/track/track_local/track_local_static_rtp.rs
```diff
@@ -117,6 +131,10 @@ impl TrackLocalWriter for TrackLocalStaticRTP {
         let mut write_errs = vec![];
         let mut pkt = p.clone();
 
+        if self.is_paused().await {
```
k0nserv, line 134: Here too, this is binding specific

webrtc/src/track/track_local/track_local_static_sample.rs
```diff
@@ -98,6 +113,8 @@ impl TrackLocal for TrackLocalStaticSample {
 
         let mut internal = self.internal.lock().await;
 
+        internal.paused = Some(t.paused.clone());
```
k0nserv, line 116: This needs to be updated in `unbind` too, right?

k0nserv (changes requested):
We need to document `TrackLocalStaticRTP::write_rtp` to indicate that writing to a paused track is likely to cause problems. Even though nothing is sent the same problem can happen where the rollover count is becomes mismatched, since the user is constructing the RTP packets and setting the sequence number themselves

webrtc/src/track/track_local/track_local_static_rtp.rs
algesten: Can it?

It seems to me the multiple `TrackBinding` is because we can send on multiple PT (codecs?) at the same time. All of them share the same `SSRC` though, which means we are still talking about the same m-line.

And the `paused` is derived by direction, which is m-line… so it seems right. Or?

algesten: Or are you saying there can be multiple call to `bind` for the same `TrackLocal`?

k0nserv: I think so, but let me double check that my understanding is correct

lolgesten: Ok. I can see that. In that case we only need the pause in the TrackBinding, since the `TrackLocalSample` has an internal such anyway.

algesten:
This is a bad idea.

One TrackLocal can (probably) be bound to multiple tracks. That in itself breaks setting m-lines to `inactive`, regardless of this PR however.

PR #320, opened 2022-10-12
Remove plan-b support

algesten:
M96 was released 2021-11-15. All Chrome releases since then have been (largely) without plan-b support.

The SDP used to establish a connection in WebRTC has a non-standard dialect: Plan B. [Removal timeline][1]:

* M93: Exception thrown in Canary.
* M96: Exception thrown in Beta and Stable.

[1]: https://chromestatus.com/feature/5823036655665152


k0nserv:
While plan-b removal is a useful data-point I think what's more important is when unified support was added in each browser, since those are the versions that we'll no longer support

webrtc/src/peer_connection/policy/sdp_semantics.rs
```diff
@@ -2,7 +2,9 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 
 /// SDPSemantics determines which style of SDP offers and answers
-/// can be used
+/// can be used.
+///
+/// This is unsused, we only support UnifiedPlan.
```
melekes, line 7:
```suggestion
/// This is unsused, we only support UnifiedPlan.
#[deprecated]
```

melekes (approved):
👍

algesten:
@k0nserv AFAIK Safari and Firefox were always unified.

k0nserv:
Ah true, so it's only Chromium and various other SFU implementation that are of concern. The fact the Chromium has deprecated plan b should mean that other SFU implementation must support it now

PR #321, opened 2022-10-12
One single direction per extmap

algesten:
SDP has the syntax: `a=extmap:<id>/<direction> <uri>`

The /<direction> part is optional (and rarely seen it in the wild).

The current MediaEngine API makes it seem like it's possible to provide many directions for each map (it takes a Vec). This commit changes that sov only single optional direction can be provided.

This simplifies internal matching logic and fixes a bug where setting the transceiver direction `inactive`, would (incorrectly) result in no extmaps matching for the m-line.

### DIFF
--- a/webrtc/src/api/media_engine/mod.rs
+++ b/webrtc/src/api/media_engine/mod.rs
@@ -334,8 +334,12 @@ impl MediaEngine {
         }
     }
 
-    /// register_header_extension adds a header extension to the MediaEngine
-    /// To determine the negotiated value use [`get_header_extension_id`] after signaling is complete
+    /// Adds a header extension to the MediaEngine
+    /// To determine the negotiated value use [`get_header_extension_id`] after signaling is complete.
+    ///
+    /// The `allowed_direction` controls for which transceiver directions the extension matches. If
+    /// set to `None` it matches all directions. The `SendRecv` direction would match all transceiver
+    /// directions apart from `Inactive`. Inactive ony matches inactive.
     pub fn register_header_extension(
         &mut self,
         extension: RTCRtpHeaderExtensionCapability,
--- a/webrtc/src/api/media_engine/mod.rs
+++ b/webrtc/src/api/media_engine/mod.rs
@@ -11,9 +11,7 @@ use crate::rtp_transceiver::rtp_codec::{
     RTCRtpHeaderExtensionCapability, RTCRtpHeaderExtensionParameters, RTCRtpParameters,
     RTPCodecType,
 };
-use crate::rtp_transceiver::rtp_transceiver_direction::{
-    have_rtp_transceiver_direction_intersection, RTCRtpTransceiverDirection,
-};
+use crate::rtp_transceiver::rtp_transceiver_direction::RTCRtpTransceiverDirection;
 use crate::rtp_transceiver::{PayloadType, RTCPFeedback};
 use crate::stats::stats_collector::StatsCollector;
 use crate::stats::CodecStats;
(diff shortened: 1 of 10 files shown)


webrtc/src/api/media_engine/mod.rs
```diff
@@ -58,8 +56,21 @@ pub(crate) struct MediaEngineHeaderExtension {
...
-    pub(crate) allowed_directions: Vec<RTCRtpTransceiverDirection>,
+    pub(crate) allowed_direction: Option<RTCRtpTransceiverDirection>,
+}
+
+impl MediaEngineHeaderExtension {
+    pub fn is_matching_direction(&self, dir: RTCRtpTransceiverDirection) -> bool {
+        if let Some(allowed_direction) = self.allowed_direction {
+            use RTCRtpTransceiverDirection::*;
+            allowed_direction == Inactive && dir == Inactive
+                || allowed_direction.has_send() && dir.has_send()
```
k0nserv, line 67: So this means `allowed_direction: SendRecv` matches all of `SendRecv`, `SendOnly`, and `RecvOnly`. That might be a bit unexpected and should probably be documented?

webrtc/src/api/media_engine/mod.rs
```diff
@@ -559,7 +559,7 @@ impl MediaEngine {
                         uri: extension.to_owned(),
                         is_audio: local_extension.is_audio && typ == RTPCodecType::Audio,
                         is_video: local_extension.is_video && typ == RTPCodecType::Video,
-                        allowed_directions: local_extension.allowed_directions.clone(),
+                        allowed_direction: local_extension.allowed_direction.clone(),
```
k0nserv, line 562: `allowed_direction` is `Copy`, no? No need to `clone()`

webrtc/src/peer_connection/sdp/mod.rs
```diff
@@ -505,16 +505,8 @@ pub(crate) async fn add_transceiver_sdp(
...
-    if t.sender().await.is_some() {
-        directions.push(RTCRtpTransceiverDirection::Sendonly);
-    }
-    if t.receiver().await.is_some() {
-        directions.push(RTCRtpTransceiverDirection::Recvonly);
-    }
-
     let parameters = media_engine
-        .get_rtp_parameters_by_kind(t.kind, &directions)
+        .get_rtp_parameters_by_kind(t.kind, t.direction())
```
k0nserv, line 509: This is subtly wrong and should be different when we are generating an answer I think, see the code at line 594

webrtc/src/api/media_engine/mod.rs
algesten: That's how it worked-ish before though. Leaving the previous vector empty would have resulted in SendOnly+RecvOnly and a receiver/sender would have created both a SendOnly RecvOnly for the matching.

webrtc/src/peer_connection/sdp/mod.rs
algesten: What is the RTCRtpTransceiverDirection on transceiver then?

Is it storing the mirrored or straight dir?

algesten: I'm thinking like this:

If we've added an extension to MediaEngine as `SendOnly`
We are figuring out which extensions to use, and the transceiver is also `SendOnly` or `Sendrecv`.

This should match. I.e. unless transceiver direction is stored mirrored, none of these should be inverted.

algesten: Notice that we never communicate the direction. Even though the SDP could we always just create a directionless entry in the SDP.

```
        media = media.with_extmap(sdp::extmap::ExtMap {
            value: rtp_extension.id,
            uri: Some(ext_url),
            ..Default::default()
        });
```

webrtc/src/api/media_engine/mod.rs
k0nserv: Aight! Should we have a mini-migration guide in `webrtc/CHAGNELOG.md` to explain how to change?

k0nserv (approved):

webrtc/src/api/media_engine/mod.rs
algesten: Ok!

algesten: Done

PR #322, opened 2022-10-13
add support for the telephone-event codec

KillingSpark:
Hi,

This just adds a mime type and resolves it to the G7xxPayloader. 

Telephone event packets are just a 4 byte payload, and I could add an even simpler payloader but I think the G7xxPayloader does the right thing here, even if the name is not really fitting this usecase.

Any other places this new codec would need mentioning?

### DIFF
--- a/webrtc/CHANGELOG.md
+++ b/webrtc/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add IP filter to WebRTC `SettingEngine` [#306](https://github.com/webrtc-rs/webrtc/pull/306)
 * Stop sequence numbers from increasing in `TrackLocalStaticSample` while the bound `RTCRtpSender` have
 directions that should not send. [#316](https://github.com/webrtc-rs/webrtc/pull/316)
+* Add support for a mime type "audio/telephone-event" (rfc4733)
 
 #### Breaking changes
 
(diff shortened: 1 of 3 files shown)


k0nserv:
Is there a spec reference for `audio/telephone-event` that you can provide?

KillingSpark:
RFC for telephone event would be this: https://www.rfc-editor.org/rfc/rfc4733.html

Should I document this in the codebase and if yes, where do you think is the best place?

k0nserv:
I don't think that's needed. It should be as documented as other codecs are I think

k0nserv:
Can you add an entry to `webrtc/CHANGELOG.md` under `unreleased`?

webrtc/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Add IP filter to WebRTC `SettingEngine` [#306](https://github.com/webrtc-rs/webrtc/pull/306)
 * Stop sequence numbers from increasing in `TrackLocalStaticSample` while the bound `RTCRtpSender` have
 directions that should not send. [#316](https://github.com/webrtc-rs/webrtc/pull/316)
+* Add support for a mime type "audio/telephone-event" (rfc4733)
```
k0nserv, line 10: Sorry to nitpic, but can you link to the PR(and credit yourself if you wish) as the above lines do?

KillingSpark: Should have thought that a bit :D

melekes (approved):

PR #324, opened 2022-10-13
Don't generate empty TWCC feedback packets

k0nserv:
The previous version of this code would very often generate a feedback
packet with no chunks. These packets would be ignored by libWebRTC which
logs an error message saying:

```
Buffer too small (16 bytes) to fit a FeedbackPacket. Minimum size = 18
```

When receiving such a packet.

It also felt weird to clear the received packets when no chunks were
generated for those packets.


interceptor/src/twcc/mod.rs
```diff
@@ -57,11 +57,10 @@ impl Recorder {
     /// build_feedback_packet creates a new RTCP packet containing a TWCC feedback report.
     pub fn build_feedback_packet(&mut self) -> Vec<Box<dyn rtcp::packet::Packet + Send + Sync>> {
         let mut feedback = Feedback::new(self.sender_ssrc, self.media_ssrc, self.fb_pkt_cnt);
```
melekes, line 59: should we move this down (after if condition)?

melekes (commented):
Should we raise the same issue in Pion (https://github.com/pion/interceptor/blob/97f0f8d6ca3cb7d0216ae5d557ddcde8ae4528ae/pkg/twcc/twcc.go#L76-L77)?

k0nserv:
> Should we raise the same issue in Pion

Yup, https://github.com/pion/interceptor/issues/159 :)

interceptor/src/twcc/mod.rs
k0nserv: Indeed we should

melekes (approved):

Issue #232, opened 2022-08-08
Udp connection not close when using TURN (reopen #195)


shiqifeng2000:
hi

This is a bug similar to bug 195, the out come is udp port not closing.

It seems when I try to deploy a TURN server, the candidate would use the `related_address` to communicate with the TURN server.  But when the peer connection is dropping and all candidate are dropped, it seems the `related_address` is still trying to listen to the opened port. 

Everytime I opened a new peerconnection, a new port shall be unreleased. It is a sure to come issue, could you please take a look into it?

<img width="965" alt="Screen Shot 2022-08-08 at 10 37 09 AM" src="https://user-images.githubusercontent.com/9416410/183327208-fc580f96-8b98-4c32-9f40-db07c229df11.png">

```
[Rolling] 2022-08-08T10:36:37.769958+08:00 - INFO - 1339329739 - Ice candidate Some(
    RTCIceCandidate {
        stats_id: "candidate:7nc6VphUbPm3n0qLtV5oRUfCXOr/auPt",
        foundation: "2842371988",
        priority: 16777215,
        address: "0.0.0.0",
        protocol: Udp,
        port: 55933,
        typ: Relay,
        component: 1,
        related_address: "0.0.0.0",
        related_port: 54677,
        tcp_type: "unspecified",
    },
)
```

shiqifeng2000:
UPDATE:
It seems I have targeted the reason, 
you see, in the submodule `TURN` crate, 
when `agent_gather` is starting to gather relay candidate by invoking `gather_candidates_relay`
a TURN client is created to listen to the remote turn socket.

In the `listen` method of the TURN client, I found a tokio task spawn to listen to that socket, without safe guarding
```
tokio::spawn(async move {
            let mut buf = vec![0u8; MAX_DATA_BUFFER_SIZE];
            loop {
                //TODO: gracefully exit loop
                let (n, from) = match conn.recv_from(&mut buf).await {
                    Ok((n, from)) => (n, from),
                    Err(err) => {
                        log::debug!("exiting read loop: {}", err);
                        break;
                    }
                };

                log::debug!("received {} bytes of udp from {}", n, from);

                if let Err(err) = ClientInternal::handle_inbound(
                    &read_ch_tx,
                    &buf[..n],
                    from,
                    &stun_serv_str,
                    &tr_map,
                    &binding_mgr,
                )
                .await
                {
                    log::debug!("exiting read loop: {}", err);
                    break;
                }
            }
        });
```
[...]

PR #330, opened 2022-10-25
[TURN] Fix server relay UDP port not release problem.

clia:
On our servers, we found that after the TURN service runs for a long time, there will be a large number of UDP ports that are not released.

We found that the reason is that after the `packet_handler` task is started, if the client no longer uses the UDP connection, `relay_socket.recv_from` will never return. So we add a 10-second timeout here to check whether the allocation has been dropped every 10 seconds.

Fixes https://github.com/webrtc-rs/webrtc/issues/232
Refs https://github.com/webrtc-rs/turn/issues/13

### DIFF
--- a/turn/src/allocation/mod.rs
+++ b/turn/src/allocation/mod.rs
@@ -328,14 +328,29 @@ impl Allocation {
             let mut buffer = vec![0u8; RTP_MTU];
 
             loop {
-                let (n, src_addr) = match relay_socket.recv_from(&mut buffer).await {
-                    Ok((n, src_addr)) => (n, src_addr),
-                    Err(_) => {
-                        if let Some(allocs) = &allocations {
-                            let mut alls = allocs.lock().await;
-                            alls.remove(&five_tuple);
+                let timeout = tokio::time::sleep(Duration::from_secs(10));
+                tokio::pin!(timeout);
+
+                let (n, src_addr) = tokio::select! {
+                    _ = &mut timeout =>{
+                        if Arc::strong_count(&relay_socket) <= 1 {
+                            log::debug!("allocation has stopped, stop packet_handler. five_tuple: {:?}", five_tuple);
+                            break;
+                        } else {
+                            continue;
                         }
-                        break;
+                    }
+                    result = relay_socket.recv_from(&mut buffer) => {
+                        match result {
+                            Ok((n, src_addr)) => (n, src_addr),
+                            Err(_) => {
+                                if let Some(allocs) = &allocations {
+                                    let mut alls = allocs.lock().await;
+                                    alls.remove(&five_tuple);
+                                }
+                                break;
+                            }
+                        }
                     }
                 };
 
--- a/turn/src/allocation/mod.rs
+++ b/turn/src/allocation/mod.rs
@@ -332,14 +332,6 @@ impl Allocation {
                 tokio::pin!(timeout);
 
                 let (n, src_addr) = tokio::select! {
-                    _ = &mut timeout =>{
-                        if Arc::strong_count(&relay_socket) <= 1 {
-                            log::debug!("allocation has stopped, stop packet_handler. five_tuple: {:?}", five_tuple);
-                            break;
-                        } else {
-                            continue;
-                        }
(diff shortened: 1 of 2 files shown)


turn/src/allocation/mod.rs
```diff
@@ -328,14 +328,29 @@ impl Allocation {
...
-                    Ok((n, src_addr)) => (n, src_addr),
-                    Err(_) => {
-                        if let Some(allocs) = &allocations {
-                            let mut alls = allocs.lock().await;
-                            alls.remove(&five_tuple);
+                let timeout = tokio::time::sleep(Duration::from_secs(10));
+                tokio::pin!(timeout);
+
+                let (n, src_addr) = tokio::select! {
+                    _ = &mut timeout =>{
```
melekes, line 335: can you format the code?

melekes (approved):
this seems reasonable to me

turn/src/allocation/mod.rs
clia: I have fixed the format.
I wonder if there is a better way to determine the allocation is closed?

k0nserv: Ideally we have something like a [CancelationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) that fires when some type is dropped or explicitly shutdown instead of relying on timeout. 

Generally for spawned tasks we want have the task stop automatically when the type relates to it is stopped or dropped

k0nserv: cc @clia in case you missed this conversation. Can you explore?

clia: I'll look into this later when I have time, if you haven't fixed it. Thanks for replies!

clia: @melekes @k0nserv I have rewritten the implementation, listening to the drop event and handle it. I have tested on my server and assumed its  effectiveness. Can you guys check these codes for me?

melekes (commented):
Thanks for your work @clia 👍 One comment I have is about `DropNotifier`.

turn/src/allocation/mod.rs
```diff
@@ -16,12 +16,16 @@ use util::sync::Mutex as SyncMutex;
 
 use util::Conn;
 
+use futures::channel::oneshot::{self, Receiver, Sender};
```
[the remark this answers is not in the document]
k0nserv: Why not `tokio_util::sync::CancellationToken` though?

turn/src/allocation/mod.rs
```diff
@@ -313,26 +319,39 @@ impl Allocation {
...
         let five_tuple = self.five_tuple;
         let relay_addr = self.relay_addr;
         let relay_socket = Arc::clone(&self.relay_socket);
         let turn_socket = Arc::clone(&self.turn_socket);
         let allocations = self.allocations.clone();
         let channel_bindings = Arc::clone(&self.channel_bindings);
         let permissions = Arc::clone(&self.permissions);
+        let (closed_tx, closed_rx) = oneshot::channel();
+        self.closed_tx = Some(closed_tx);
+        let drop_listener = DropListener::new(closed_rx);
```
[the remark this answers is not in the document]
clia: I have tested it on my server and it works.

melekes (approved):
👍

melekes:
@clia could you please add a changelog entry (CHANGELOG.md)?

turn/src/allocation/mod.rs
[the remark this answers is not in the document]
melekes: 🤷

I can rewrite this to use `CancellationToken` (perhaps after we merge?).

clia:
Changelog entry added.

k0nserv:
We can merge this I think, I do think using the `oneshot` this way is a bit weird but it'll work so 🤷🏼 

Tokio's `CancellationToken` seems more suitabl

clia:
I just don't want to introduce another dependency library.

k0nserv:
Hmm, good point. I thought it would already be somewhere in the dependency graph, but I guess it isn't

PR #332, opened 2022-10-26
Marshalled extension types derive Serialize/Deserialize

algesten:
To make them more usable for the consumer of these types.

### DIFF
--- a/rtp/Cargo.toml
+++ b/rtp/Cargo.toml
@@ -17,6 +17,7 @@ bytes = "1"
 rand = "0.8.5"
 thiserror = "1.0"
 async-trait = "0.1.56"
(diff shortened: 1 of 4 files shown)


k0nserv:
Hmm serde support is usually feature gated when you used like this, should we do the same?

algesten:
We have it straight up in `webrtc` crate. I figured it doesn't matter if a deeper crate has it too. Or?

k0nserv:
I suppose, although folks using RTP only might care. Dunno how much we care about that. 

Can you add a changelog entry too?

lolgesten:
@k0nserv let's release and see if anyone cares? :)

PR #333, opened 2022-10-28
[peer_connection] RTCCertificate: rework `from_pem` and `serialize_pem`

melekes:
which makes it possible to serialise and persist a certificate to disk. Both methods require `pem` feature to be enabled (I think having a feature is good since I don't expect many people to use persistent certificate). Previous implementation of `from_pem` was incorrect. It used a given certificate as a CA (hence the 2nd argument) instead of just using the certificate directly (see the related TODO in tests where `cert1 == cert3` should be matching).

API reflects one in Pion - https://github.com/pion/webrtc/blob/master/certificate.go#L179 (see `CertificateFromPEM` and `PEM` functions).

Also

- modify `RTCCertificate::get_fingerprints` to return `Vec<Fingerprint>`. No need for `Result`.
- refactor `dtls::crypto::Certificate`
- fixes TODO where two certificates should be equal when `from_pem` is being used.


dtls/src/crypto/mod.rs
```diff
@@ -15,101 +15,47 @@ use der_parser::{oid, oid::Oid};
...
 use ring::signature::{EcdsaKeyPair, Ed25519KeyPair, RsaKeyPair};
+use std::convert::TryFrom;
 use std::sync::Arc;
 
-#[derive(Clone, PartialEq)]
+/// A X.509 certificate(s) used to authenticate a DTLS connection.
+#[derive(Clone, PartialEq, Debug)]
 pub struct Certificate {
+    /// DER-encoded certificates.
     pub certificate: Vec<rustls::Certificate>,
```
melekes, line 25: btw, I'm still confused as to why we have a list of certificates here. I've checked all the places and we only create a single certificate.

webrtc/src/peer_connection/certificate.rs
```diff
@@ -295,18 +227,21 @@ mod test {
...
         let kp3 = KeyPair::from_pem(kp1_pem.as_str())?;
         let kp3_pem = kp3.serialize_pem();
-        let _cert3 = RTCCertificate::from_pem(cert1_pem, kp3)?;
+        let cert3 = RTCCertificate::from_pem(&cert1_pem, kp3)?;
 
         assert_eq!(kp1_pem, kp3_pem);
-        //assert!(cert1 != cert2);
-        //TODO: assert!(cert1 == cert3);
+        assert_ne!(cert1, cert2);
+        // TODO: assert!(cert1 == cert3);
```
melekes, line 241: this TODO is also super weird. these certificates can't be equal because we're using `cert1` as a CA to create `cert3`.

webrtc/src/peer_connection/certificate.rs
```diff
@@ -1,49 +1,45 @@
...
+use crate::stats::stats_collector::StatsCollector;
+use crate::stats::{CertificateStats, StatsReportType};
 
-    pem: String,
-    expires: SystemTime,
+/// Certificate represents a X.509 certificate used to authenticate WebRTC communications.
+#[derive(Clone, Debug)]
+pub struct RTCCertificate {
+    /// DTLS certificate.
+    pub dtls_certificate: dtls::crypto::Certificate,
```
melekes, line 19: The other option is to provide `to_bytes()` and `from_bytes()` methods, which will serialise/deserialise `RTCCertificate`. Let me know if you think this is better than making fields public.

webrtc/src/peer_connection/certificate.rs
melekes: fixed

webrtc/src/peer_connection/certificate.rs
melekes: done ✅

algesten (approved):
LGTM!

PR #335, opened 2022-10-31
Upgrade to `ring` 0.17

thomaseizinger:
`ring 0.17` has been released! This PR updates to that version.


melekes (commented):
thanks 👍

k0nserv:
I don't think we should hardcode a git reference of ring like this. Until a new version has been released we should keep the existing versions. Folks should still be able to use cargo's patch feature to override the `ring` version until then.

thomaseizinger:
> I don't think we should hardcode a git reference of ring like this. Until a new version has been released we should keep the existing versions.

Sorry, bad communication on my end. I don't expect this to be merged until the version is actually released!

I only opened this PR to make sure we can get this released ASAP once `ring` actually releases the new version and to be able to test that we can do what we want to do (https://github.com/melekes/rust-libp2p/pull/12) with these changes.

> Folks should still be able to use cargo's patch feature to override the `ring` version until then.

Negative. This only works if the version in the `Cargo.toml` is not bumped yet. `ring` already has it set to `0.17.0-not-released-yet` so a `[patch.crates-io]`-entry in a manifest is a no-op.

thomaseizinger:
> > I don't think we should hardcode a git reference of ring like this. Until a new version has been released we should keep the existing versions.
> 
> Sorry, bad communication on my end. I don't expect this to be merged until the version is actually released! Hence the draft PR.
> 
> I only opened this PR to make sure we can get this released ASAP once `ring` actually releases the new version and to be able to test that we can do what we want to do ([melekes/rust-libp2p#12](https://github.com/melekes/rust-libp2p/pull/12)) with these changes.

I'd appreciate a review to make sure we aren't blocked on something once `ring` cuts a new release!

k0nserv (approved):
LGTM but this needs rebasing

thomaseizinger:
> LGTM but this needs rebasing

I fixed the conflicts by merging. I hope that is fine! You seem to squash-merge in here so I didn't bother with a rebase :)

thomaseizinger:
@yngrtc This is now ready for review.

rainliu:
Could you fix clippy/fmt error?

thomaseizinger:
> Could you fix clippy/fmt error?

It seems unrelated to my changes and likely caused by an implicit dependency bump as you are not tracking `Cargo.lock` in git. Do you want it fixed in this PR or in a separate one?

thomaseizinger:
I sent a PR here: https://github.com/webrtc-rs/webrtc/pull/503

rainliu:
Rebase it?

thomaseizinger:
> Rebase it?

I am assuming because you squash-merge anyway in here that merge was fine too. Btw, there is a setting in GitHub to "always suggest updating branch" that you can enable to always have the button for manually merging `master` in PR branches!

PR #336, opened 2022-11-01
Support providing RTP extensions in local tracks

k0nserv:
- Add generic `rtp::extension::HeaderExtension`
- Add a method to write extensions for local tracks

### DIFF
--- a/webrtc/src/track/track_local/track_local_static_rtp.rs
+++ b/webrtc/src/track/track_local/track_local_static_rtp.rs
@@ -1,7 +1,9 @@
 use super::*;
 
 use crate::error::flatten_errs;
+use bytes::BytesMut;
 use tokio::sync::Mutex;
+use util::{Marshal, MarshalSize};
 
 /// TrackLocalStaticRTP  is a TrackLocal that has a pre-set codec and accepts RTP Packets.
 /// If you wish to send a media.Sample use TrackLocalStaticSample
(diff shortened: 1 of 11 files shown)


webrtc/src/track/track_local/track_local_static_rtp.rs
```diff
@@ -42,6 +44,83 @@ impl TrackLocalStaticRTP {
...
+    /// while the sender is paused. While the actual _sending_ is blocked, the receiver will
+    /// miss out when the sequence number "rolls over", which in turn will break SRTP.
+    ///
+    /// Extensions that are already configured on the packet are overwritten by extensions in
+    /// `extensions`.
+    pub async fn write_rtp_with_extensions(
+        &self,
+        p: rtp::packet::Packet,
+        extensions: &[rtp::extension::HeaderExtension],
+    ) -> Result<usize> {
```
algesten, line 64: Why is this not in `TrackLocal` trait?

webrtc/src/track/track_local/track_local_static_rtp.rs
```diff
@@ -42,6 +44,83 @@ impl TrackLocalStaticRTP {
...
+        let mut n = 0;
+        let mut write_errs = vec![];
+
+        let bindings = {
+            let bindings = self.bindings.lock().await;
+            bindings.clone()
+        };
+        let packet_iterator = RepeatNIterator::new(p, bindings.len());
```
algesten, line 72: Not quite feeling this `RepeatNIterator`. It hides the fact that we're doing `.clone()` for every iteration.

It would be clearer if we just loop over the binding and do an explicit `.clone()` for each loop.

algesten (approved):
Looks good.

I think that `&[rtp::extension::HeaderExtension]` is slightly inelegant as a parameter type because it forces the caller into a very specific way to organize the code. Say you need to write 1-2 of these values, you will either be required to allocate a fixed stack space (array) or a `Vec`.

Something like `impl Iterator` could be better…. or a builder pattern! ;)

webrtc/src/track/track_local/track_local_static_rtp.rs
k0nserv: I wanted to keep the change small and not have to deal with changing the additional places that implement `TrackLocal`. My thinking is we can always move this into `TrackLocal` in the future

webrtc/src/track/track_local/track_local_static_rtp.rs
k0nserv: The reason I did this was actually to avoid one clone. Previously we were taking a `&rtp::packet::Packet` and then always cloning it(even with a single track binding). With this I've changed it so we take it by value and clone `N - 1` times where `N` is the number of track bindings. Since the common case is to have a single track binding this means no more cloning.

An iterator felt like the most elegant way to achieve this "clone except the last time" behaviour

k0nserv:
I agree that `impl Iterator` or `impl IntoIterator` would be nice, but we need to iterate  over these extensions multiple times which makes it not suitable, e.g.:

```rust
        for p in packets {
            if let Err(err) = self
                .rtp_track
                .write_rtp_with_extensions(p, extensions)
                .await
            {
                write_errs.push(err);
            }
        }
```

I'm not sure having to allocate some fixed space on the stack is too cumbersome, I imagine that typically the number of extensions is small anyway.

k0nserv:
What did you have in mind with the builder pattern?

webrtc/src/track/track_local/track_local_static_rtp.rs
algesten: So you want to save one clone, and therefore we need the cognitive load of understanding a bespoke implemented iterator instead of spelling it out in code? Not sure I agree with that. Code is a liability, less code = good.

algesten:
> What did you have in mind with the builder pattern?

Something like: `TrackLocal.sample_writer().with_audio_level(28).with_video_orientation(VideoOrientation { front_facing: true, .. }).write_sample(sample)`

webrtc/src/track/track_local/track_local_static_rtp.rs
k0nserv: Hmm yeah fair. There's a method that does exactly this in `itertools` but I didn't want to pull in a whole dependency for just that iterator hence building it ourselves. Point taken though, let me rethink

k0nserv:
Pushed something like this in [06a47ed](https://github.com/webrtc-rs/webrtc/pull/336/commits/06a47ed68e0dc4648ac10ce979c0a5b57eb73dd7)

lolgesten (approved):
LGTM!

webrtc/src/track/track_local/track_local_static_rtp.rs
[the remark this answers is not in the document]
k0nserv: Answered above

PR #342, opened 2022-11-07
[DTLS] Use `RSA_PKCS1_1024_8192_SHA256_FOR_LEGACY_USE_ONLY` for short signatures

### LINKED ISSUE #339: [DTLS] potentially incorrect/buggy behavior of `webrtc_dtls::crypto::verify_signature`
I'm writting a WebRTC MCU "Client". After the ICE gathering finished, the MCU server initiates a DTLS connection. However, the server certificate was rejected by the DTLS module of `webrtc-rs`:

图片

The same certificate was accepted by other WebRTC clients (Firefox, for example):

图片

The certificate verify "frame", sent by server, looks like:

图片

After some debugging work, I located the following code inside `webrtc-dtls` crate:

https://github.com/webrtc-rs/webrtc/blob/de6e781ec437198ffe329d68e99b9348b7e83cc6/dtls/src/crypto/mod.rs#L279-L313

Here `verify_alg` is picked according to hash algorithm. For the SHA256 + RSA combination, `RSA_PKCS1_2048_8192_SHA256` was picked. However that algorithm expects a minimal signature length of 2048 bits, while the signature given by my server has only 1024 bits (128 bytes as shown in previous figures).

图片

图片

------

Actually, `ring` provides `RSA_PKCS1_1024_8192_SHA256_FOR_LEGACY_USE_ONLY` which accepts 1024 bit length signatures:

https://github.com/briansmith/ring/blob/83298f55aa802879a04013a130376e639f5b05b3/src/rsa/verification.rs#L73-L80

Checking the length of signature and choosing algorithm accordingly might resolve the problem. I'll make a fork and take an experiment then.

chuigda:
Will fix #339


dtls/src/crypto/mod.rs
```diff
@@ -300,7 +300,12 @@ fn verify_signature(
         SignatureAlgorithm::Rsa if hash_algorithm.hash == HashAlgorithm::Sha1 => {
             &ring::signature::RSA_PKCS1_1024_8192_SHA1_FOR_LEGACY_USE_ONLY
         }
-        SignatureAlgorithm::Rsa if hash_algorithm.hash == HashAlgorithm::Sha256 => {
+        SignatureAlgorithm::Rsa if (hash_algorithm.hash == HashAlgorithm::Sha256)
+            && (remote_key_signature.len() < 256) => {
```
k0nserv, line 304: Shouldn't this be an exact match for 128 bytes rather than `< 256`.

k0nserv: My question was more if `RSA_PKCS1_1024_8192_SHA256_FOR_LEGACY_USE_ONLY` will work with a key length that isn't exactly `1024`. Reading the [docs](https://docs.rs/ring/0.16.20/ring/signature/static.RSA_PKCS1_1024_8192_SHA256_FOR_LEGACY_USE_ONLY.html) it does say that the key length is variable(between 1024 and 8129 bits) so I guess it will. Sounds like we should keep the current conditions.

k0nserv:
I think this PR is fine to merge, but I think using 1024 bit keys should be opt in given the weakness of these keys. Can you add a setting to control this please?

chuigda:
> I think this PR is fine to merge, but I think using 1024 bit keys should be opt in given the weakness of these keys. Can you add a setting to control this please?

Okay, maybe tomorrow I'll have some time to do this.

k0nserv:
Great, please also add a CHANGELOG.md entry in `dtls/CHANGELOG.md` and `webrtc/CHANGELOG.md` when you are done

melekes:
> > I think this PR is fine to merge, but I think using 1024 bit keys should be opt in given the weakness of these keys. Can you add a setting to control this please?
> 
> Okay, maybe tomorrow I'll have some time to do this.

Friendly ping @chuigda

chuigda:
Sorry for the very late response >_<. I'll start ASAP.

chuigda:
> I think this PR is fine to merge, but I think using 1024 bit keys should be opt in given the weakness of these keys. Can you add a setting to control this please?

According to my investigation, there're (at least) 3 positions where we may add this settings
- make this a crate `feature`
- add this setting to `RTCConfiguration`
- add this setting to `SettingsEngine`

I think the third one may be the best way. Do you have any suggestion?

k0nserv:
Yes I agree, at the top level the configuration should live in the `SettingsEngine`. It also needs to be passed down to this level since the choice is made here

k0nserv:
Excellent @chuigda! Can you add entries under `unreleased` in `webrtc/CHANGELOG.md` and `dtls/CHANGELOG.md` and then we can merge.

melekes (approved):

chuigda:
> Excellent @chuigda! Can you add entries under `unreleased` in `webrtc/CHANGELOG.md` and `dtls/CHANGELOG.md` and then we can merge.

Done.

k0nserv:
Thanks a bunch Chuigda

PR #344, opened 2022-11-08
[sctp] make `write` sync

melekes:
There's no reason for it to be async because it just buffers the data in memory (actual IO is happening in a separate thread).

### DIFF
--- a/sctp/src/association/association_internal.rs
+++ b/sctp/src/association/association_internal.rs
@@ -1897,67 +1897,68 @@ impl AssociationInternal {
     async fn pop_pending_data_chunks_to_send(&mut self) -> (Vec<ChunkPayloadData>, Vec<u16>) {
         let mut chunks = vec![];
         let mut sis_to_reset = vec![]; // stream identifiers to reset
-        let is_empty = self.pending_queue.len() == 0;
-        if !is_empty {
-            // RFC 4960 sec 6.1.  Transmission of DATA Chunks
-            //   A) At any given time, the data sender MUST NOT transmit new data to
-            //      any destination transport address if its peer's rwnd indicates
-            //      that the peer has no buffer space (i.e., rwnd is 0; see Section
-            //      6.2.1).  However, regardless of the value of rwnd (including if it
-            //      is 0), the data sender can always have one DATA chunk in flight to
-            //      the receiver if allowed by cwnd (see rule B, below).
 
-            while let Some(c) = self.pending_queue.peek().await {
-                let (beginning_fragment, unordered, data_len, stream_identifier) = (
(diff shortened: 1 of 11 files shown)


sctp/src/queue/pending_queue.rs
```diff
@@ -1,19 +1,21 @@
 use crate::chunk::chunk_payload_data::ChunkPayloadData;
 
-use std::collections::VecDeque;
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use tokio::sync::Mutex;
+use std::{
```
k0nserv, line 3: We should use the locks from `util::sync` here. Might be best to add a TODO for now

sctp/src/association/association_internal.rs
```diff
@@ -1894,70 +1894,68 @@ impl AssociationInternal {
...
-                        .pending_queue
-                        .pop(beginning_fragment, unordered)
-                        .await
-                        .is_none()
-                    {
-                        log::error!("failed to pop from pending queue");
-                    }
-                    continue;
-                }
+        if self.pending_queue.len() == 0 {
```
k0nserv, line 1901: Very nice 👨🏼‍🍳 👌🏼

sctp/src/association/association_internal.rs
```diff
@@ -1894,70 +1894,68 @@ impl AssociationInternal {
...
+            if data_len > self.rwnd as usize {
+                break; // no more rwnd
             }
 
-            // the data sender can always have one DATA chunk in flight to the receiver
-            if chunks.is_empty() && self.inflight_queue.is_empty() {
-                // Send zero window probe
-                if let Some(c) = self.pending_queue.peek().await {
-                    let (beginning_fragment, unordered) = (c.beginning_fragment, c.unordered);
+            self.rwnd -= data_len as u32;
```
k0nserv, line 1940: Just a thought that struck me, can this underflow? Out of an abundance of caution how would we handle this possibility pre-emptively?

k0nserv (approved):
LGTM

sctp/src/association/association_internal.rs
melekes: No, we cannot because we check that `self.rwnd` is big enough

```rust
if data_len > self.rwnd as usize {
   break; // no more rwnd
}
```

https://github.com/webrtc-rs/webrtc/blob/39e8427c5917985bcbaaae907d796b66263a0618/sctp/src/association/association_internal.rs#L1936-L1938

k0nserv: Cool 👍🏼

PR #345, opened 2022-11-09
add `PendingQueue::append`

melekes:
which appends chunks to the back of the pending queue.

Consider the following two simultaneous calls to `send_payload_data`:

```
  1. [beginning_fragment1, data2, data3, ending_fragment4]
  2. [beginning_fragment5, data6, data7, ending_fragment8]
```

If `push` is used, due to individual locks, it's possible to end up with:

```
  [beginning_fragment1, data2, beginning_fragment5, data3, data6, ending_fragment4,
     data6, data7, ending_fragment8]
```

This will result in wrong data received by the remote.

### DIFF
--- a/sctp/src/queue/pending_queue.rs
+++ b/sctp/src/queue/pending_queue.rs
@@ -25,8 +25,11 @@ impl PendingQueue {
         PendingQueue::default()
     }
 
+    /// Appends a chunk to the back of the pending queue.
     pub(crate) async fn push(&self, c: ChunkPayloadData) {
         self.n_bytes.fetch_add(c.user_data.len(), Ordering::SeqCst);
+        self.queue_len.fetch_add(1, Ordering::SeqCst);
+
         if c.unordered {
             let mut unordered_queue = self.unordered_queue.lock().await;
             unordered_queue.push_back(c);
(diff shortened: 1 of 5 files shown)


sctp/src/queue/pending_queue.rs
```diff
@@ -25,16 +25,52 @@ impl PendingQueue {
...
+        self.queue_len.fetch_add(chunks.len(), Ordering::SeqCst);
+        let unordered = chunks
+            .first()
+            .expect("chunks to not be empty because of the above check")
+            .unordered;
+
+        if unordered {
+            let mut unordered_queue = self.unordered_queue.lock().await;
+            for c in chunks {
+                assert!(c.unordered);
```
k0nserv, line 64: Can we add a message to this assert and the one below

sctp/src/queue/pending_queue.rs
```diff
@@ -25,16 +25,52 @@ impl PendingQueue {
...
+    /// # Panics
+    ///
+    /// If it's a mix of unordered and ordered chunks.
+    pub(crate) async fn append(&self, chunks: Vec<ChunkPayloadData>) {
+        if chunks.is_empty() {
+            return;
+        }
+
+        let total_user_data_len = chunks.iter().fold(0, |acc, c| acc + c.user_data.len());
+        self.n_bytes
```
k0nserv, line 53: Shouldn't we increment these numbers only after the data is added to the queue?

sctp/src/queue/pending_queue.rs
```diff
@@ -25,16 +25,52 @@ impl PendingQueue {
         PendingQueue::default()
     }
 
+    /// Appends a chunk to the back of the pending queue.
     pub(crate) async fn push(&self, c: ChunkPayloadData) {
```
k0nserv, line 29: Do we need to keep push at all?

k0nserv (approved):
Very nice ✨  
Please add a CHANGELOG.md entry

sctp/src/queue/pending_queue.rs
melekes: we're using it to send "reset" chunk. although it's possible to wrap it in a vector.

k0nserv: I guess not having `push` would encourage us to not make a similar mistake to this in the future. I'll leave it with you to make a decision

sctp/src/queue/pending_queue.rs
```diff
@@ -25,16 +25,52 @@ impl PendingQueue {
...
         }
-        self.queue_len.fetch_add(1, Ordering::SeqCst);
+    }
+
+    /// Appends chunks to the back of the pending queue.
+    ///
+    /// # Panics
+    ///
+    /// If it's a mix of unordered and ordered chunks.
+    pub(crate) async fn append(&self, chunks: Vec<ChunkPayloadData>) {
```
k0nserv, line 47: Related to the conversation about `push`, this can be `chunks: impl IntoIterator<ChunkPayloadData>` to allow for more flexibility for the caller. It would require structuring this method a bit differently however

PR #348, opened 2022-11-22
implement from and tryfrom string traits for SessionDescription

xnorpx:
Nice to have implement traits to convert to and from string

let sdp_string = String::from(sdp);
let sdp = SessionDescription::TryFrom(sdp_string).unwrap();

### DIFF
--- a/sdp/CHANGELOG.md
+++ b/sdp/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


k0nserv:
Seems reasonable. I think we should re-consider the `Marshal`/`Unmarshal` traits in favour of Rust native traits. This would be more flexible.

Can you add a changelog entry to `sdp/CHANGELOG.md`

PR #356, opened 2022-12-05
webrtc-constraints: A complete and spec-compliant `SelectSettings` algorithm implementation

regexident:
Hi @rainliu,

this PR replaces the existing and incomplete and spec-non-compliant implementation of `webrtc-media` with a complete and —most importantly!— spec-compliant, while extracting the code into a separate `webrtc-constraints` crate.

## What does it do?

The PR implements the `SelectSettings` algorithm as defined in the [Media Capture and Streams](https://www.w3.org/TR/mediacapture-streams/) W3C spec.

The API is used like this:

```rust
// A set of constraints supported by the user-agent:
let supported_constraints: MediaTrackSupportedConstraints = ...;

// A list of possible settings supported by the user-agent:
let possible_settings: Vec<MediaTrackSettings> = ...; 

// Ideal settings used for tie-breaking:
let ideal_settings: MediaTrackSettings = ...; 

// A desired tie-breaking policy (these can be fully customized!), such as "closest-to-ideal":
let policy = ClosestToIdealPolicy::new(ideal_settings, &supported_constraints);

// A set of user-provided "basic" and/or "advanced" constraints:
let constraints: MediaTrackConstraints = ...;

// Lowering of constraints:
let resolved_constraints: ResolvedMediaTrackConstraints = constraints.into_resolved();
let sanitized_constraints: SanitizedMediaTrackConstraints = constraints.into_sanitized(&supported_constraints);

// Select settings candidates:
let selected_settings_candidates: Vec<&MediaTrackSettings> = select_settings_candidates(
    &possible_settings,
    &sanitized_constraints,
    DeviceInformationExposureMode::Protected,
)?;

// Select settings:
let selected_settings: &MediaTrackSettings = policy.select_candidate(selected_settings_candidates);
```

In addition to its rich, yet safe and fool-proof Rust API it also provides support for parsing settings/constraints from their JSON representation (optional, via `serde` feature) for easy integration with existing JS-based interfaces.

## Why do we need this?

Well, my original implementation of settings/constraints (`webrtc_media::track::{settings, constraints}`) has a couple of problems:
- it is **incomplete**: it doesn't support "advanced" constraints, for example.
- it is **not spec-compliant**, which is in part due to being incomplete, but it also by ignoring some of the spec requirements
- it is **getting in the way of usage**, due to some unfortunate API decisions I had made with it
- it **does not support interfacing with JSON**-encoded settings/constraints, thus making it hard to integrate with existing web APIs

This new implementation aims to fix these! It is —to the best of my knowledge—:
- a **complete implementation** of the `SelectSettings` algorithm and corresponding types
- **spec-compliant**, afaict
- designed with **versatility** and **ergonomics** in mind
- parsing settings/constraints from **JSON is supported** (optional)

## How does it work?

The [Media Capture and Streams](https://www.w3.org/TR/mediacapture-streams/) W3C spec defines a very flexible structure for specifying media constraints. This results in somewhat complex type-definitions in Rust, since there are a lot of edge-cases to consider.

In order to allow for parsing the flexible JSON structure, while also keeping redundancy in the algorithm implementation to a minimum it implements a kind of lowering pipeline (similar to those found in modern compilers):

```
MediaTrackConstraint
└──▸ ResolvedMediaTrackConstraint
     └──▸ SanitizedMediaTrackConstraint
```

### `MediaTrackConstraint`

`MediaTrackConstraint` can express empty constraints, bare value constraints or "standard" constraints.

### `ResolvedMediaTrackConstraint`

By lowering `MediaTrackConstraint` to a `ResolvedMediaTrackConstraint` (via `into_resolved()`/`to_resolved()`, based on a `MediaTrackConstraintResolutionStrategy` of either `BareToIdeal` or `BareToExact`) any bare value constraints get resolved to their corresponding "standard" constraint representation.

### `SanitizedMediaTrackConstraint`

By lowering `ResolvedMediaTrackConstraint` to a `SanitizedMediaTrackConstraint` (via `into_sanitized()`/`to_sanitized()`, based on a set of supported constraints) any empty or unsupported constraints get dropped.

This allows for the `SelectSettings` algorithm to only deal with "standard" constraints (i.e. `SanitizedMediaTrackConstraint`), thus greatly simplifying the implementation.

## Safety

- The code does not make use of `unsafe { … }` whatsoever.
- The code handles `NaN` values without crashing.
- The code makes minimal use of `.expect()`/`.unwrap()`.
- The code is covered by hundreds of unit tests checking for edge-cases.

## Tests?

The crate comes with 240+ unit tests (making up about 20% of the total code) covering the implementation of the fitness-distance function, the select-settings algorithm, as well as the json parsing, and more …

## Why a separate crate?

1. The crate is just shy over 9000 lines of Rust (out of which ~1800 lines/~20% are for tests) and self-contained.
2. The implementation has potential use-cases unrelated to `webrtc-media` as Google's `libwebrtc` lacks an implementation of `SelectSettings`.

```
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
===============================================================================
 TOML                    1           23           18            1            4
-------------------------------------------------------------------------------
 Rust                   56         9136         7666          385         1085
 |- Markdown            28          553            0          411          142
 (Total)                           9689         7666          796         1227
===============================================================================
 Total                  57         9159         7684          386         1089
===============================================================================
```


k0nserv:
@regexident this looks great! I've pinged a link to it for people in the [Discord](https://discord.com/invite/4Ju8UHdXMs) to have a look at it. Since it's a large PR more eyes seems good. I'll try to review it in the coming week

constraints/src/property.rs
```diff
@@ -0,0 +1,168 @@
+//! Constants identifying the properties of a [`MediaStreamTrack`][media_stream_track] object,
```
k0nserv, line 1: Would it be valuable to wrap these in a new type to prevent mistakes with regular string values?

k0nserv (approved):
Great stuff @regexident. I've done a shallow pass over it, I think the best path forward is to merge it and you can keep working on it and start improving it based on usage if needed.

Can you please add a CHANGELOG.m entry in the crate? It would also be nice to have an examples or two in `examples/` that showcase typical public usage of this crate

constraints/src/property.rs
regexident: Absolutely! Fixed, by introducing `pub struct MediaTrackProperty(Cow<'static, str>);`.

This way we still allow working with unanticipated future property names via either `MediaTrackProperty::from("foo")`, `MediaTrackProperty::from("foo".to_owned())`, or the non-allocating `MediaTrackProperty::named("foo")`.

constraints/src/algorithms/fitness_distance.rs
```diff
@@ -0,0 +1,85 @@
...
+    // As specified in step 7 of the `fitness distance` algorithm:
+    // https://www.w3.org/TR/mediacapture-streams/#dfn-fitness-distance
+    //
+    // > For all positive numeric constraints […],
+    // > the fitness distance is the result of the formula
+    // >
+    // > ```
+    // > (actual == ideal) ? 0 : |actual - ideal| / max(|actual|, |ideal|)
+    // > ```
+    if actual == ideal {
```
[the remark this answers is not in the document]
regexident: Probably, yeah. The spec defines the formula as given above though, without making any mention to an epsilon or something similar.

That said, Blink does use an epsilon, too:
```cpp
const double DoubleConstraint::kConstraintEpsilon = 0.00001;

double NumericConstraintFitnessDistance(double value1, double value2) {
  if (std::fabs(value1 - value2) <= DoubleConstraint::kConstraintEpsilon)
    return 0.0;

  return std::fabs(value1 - value2) /
         std::max(std::fabs(value1), std::fabs(value2));
}
```

Fixed!

PR #362, opened 2022-12-10
[examples] Add an example of how to stream h264 from a webcam

D1plo1d:
This adds an example of how to stream from a server-side webcam. Not having experience in `v4l2`, I initially had some trouble learning how to do this so I figured an example might be useful to future developers who find themselves in a similar position.

Note the `capture-webcam-h264` dependency is my own work - it provides a common abstraction over webcam hardware differences (specifically, native h264 cameras and mjpeg-only cameras).

### DIFF
--- a/examples/examples/capture-webcam-h264/capture-webcam-h264.rs
+++ b/examples/examples/capture-webcam-h264/capture-webcam-h264.rs
@@ -1,0 +1,280 @@
+use anyhow::Result;
+use bytes::Bytes;
+use clap::{AppSettings, Arg, Command};
+use std::io::Write;
+use std::path::PathBuf;
+use std::sync::Arc;
+use tokio::sync::Notify;
+use tokio::time::Duration;
+use webrtc::api::interceptor_registry::register_default_interceptors;
+use webrtc::api::media_engine::{MediaEngine, MIME_TYPE_H264};
+use webrtc::api::setting_engine::SettingEngine;
+use webrtc::api::APIBuilder;
+use webrtc::ice::udp_network::{EphemeralUDP, UDPNetwork};
+use webrtc::ice_transport::ice_connection_state::RTCIceConnectionState;
+use webrtc::ice_transport::ice_server::RTCIceServer;
+use webrtc::interceptor::registry::Registry;
+use webrtc::media::Sample;
+use webrtc::peer_connection::configuration::RTCConfiguration;
+use webrtc::peer_connection::peer_connection_state::RTCPeerConnectionState;
+use webrtc::peer_connection::sdp::session_description::RTCSessionDescription;
+use webrtc::rtp_transceiver::rtp_codec::RTCRtpCodecCapability;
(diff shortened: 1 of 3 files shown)


examples/examples/capture-webcam-h264/capture-webcam-h264.rs
```diff
@@ -0,0 +1,280 @@
...
+    pool.spawn_pinned(move || {
+        async move {
+            // Connect to the webcam
+            let mut device = h264_webcam_stream::get_device(&device_path_clone).unwrap();
+            let mut stream = h264_webcam_stream::stream(&mut device, max_fps).unwrap();
+
+            // It is important to use a time.Ticker instead of time.Sleep because
+            // * avoids accumulating skew, just calling time.Sleep didn't compensate for the time spent parsing the data
+            // * works around latency issues with Sleep
+            let mut ticker = tokio::time::interval(Duration::from_millis(33));
```
morajabi, line 123: Why is 33 picked here, maybe a comment or a constant declaration could help clarify this

examples/examples/capture-webcam-h264/capture-webcam-h264.rs
```diff
@@ -0,0 +1,280 @@
...
+                let samples = h264_webcam_stream::openh264::nal_units(&h264_bitstream[..])
+                    .map(|nal| Sample {
+                        data: Bytes::copy_from_slice(nal),
+                        duration: Duration::from_secs(1),
+                        ..Default::default()
+                    })
+                    .collect::<Vec<_>>();
+
+                // Send the video to WebRTC clients
+                for sample in samples {
```
morajabi, line 138: Currently, you're doing 2 loops (first .map.collect) and then here. Can it be done in one loop?

examples/examples/capture-webcam-h264/capture-webcam-h264.rs
D1plo1d: I think I copied that from the play-from-disk-h264 example. AFAIK this looks to be for maintain playback speed but since the webcam will be streaming live video we won't need it here. I'll get rid of it.

examples/examples/capture-webcam-h264/capture-webcam-h264.rs
D1plo1d: Yup, good catch.

D1plo1d:
@morajabi I've pushed the issues you found but in trying to test this today I haven't been able to get the browser to display any video. I could have sworn it was working previously but I've had no luck with either my updated version or the original pull request version from a month ago.

Do you see anything here that would prevent the video from making it's way to the browser?

melekes:
@D1plo1d have you tried to run this example since your last comment?

D1plo1d:
No, sorry I've gotten pulled away by other things. I don't expect to have time to fix this any time soon.

melekes:
> No, sorry I've gotten pulled away by other things. I don't expect to have time to fix this any time soon.

Okay. If you (or anyone else) wants to pick this up, please let us know by dropping a comment below.

Issue #360, opened 2022-12-08
Data channel fails to send packets at a frequently speed when works as a rtc server and connected by a browser.


HsuJv:
Hi all,

I'm trying to test the bandwidth of the Data Channel based on [your example](https://github.com/webrtc-rs/webrtc/blob/master/examples/examples/data-channels/data-channels.rs)

I didn't change too many codes except the data channel callbacks.
I tried to send as much as possible to the client when a data channel was created (will show in the codes later).
And with a controllable delay, I tried send frequency at10Hz, 20Hz, 25Hz, 40Hz, 50Hz...etc

The problem is, when it comes to over than 40Hz, the data received in the browser mismatches the data I'm trying to send.
With the send frequency set to 40Hz (65535 bytes per send call), I've got the throughput jitter sharply and the browser receives fewer packets than I sent.
![image](https://user-images.githubusercontent.com/14556195/206388653-daf098f7-14a4-470a-b109-7c140c38f0ee.png)
The gap increases when the send frequency becomes larger.
(Note here in the picture 192.168.1.2 is my webrtc server which runs the codes and sends data)

With the send frequency set to 25Hz (65535 bytes per send call as well),  I've got the throughput smoothly and the browser receives all the data I sent.
![image](https://user-images.githubusercontent.com/14556195/206389626-d6ba7250-ddbd-4d73-aedd-1960c2e817f3.png)

I've got no idea what is going wrong, please if anyone here can help?

P.S.
Below are my code snippets if they can be helpful.

I rewrite these codes
https://github.com/webrtc-rs/webrtc/blob/20b59b7a40663c7546b105d7cd698ca4655a1fc4/examples/examples/data-channels/data-channels.rs#L117-L155
to the following
```Rust
    // Register data channel creation handling
    peer_connection.on_data_channel(Box::new(move |d: Arc<RTCDataChannel>| {
        let d_label = d.label().to_owned();
        let d_id = d.id();
        println!("New DataChannel {} {}", d_label, d_id);

        // Register channel opening handling
        Box::pin(async move {
            use tokio::time;
            let d2 = Arc::clone(&d);
            let d_label2 = d_label.clone();
            let d_id2 = d_id;

            d.on_open(Box::new(move || {
                println!("Data channel '{}'-'{}' open.", d_label2, d_id2);
                let mut buf = Vec::with_capacity(65535);
                let mut current = time::Instant::now();
                let mut pkt_num = 0_usize;
                unsafe {
[...]

PR #363, opened 2022-12-14
[SCTP] Associtation lock contention

KillingSpark:
As discussed in #360 the lock on the internal association is very contended in high send-bandwidth situations. This PR achieves two things:

1. Pull the marshalling of packets outside of the critical section thus reducing the time the lock is taken by the write loop
2. Schedule the marshalling and sending as a new tokio::task::Task which makes tokio schedule this on another thread allowing the read loop to make progress in parallel while the write loop is working on that

This in itself does not really increase the bandwidth but with improvements to the marshalling code itself gains can be head now, that were previously (at least partially) blocked by the lock contention.

It should also improve situations where both sides send a lot of data because then the write and read loops would both be very busy and fighting for the lock even more.

### DIFF
--- a/sctp/src/association/association_internal.rs
+++ b/sctp/src/association/association_internal.rs
@@ -326,20 +326,12 @@ impl AssociationInternal {
         }
 
         self.handle_chunk_end();
-
         Ok(())
     }
 
-    fn gather_data_packets_to_retransmit(&mut self, mut raw_packets: Vec<Bytes>) -> Vec<Bytes> {
-        for p in &self.get_data_packets_to_retransmit() {
-            if let Ok(raw) = p.marshal() {
-                raw_packets.push(raw);
-            } else {
-                log::warn!(
-                    "[{}] failed to serialize a DATA packet to be retransmitted",
-                    self.name
-                );
-            }
+    fn gather_data_packets_to_retransmit(&mut self, mut raw_packets: Vec<Packet>) -> Vec<Packet> {
+        for p in self.get_data_packets_to_retransmit() {
+            raw_packets.push(p);
         }
 
         raw_packets
@@ -347,8 +339,8 @@ impl AssociationInternal {
 
     async fn gather_outbound_data_and_reconfig_packets(
         &mut self,
-        mut raw_packets: Vec<Bytes>,
-    ) -> Vec<Bytes> {
+        mut raw_packets: Vec<Packet>,
+    ) -> Vec<Packet> {
         // Pop unsent data chunks from the pending queue to send as much as
         // cwnd and rwnd allow.
         let (chunks, sis_to_reset) = self.pop_pending_data_chunks_to_send();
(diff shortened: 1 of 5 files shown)


KillingSpark:
I really don't get how tokio schedules things.... I wanted to make the sending a third long running loop but that just results in the three loops occupying the same thread again... I would really like a cleaner solution than spawning a new task this frequently. Maybe someone with more tokio knowledge can help here.

But for now this seems to be an improvement performance wise :)

KillingSpark:
Converted to draft. I just tried this out at home and here the scheduling of tokio is even worse with this PR.

I think I'll need to figure out a way to get these to cooperate better, this is apparently not a real solution :(

KillingSpark:
I don't really get why this test fails now. I can get it tos succeed by limiting the semaphore to 1 permit, which is essentially the same as just awaiting the spawned task.

Maybe the test setup doesn't drive these tasks as fast as one would wish and that's why the association ends up in an unexpected state? The failure isn't always on the same check so something non-deterministic is happening...

sctp/src/association/mod.rs
```diff
@@ -498,36 +498,68 @@ impl Association {
...
+                8
+            }
+        };
+
+        let sem = Arc::new(Semaphore::new(limit));
+        while !done.load(Ordering::Relaxed) {
             //log::debug!("[{}] gather_outbound begin", name);
-            let (raw_packets, mut ok) = {
+            let (packets, continue_loop) = {
                 let mut ai = association_internal.lock().await;
```
melekes, line 519: I feel like there has to be a way for us to remove this lock completely... we just need a bounded channel or lock-free queue to add and remove items. No need to lock entire association

sctp/src/association/association_internal.rs
```diff
@@ -553,62 +515,42 @@ impl AssociationInternal {
...
 
         (raw_packets, ok)
     }
 
     /// gather_outbound gathers outgoing packets. The returned bool value set to
     /// false means the association should be closed down after the final send.
-    pub(crate) async fn gather_outbound(&mut self) -> (Vec<Bytes>, bool) {
-        let mut raw_packets = vec![];
+    pub(crate) async fn gather_outbound(&mut self) -> (Vec<Packet>, bool) {
+        let mut raw_packets = Vec::with_capacity(16);
```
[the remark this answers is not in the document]
KillingSpark: Seemed like a good middle-ground value that would fit most common scenarios I saw without wasting too much memory. At it saves a lot of reallocations because rust vecs go in steps of power of two. So collecting 3 packets would mean 3 reallocations.

sctp/src/association/mod.rs
KillingSpark: I agree, and it doesn't even need to be bounded, the amount of packets the can be in this channel would be naturally limited by the amount of bytes allowed to be in flight by the normal SCTP mechanisms.

I just didn't want to change to architecture too much.

melekes (approved):
👍

PR #364, opened 2022-12-14
[SCTP] Optimize sctp packet marshalling

KillingSpark:
As discussed in #360 the marshalling code has become a bottle neck in high bandwidth sending situations. I found two places that had a big effect on the performance, the hot path for this situation is marshalling packets with exactly one data chunk in them.

After this PR the marshalling is largely dominated by the CRC32 calculation which is... not easy to speed up.

### DIFF
--- a/sctp/src/packet.rs
+++ b/sctp/src/packet.rs
@@ -155,30 +155,30 @@ impl Packet {
         writer.put_u16(self.destination_port);
         writer.put_u32(self.verification_tag);
 
-        // Populate chunks
-        let mut raw = BytesMut::new();
-        for c in &self.chunks {
-            let chunk_raw = c.marshal()?;
-            raw.extend(chunk_raw);
+        // This is where the checksum will be written
+        let checksum_pos = writer.len();
+        writer.extend_from_slice(&[0, 0, 0, 0]);
 
-            let padding_needed = get_padding_size(raw.len());
+        // Populate chunks
+        for c in &self.chunks {
+            c.marshal_to(writer)?;
+
+            let padding_needed = get_padding_size(writer.len());
             if padding_needed != 0 {
-                raw.extend(vec![0u8; padding_needed]);
+                // padding needed is < 4 because we pad to 4
+                writer.extend_from_slice(&[0u8; 16][..padding_needed]);
             }
         }
-        let raw = raw.freeze();
 
         let hasher = Crc::<u32>::new(&CRC_32_ISCSI);
         let mut digest = hasher.digest();
         digest.update(writer);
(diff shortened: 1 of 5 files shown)


sctp/src/packet.rs
```diff
@@ -155,30 +154,29 @@ impl Packet {
...
         for c in &self.chunks {
-            let chunk_raw = c.marshal()?;
-            raw.extend(chunk_raw);
+            c.marshal_to(writer)?;
 
-            let padding_needed = get_padding_size(raw.len());
+            let padding_needed = get_padding_size(writer.len());
             if padding_needed != 0 {
-                raw.extend(vec![0u8; padding_needed]);
+                // padding needed is < 4 because we pad to 4
```
melekes, line 167:
```suggestion
                // padding needed if < 4 because we pad to 4
```

melekes (approved):
👍

KillingSpark:
Just an FYI, I implemented a faster crc32 algorithm on the merged branch in my repo just as an PoC that make the crc calucations basically disappear in the profile and increases throughput even more. I'll try to port this to the crc crate to benefit the whole ecosystem. If they don't want it I'll make an PR to at least let this crate benefit :)

sctp/src/packet.rs
```diff
@@ -155,30 +154,29 @@ impl Packet {
...
-            let chunk_raw = c.marshal()?;
-            raw.extend(chunk_raw);
+            c.marshal_to(writer)?;
 
-            let padding_needed = get_padding_size(raw.len());
+            let padding_needed = get_padding_size(writer.len());
             if padding_needed != 0 {
-                raw.extend(vec![0u8; padding_needed]);
+                // padding needed if < 4 because we pad to 4
+                writer.extend_from_slice(&[0u8; 16][..padding_needed]);
```
melekes, line 168: why do we need 16 bytes then? if max is 4

melekes: actually it's a bit dangerous in a sense that `PADDING_MULTIPLE` may change and this will throw a runtime error. why can't we do `[0u8; padding_needed]`? because it's slower?

KillingSpark: > why can't we do `[0u8; padding_needed]`

Because rust doesn't allow dynamically sized arrays on the stack right?

> actually it's a bit dangerous in a sense that `PADDING_MULTIPLE` may change and this will throw a runtime error

This is a good point, if unlikely since the RFC for SCTP will probably never change in this regard. I can just use `PADDING_MULTIPLE` as the size though. since `padding_needed` is guaranteed to be < `PADDING_MULTIPLE`

PR #365, opened 2022-12-14
[SCTP] improve payload queue push performance

KillingSpark:
As discussed in #360 Gathering packets to send is a big chunk of the work the Association::write_loop is doing while in a critical section.

This PR improves the performance of this by making the payload queue more performant to push to.

Previously this did a full mergesort (O(n*log(n))) on all the in-flight TSNs, now it does a binary search (O(log(n)))

### DIFF
--- a/sctp/src/queue/payload_queue.rs
+++ b/sctp/src/queue/payload_queue.rs
@@ -39,7 +39,17 @@ impl PayloadQueue {
         } else if sna32lt(tsn, *self.sorted.front().unwrap()) {
             self.sorted.push_front(tsn);
         } else {
-            let pos = match self.sorted.binary_search(&tsn) {
+            fn compare_tsn(a: u32, b: u32) -> std::cmp::Ordering {
+                if sna32lt(a, b) {
+                    std::cmp::Ordering::Less
+                } else {
+                    std::cmp::Ordering::Greater
+                }
+            }
+            let pos = match self
+                .sorted
+                .binary_search_by(|element| compare_tsn(*element, tsn))
+            {
                 Ok(pos) => pos,
                 Err(pos) => pos,
             };
--- a/sctp/src/queue/payload_queue.rs
+++ b/sctp/src/queue/payload_queue.rs
@@ -74,7 +74,8 @@ impl PayloadQueue {
 
     /// pop pops only if the oldest chunk's TSN matches the given TSN.
     pub(crate) fn pop(&mut self, tsn: u32) -> Option<ChunkPayloadData> {
-        if Some(tsn) == self.sorted.pop_front() {
+        if Some(&tsn) == self.sorted.front() {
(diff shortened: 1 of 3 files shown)


sctp/src/queue/payload_queue.rs
```diff
@@ -24,26 +24,27 @@ impl PayloadQueue {
...
     }
 
     pub(crate) fn push_no_check(&mut self, p: ChunkPayloadData) {
+        let tsn = p.tsn;
         self.n_bytes += p.user_data.len();
-        self.sorted.push(p.tsn);
-        self.chunk_map.insert(p.tsn, p);
+        self.chunk_map.insert(tsn, p);
         self.length.fetch_add(1, Ordering::SeqCst);
-        self.update_sorted_keys();
```
melekes, line 46: 🤦‍♂️

sctp/src/queue/payload_queue.rs
```diff
@@ -57,19 +58,14 @@ impl PayloadQueue {
...
-
+        self.push_no_check(p);
         true
     }
 
     /// pop pops only if the oldest chunk's TSN matches the given TSN.
     pub(crate) fn pop(&mut self, tsn: u32) -> Option<ChunkPayloadData> {
-        if !self.sorted.is_empty() && tsn == self.sorted[0] {
-            self.sorted.remove(0);
+        if Some(&tsn) == self.sorted.front() {
```
melekes, line 67:
```suggestion
        if Some(&tsn) == self.sorted.pop_front() {
```

sctp/src/queue/payload_queue.rs
```diff
@@ -57,19 +58,14 @@ impl PayloadQueue {
...
+            self.sorted.pop_front();
```
melekes, line 68:
```suggestion
```

melekes (commented):
❤️

KillingSpark:
Sorry I must have made a mistake, the tests started failing. I'll fix it soon

sctp/src/queue/payload_queue.rs
```diff
@@ -65,7 +75,7 @@ impl PayloadQueue {
     /// pop pops only if the oldest chunk's TSN matches the given TSN.
     pub(crate) fn pop(&mut self, tsn: u32) -> Option<ChunkPayloadData> {
         if Some(&tsn) == self.sorted.front() {
```
melekes, line 77: you can pop_front here. that way you won't need to do it later and would be able to remove assert_eq! completely.

KillingSpark: I don't think so. Doing so broke a few tests, and it wouldn't do what the signature says anymore.
Note that this is not binding `tsn` as part of a pattern but that this actually an equality check, and pop is only called if the tsns match. That tripped me up the first time I changed this part.

I should just remove the assert_eq! it doesn't do anything.

melekes (approved):

PR #367, opened 2022-12-15
[SCTP] limit the bytes in the PendingQueue by using a semaphore

KillingSpark:
As discussed in #360 the pending queue can grow indefinitly if the sender writes packets faster than the association is able to transmit them.

This PR solves this by enforcing a limit on the pendig queue. This blocks the sender until enough space is free.

### DIFF
--- a/sctp/src/queue/pending_queue.rs
+++ b/sctp/src/queue/pending_queue.rs
@@ -51,6 +51,9 @@ impl PushLimitSemaphore {
                 self.n.notified().await;
             } else {
                 *capacity = *capacity - credits;
+                if *capacity > 0 {
+                    self.n.notify_one();
+                }
                 break;
             }
         }
--- a/sctp/src/queue/pending_queue.rs
+++ b/sctp/src/queue/pending_queue.rs
@@ -155,7 +155,7 @@ impl PendingQueue {
         ordered_queue.get(0).cloned()
     }
 
-    pub(crate) async fn pop(
+    pub(crate) fn pop(
         &self,
         beginning_fragment: bool,
         unordered: bool,
--- a/sctp/src/queue/pending_queue.rs
+++ b/sctp/src/queue/pending_queue.rs
@@ -1,11 +1,9 @@
 use std::{
     collections::VecDeque,
-    sync::{
-        atomic::{AtomicBool, AtomicUsize, Ordering},
-        Condvar, Mutex,
-    },
+    sync::atomic::{AtomicBool, AtomicUsize, Ordering},
 };
 
+use tokio::sync::{Mutex, Notify};
 use util::sync::RwLock;
 
 use crate::chunk::chunk_payload_data::ChunkPayloadData;
@@ -31,35 +29,38 @@ pub(crate) struct PendingQueue {
 #[derive(Debug)]
 struct PushLimitSemaphore {
     m: Mutex<u64>,
-    c: Condvar,
+    n: Notify,
 }
 
 impl Default for PushLimitSemaphore {
     fn default() -> Self {
         Self {
             m: Mutex::new(1_000_000_000),
-            c: Condvar::new(),
+            n: Notify::new(),
         }
     }
 }
 
 impl PushLimitSemaphore {
     /// blocks until the credits where sucessfully taken
-    fn aquire(&self, credits: u64) {
-        let capacity = self.m.lock().unwrap();
-        let mut capacity = self
-            .c
-            .wait_while(capacity, |capacity| *capacity < credits)
-            .unwrap();
-        assert!(*capacity >= credits);
-        *capacity = *capacity - credits;
+    async fn aquire(&self, credits: u64) {
+        loop {
+            let mut capacity = self.m.lock().await;
+            if *capacity < credits {
+                drop(capacity);
+                self.n.notified().await;
+            } else {
+                *capacity = *capacity - credits;
+                break;
+            }
+        }
     }
 
     /// releases credits and allows them to be taken by a process calling aquire
-    fn release(&self, credits: u64) {
-        let mut capacity = self.m.lock().unwrap();
+    async fn release(&self, credits: u64) {
+        let mut capacity = self.m.lock().await;
         *capacity = *capacity + credits;
-        self.c.notify_one();
+        self.n.notify_one();
     }
 }
 
@@ -69,10 +70,10 @@ impl PendingQueue {
     }
 
     /// Appends a chunk to the back of the pending queue.
-    pub(crate) fn push(&self, c: ChunkPayloadData) {
+    pub(crate) async fn push(&self, c: ChunkPayloadData) {
         let user_data_len = c.user_data.len();
 
-        self.semaphore.aquire(user_data_len as u64);
+        self.semaphore.aquire(user_data_len as u64).await;
 
         if c.unordered {
             let mut unordered_queue = self.unordered_queue.write();
@@ -91,7 +92,7 @@ impl PendingQueue {
     /// # Panics
     ///
     /// If it's a mix of unordered and ordered chunks.
-    pub(crate) fn append(&self, chunks: Vec<ChunkPayloadData>) {
+    pub(crate) async fn append(&self, chunks: Vec<ChunkPayloadData>) {
         if chunks.is_empty() {
             return;
         }
@@ -99,7 +100,7 @@ impl PendingQueue {
         let total_user_data_len = chunks.iter().fold(0, |acc, c| acc + c.user_data.len());
         let chunks_len = chunks.len();
 
-        self.semaphore.aquire(total_user_data_len as u64);
+        self.semaphore.aquire(total_user_data_len as u64).await;
 
         let unordered = chunks
             .first()
@@ -150,7 +151,7 @@ impl PendingQueue {
         ordered_queue.get(0).cloned()
     }
 
-    pub(crate) fn pop(
+    pub(crate) async fn pop(
         &self,
         beginning_fragment: bool,
         unordered: bool,
@@ -204,7 +205,7 @@ impl PendingQueue {
             let user_data_len = p.user_data.len();
             self.n_bytes.fetch_sub(user_data_len, Ordering::SeqCst);
             self.queue_len.fetch_sub(1, Ordering::SeqCst);
-            self.semaphore.release(user_data_len as u64);
+            self.semaphore.release(user_data_len as u64).await;
         }
 
         popped
--- a/sctp/src/queue/pending_queue.rs
+++ b/sctp/src/queue/pending_queue.rs
@@ -75,7 +75,6 @@ impl PendingQueue {
         }
 
         let total_user_data_len = chunks.iter().fold(0, |acc, c| acc + c.user_data.len());
-        let chunks_len = chunks.len();
 
         let permits = self
             .semaphore
@@ -84,6 +83,29 @@ impl PendingQueue {
         // unwrap ok because we never close the semaphore unless we have dropped self
         permits.unwrap().forget();
 
+        self.append_unlimited(chunks, total_user_data_len);
+    }
+
+    pub(crate) fn try_append(&self, chunks: Vec<ChunkPayloadData>) -> bool {
+        if chunks.is_empty() {
+            return true;
+        }
+
+        let total_user_data_len = chunks.iter().fold(0, |acc, c| acc + c.user_data.len());
+
+        match self.semaphore.try_acquire_many(total_user_data_len as u32) {
+            Ok(permits) => {
+                permits.forget();
+
+                self.append_unlimited(chunks, total_user_data_len);
+                true
+            }
+            Err(_) => false,
+        }
+    }
+
+    fn append_unlimited(&self, chunks: Vec<ChunkPayloadData>, total_user_data_len: usize) {
+        let chunks_len = chunks.len();
         let unordered = chunks
             .first()
             .expect("chunks to not be empty because of the above check")
--- a/sctp/src/queue/pending_queue.rs
+++ b/sctp/src/queue/pending_queue.rs
@@ -3,7 +3,7 @@ use std::{
     sync::atomic::{AtomicBool, AtomicUsize, Ordering},
 };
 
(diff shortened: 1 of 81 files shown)


KillingSpark:
Converted to draft because this has some weird behaviour on my branch where I have all the performance PRs merged.

It causes the sender to no longer print its `Send xxxx pkts` lines but there is still throughput even long after it stops printing. I need to investigate what is happening there.

sctp/src/queue/pending_queue.rs
```diff
@@ -23,6 +27,42 @@ pub(crate) struct PendingQueue {
     unordered_is_selected: AtomicBool,
 }
 
+/// Simple diy semaphore not directly protecting a resource other than the current capacity of the Queue in bytes
+#[derive(Debug)]
+struct PushLimitSemaphore {
```
k0nserv, line 32: `tokio::sync::Sempahore` instead?

sctp/src/queue/pending_queue.rs
```diff
@@ -23,6 +27,42 @@ pub(crate) struct PendingQueue {
...
+        Self {
+            m: Mutex::new(1_000_000_000),
+            c: Condvar::new(),
+        }
+    }
+}
+
+impl PushLimitSemaphore {
+    /// blocks until the credits where sucessfully taken
+    fn aquire(&self, credits: u64) {
```
k0nserv, line 48: This is ultimately called from an async context right? It's not goodo to block in a way that doesn't yield to the tokio runtime. In fact if we manage to block all threads it's possible to dead lock the whole runtime since no forward progress an be made to release any permits.

sctp/src/queue/pending_queue.rs
KillingSpark: The stream is in sync code, I don't think using the tokio semaphore is a good choice there?

k0nserv: See comment below

sctp/src/queue/pending_queue.rs
KillingSpark: As far as I can tell the stream presents a sync interface?

KillingSpark: Hmmm nope you are right, I was fooled because write, write_sctp, and append are sync code. They are called a lot in async context.

So these should just become async and use the tokio semaphore?

k0nserv: > The tokio semaphore does not really provide the interface we need, they only allow taking/releasing one permit at a time.

It does? [add_permits](https://docs.rs/tokio/1.23.0/tokio/sync/struct.Semaphore.html#method.add_permits) and [acquire_many](https://docs.rs/tokio/1.23.0/tokio/sync/struct.Semaphore.html#method.acquire_many). `acquire_many` presumably blocks until there are enough permits, so it can act as a condvar?

KillingSpark: .... never mind. I was looking at outdated docs. I'll just update to this.

But it will still be a breaking change to public API, is that ok?

KillingSpark: Actually no it doesn't have what we need. It releases the permits as soon as the guard is dropped.

We need to aquire capacity in one place and grant capacity in another place. I don't think that is doable with the API tokio Semaphore presents

k0nserv: Hmm, you can move the permit guard around but perhaps that's a fools errand. Maybe ask in the tokio discord if they have thoughts, otherwise your lock based approach with a `tokio::sync::Notify` is appropriate

KillingSpark: I'll ask on their discord if anyone has a better idea but yeah the Notify based impl would satisfy me too.

I did not see the stange missing prints with the async version so maybe this was actually caused by blocking something... I'll keep it running for a while and see if it occurs again.

Otherwise what are your thoughts on breaking the API? Do you see a way around it?

k0nserv: I don't care much about breaking because we are still pre 1.0.0 and are finding out the way these things should work to be optimal.

It is interesting that this change would undo work by @melekes in #344 which was quite recent. That's not necessarily wrong though, but it would be good to have @melekes's thoughts too

KillingSpark:
I think the AsyncWriter implementation was and is broken with respect to the poll_flush and poll_shutdown functions.

* Flush should wait until there are no more bytes buffered or in flight (which isn't easy to do in the current implementation outside of just busy looping).
* shutdown should 
    1. stream.shutdown(Shutdown::Write) and await the result, so new writes get rejected
    2. self.poll_flush() and await the result, so nothing is buffered anymore
    3. only now return Poll::Ready(Ok(()))

I'll need your opinion on how to proceed on this. I think it's clear that this is an entire new can of worms that I'd like not to open in this PR. Secondly I don't think this PR makes the situation significantly worse than before. poll_flush is now essentially a no-op, but it didn't do the right thing before this PR anyways.

Honestly I'd recommend just deleting (or disabling) the AsyncWriter impl on the Stream. At least poll_flush() does not do what the trait says it must do. (I think the same goes for the implementation for the datachannel). I'm sorry if this sounds harsh, because it looks like a lot of work went into these, but the current stream just doesn't support a correct implementation of poll_flush. 

I thought about how to fix this but it definitely isn't easy because you need to tie the amount of buffered / inflight bytes to some kind of signalling that needs to be reset if new bytes came in before anyone could see the signal.

To summarize the options in order of my preference:
1. Delete/Disable the AsyncWrite impl for Stream and Datachannel until Stream is able to support a flush and a shutdown
2. Keep them as is, and maybe put a warning somewhere?
3. Fix them in this PR

Sorry that this PR is such a mess with regards to the AsyncWriter impl, I am learning about this stuff as I go.

melekes:
> This limit should probably be configurable by the user right?

I'd recommend to be critical about adding more config options as they tend to grow indefinitely and most users don't change defaults any way.

In this case, if we know upper bound of UDP socket, then can't we deduct the max queue len from it?

sctp/src/queue/pending_queue.rs
melekes: I don't mind bringing back `async` if it's needed. I changed it to `sync` because there were no reason for it to be `async`.

KillingSpark:
> I'd recommend to be critical about adding more config options as they tend to grow indefinitely and most users don't change defaults any way.
> 
> In this case, if we know upper bound of UDP socket, then can't we deduct the max queue len from it?

Agreed. I think finding out the upper limit for the UDP sockets is platform dependent. One can get the current sendbuf size, but I don't know if this is a good value it might be too little for optimal througput. Just playing around with the limit indicates that as little as 128kB as the upper limit seems enough to sustain maximum throughput. MacOS seems to use this value as the send buffer size for TCP according to `sysctl -n net.inet.tcp.sendspace`.

KillingSpark:
Just realized this has a bug with packets that are bigger than the limit, they will never be able to acquire enough permits.

But the append function is used with this comment `// NOTE: append is used here instead of push in order to prevent chunks interlacing.` Is this actually important? If not I'd change this to append the chunks in bundles with size <= limit / 2 so smooth operation is possible even in the presence of big packets.

melekes:
> Is this actually important?

I think if chunks interlace, the reader won't be able to reconstruct the message and will skip it entirely.

KillingSpark:
Yeah the rfc states that the fragments of one message must be in sequential TSNs.

Fixed it by serializing the access to the semaphore. This whole thing is kinda wonky now, but I hope my comments in the code explain whats going on

melekes (commented):
thanks for your work 👍 while I do agree with the general direction, this PR brings more complexity, which we should try to avoid. A possible solution is to make inner pending queues non blocking, which may require some architecture changes. However, I believe we should still try to seek a solution easier to maintain and understand for future contributors.

KillingSpark:
> A possible solution is to make inner pending queues non blocking, which may require some architecture changes.

I don't agree. The point of this PR is, that the Association might be sending slower than the application is pushing new data into it.

If this situation exists for a sustained amount of time, the data will accumulate in the pending queue. This leads to more and more memory being allocated and in the end the OOM killer will kick in.

sctp/src/stream/mod.rs
```diff
@@ -269,14 +269,44 @@ impl Stream {
...
+        let chunks = self.prepare_write(p, ppi)?;
+        self.send_payload_data(chunks).await?;
+
+        Ok(p.len())
+    }
+
+    /// Tries nonblockingly to write `p` to the DTLS connection with the given Payload Protocol Identifier.
+    ///
+    /// Returns an error if the write half of this stream is shutdown or `p` is too large.
+    pub fn try_write_sctp(&self, p: &Bytes, ppi: PayloadProtocolIdentifier) -> Result<usize> {
```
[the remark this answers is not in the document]
KillingSpark: True, but the try_ api turned out to not be necessary. It was born out of a misunderstanding on my part and I left it because I thought it was cool that the sync interface is still possible, albeit in a slightly different form. They can be deleted without affecting the main point of this PR though.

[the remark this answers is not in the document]
melekes: thinking about it more. lets provide a single interface

sctp/src/stream/mod.rs
```diff
@@ -698,60 +763,64 @@ impl AsyncRead for PollStream {
...
+        cx: &mut Context<'_>,
         buf: &[u8],
     ) -> Poll<io::Result<usize>> {
         let bytes = Bytes::copy_from_slice(buf);
-        match self.stream.write(&bytes) {
-            Ok(n) => Poll::Ready(Ok(n)),
-            Err(e) => Poll::Ready(Err(e.into())),
-        }
+
+        // This future, when polled, either succeeds in writing to the stream or registers the waker with the ressource that would block this from succeeding
```
melekes, line 771: wouldn't it be simpler to revert back to https://github.com/webrtc-rs/webrtc/pull/344?

KillingSpark: True, and done :D

KillingSpark:
>  I'd rather see simple bounded channels / queues than the current version with locks & semaphores. But I guess we can refactor later

I don't see how that would work, but if it does it would for sure be better than this weird mutex-semaphore combo.

sctp/src/stream/mod.rs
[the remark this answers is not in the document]
KillingSpark: I'll remove the try_* API tomorrow then. If a need for it comes up again in the future it's not much work just a bit of plumbing :+1:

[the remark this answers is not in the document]
KillingSpark: Done

sctp/src/queue/pending_queue.rs
```diff
@@ -23,21 +33,52 @@ pub(crate) struct PendingQueue {
...
+            permits.unwrap().forget();
+
+            if c.unordered {
+                let mut unordered_queue = self.unordered_queue.write();
+                unordered_queue.push_back(c);
+            } else {
+                let mut ordered_queue = self.ordered_queue.write();
+                ordered_queue.push_back(c);
+            }
+            drop(sem_lock);
```
melekes, line 81: are you sure we need to drop lock manually? why it's not dropped automatically?

sctp/src/queue/pending_queue.rs
```diff
@@ -49,30 +90,74 @@ impl PendingQueue {
...
+        } else {
+            let sem_lock = self.semaphore_lock.lock().await;
+            let permits = self
+                .semaphore
+                .acquire_many(total_user_data_len as u32)
+                .await;
+            // unwrap ok because we never close the semaphore unless we have dropped self
+            permits.unwrap().forget();
+            self.append_unlimited(chunks, total_user_data_len);
+            drop(sem_lock);
```
melekes, line 111: ditto

sctp/src/queue/pending_queue.rs
```diff
@@ -49,30 +90,74 @@ impl PendingQueue {
...
+                unordered_queue.push_back(chunk);
+            } else {
+                let mut ordered_queue = self.ordered_queue.write();
+                ordered_queue.push_back(chunk);
+            }
+            self.n_bytes.fetch_add(user_data_len, Ordering::SeqCst);
+            self.queue_len.fetch_add(1, Ordering::SeqCst);
+        }
+
+        drop(sem_lock);
```
melekes, line 137: ditto

sctp/src/stream/mod.rs
```diff
@@ -545,6 +565,22 @@ impl ReadFut {
     }
 }
 
+impl ShutdownFut {
+    /// Gets a mutable reference to the future stored inside `ShuttingDown(future)`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `ReadFut` variant is not `ShuttingDown`.
```
melekes, line 573:
```suggestion
    /// Panics if `ShutdownFut` variant is not `ShuttingDown`.
```

sctp/src/stream/mod.rs
```diff
@@ -545,6 +565,22 @@ impl ReadFut {
...
+    fn get_shutting_down_mut(
+        &mut self,
+    ) -> &mut Pin<Box<dyn Future<Output = std::result::Result<(), crate::error::Error>>>> {
+        match self {
+            ShutdownFut::ShuttingDown(ref mut fut) => fut,
+            _ => panic!("expected ReadFut to be Reading"),
```
melekes, line 579:
```suggestion
            _ => panic!("expected ShutdownFut to be ShuttingDown"),
```

sctp/src/queue/pending_queue.rs
KillingSpark: Nope it's not required, but for functions where the Lock is not needed the whole time I like to do it like this. I know this would be dropped and unlocked at the end of the block anyways but I think it communicates intent better.

If you don't like it, it can be removed.

KillingSpark: Especially since the construct with the sem_lock isn't required by any language level checks to actually push to the queue I think it is good to make sure the lock actually lives as long, but not longer than needed.

melekes (approved):

PR #372, opened 2022-12-18
Fix over-NACK due to NackPair lost_packets not resetting

kraigher:


kraigher:
The corresponding Pion code is here:
https://github.com/pion/rtcp/blob/fb051e52c7e07977197f542741ae912792defd77/transport_layer_nack.go#L38

k0nserv:
Very nice. So we had been over-NACKing before. This is a good result. 

Can you add a changelog entry in `interceptor/CHANGELOG.md` under unreleased please?

kraigher:
Yes I will add the entry later today.

kraigher:
I updated the requested changelogs and crate versions as well

k0nserv:
Thanks @kraigher

PR #375, opened 2022-12-20
Make RTCRtpTransceiver.mid spec-compliant

rogurotus:
The current implementation is:

```
pub struct RTCRtpTransceiver {
    mid: Mutex<String>,                           //atomic.Value
    sender: Mutex<Option<Arc<RTCRtpSender>>>,     //atomic.Value
    receiver: Mutex<Option<Arc<RTCRtpReceiver>>>, //atomic.Value
    ...
}
```

Essentially, the empty string is used instead of `None` (`Option<String>`).
This PR makes the behavior of `mid` explicitly optional, and also replaces `Mutex` with `OnceCell`.

### DIFF
--- a/webrtc/CHANGELOG.md
+++ b/webrtc/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 * Change `RTCPeerConnection::on_track` callback signature to `|track: Arc<TrackRemote>, receiver: Arc<RTCRtpReceiver>, transceiver: Arc<RTCRtpTransceiver>|` [#355](https://github.com/webrtc-rs/webrtc/pull/355).
 
+* Change `RTCPeerConnection::mid` signature to `Option<String>` [#375](https://github.com/webrtc-rs/webrtc/pull/375).
+
 ## v0.6.0
 
 * Added more stats to `RemoteInboundRTPStats` and `RemoteOutboundRTPStats` [#282](https://github.com/webrtc-rs/webrtc/pull/282) by [@k0nserv](https://github.com/k0nserv).
--- a/webrtc/CHANGELOG.md
+++ b/webrtc/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 * Change `RTCPeerConnection::on_track` callback signature to `|track: Arc<TrackRemote>, receiver: Arc<RTCRtpReceiver>, transceiver: Arc<RTCRtpTransceiver>|` [#355](https://github.com/webrtc-rs/webrtc/pull/355).
 
-* Change `RTCPeerConnection::mid` signature to `Option<String>` [#375](https://github.com/webrtc-rs/webrtc/pull/375).
+* Change `RTCPeerConnection::mid` return signature to `Option<String>` [#375](https://github.com/webrtc-rs/webrtc/pull/375).
 
 ## v0.6.0
 
--- a/webrtc/src/peer_connection/mod.rs
+++ b/webrtc/src/peer_connection/mod.rs
@@ -549,15 +549,15 @@ impl RTCPeerConnection {
                 }
                 // Step 5.4
                 if t.stopped.load(Ordering::SeqCst) {
-                    let Some(search_mid) = t.mid() else {return false;};
+                    let search_mid = match t.mid() {
+                        Some(mid) => mid,
+                        None => return false,
+                    };
 
-                    let Some(remote_desc) =
-                        &*params.current_remote_description.lock().await else {
-                            return false;
-                        };
-
-                    return get_by_mid(&search_mid, local_desc).is_some()
-                        || get_by_mid(&search_mid, remote_desc).is_some();
+                    if let Some(remote_desc) = &*params.current_remote_description.lock().await {
+                        return get_by_mid(&search_mid, local_desc).is_some()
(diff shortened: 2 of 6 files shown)


rogurotus:
ping @k0nserv

webrtc/src/peer_connection/sdp/mod.rs
```diff
@@ -855,7 +855,7 @@ pub(crate) fn get_by_mid<'a, 'b>(
 ) -> Option<&'b MediaDescription> {
     if let Some(parsed) = &desc.parsed {
         for m in &parsed.media_descriptions {
-            if let Some(mid) = m.attribute(ATTR_KEY_MID).and_then(|o| o) {
+            if let Some(mid) = m.attribute(ATTR_KEY_MID).flatten() {
```
k0nserv, line 858: Neat, didn't know about `Option::flatten`

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -1358,30 +1356,27 @@ impl PeerConnectionInternal {
...
-            let mid = match transeiver.mid().await {
-                m if !m.is_empty() => m,
-                _ => continue,
-            };
-
-            let track = match sender.track().await {
-                Some(t) => t,
-                None => continue,
-            };
+            if let Some(mid) = transeiver.mid() {
```
k0nserv, line 1359: The old code was nice because it didn't increasing nesting, maybe keep it as

```rust
let mid = match transceiver.mid() {
  m if !m.is_empty() => m,
  _ => continue,
}
```

Strictly speaking `mid` shouldn't ever be empty any-more, maybe that's an invariant we should enforce when first setting it.

Alternatively `let-else` syntax

```rust
let Some(mid) = transceiver.mid() else {
  continue;
}
```

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -1228,27 +1228,25 @@ impl PeerConnectionInternal {
...
-                let track_id = track.id().await;
-                let kind = match track.kind() {
-                    RTPCodecType::Unspecified => continue,
-                    RTPCodecType::Audio => "audio",
-                    RTPCodecType::Video => "video",
-                };
+                for track in tracks {
+                    let track_id = track.id().await;
+                    let kind = match track.kind() {
+                        RTPCodecType::Unspecified => continue,
```
k0nserv, line 1238: Ugh, I really don't like these `Unspecified` enum cases... Need to get rid of those in favour of `Option<Enum>`

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -1182,7 +1182,7 @@ impl PeerConnectionInternal {
     pub(super) async fn has_local_description_changed(&self, desc: &RTCSessionDescription) -> bool {
         let rtp_transceivers = self.rtp_transceivers.lock().await;
         for t in &*rtp_transceivers {
-            if let Some(m) = get_by_mid(t.mid().await.as_str(), desc) {
```
k0nserv, line 1185: Could use `let-else` here too

webrtc/src/peer_connection/mod.rs
```diff
@@ -547,13 +548,16 @@ impl RTCPeerConnection {
...
                 // Step 5.4
-                if t.stopped.load(Ordering::SeqCst) && !t.mid().await.is_empty() {
-                    let current_remote_description = params.current_remote_description.lock().await;
-                    if let Some(remote_desc) = &*current_remote_description {
-                        if get_by_mid(t.mid().await.as_str(), local_desc).is_some()
-                            || get_by_mid(t.mid().await.as_str(), remote_desc).is_some()
-                        {
-                            return true;
+                if t.stopped.load(Ordering::SeqCst) {
+                    if let Some(search_mid) = t.mid() {
```
k0nserv, line 552: Can we get rid of the extra nesting here?

k0nserv (commented):
Excellent change 🎉 

Just a few comments

webrtc/src/peer_connection/peer_connection_internal.rs
rogurotus: In my opinion, a separate PR is needed to refuse `Unspecified`.
If we make `kind -> Option<Enum>` it will be out of spec

k0nserv: Yeah for sure, I'm just whining :D

webrtc/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 * Change `RTCPeerConnection::on_track` callback signature to `|track: Arc<TrackRemote>, receiver: Arc<RTCRtpReceiver>, transceiver: Arc<RTCRtpTransceiver>|` [#355](https://github.com/webrtc-rs/webrtc/pull/355).
 
+* Change `RTCPeerConnection::mid` signature to `Option<String>` [#375](https://github.com/webrtc-rs/webrtc/pull/375).
```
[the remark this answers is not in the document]
k0nserv: Ah, of course. My bad 👍🏼

k0nserv (approved):
LGTM

PR #376, opened 2022-12-22
Fix datachannel on both sides.

qidu:
Make two terninals to can send message to each other.

Signed-off-by: Qidu Teric <teric@outlook.com>


melekes:
I'm sorry, but what is the problem you're trying to solve? This example wasn't working?

PR #377, opened 2022-12-26
Make RTCRtpTransceiver (Sender, Receiver) spec-compliant

rogurotus:
This PR reworked Sender and Receiver on Transceiver to comply with specification.


webrtc/src/peer_connection/mod.rs
```diff
@@ -1370,32 +1359,22 @@ impl RTCPeerConnection {
...
-                                None,
-                                local_direction,
-                                kind,
-                                vec![],
-                                Arc::clone(&self.internal.media_engine),
-                                Some(Box::new(self.internal.make_negotiation_needed_trigger())),
-                            )
-                            .await;
-
+                            let t = self
```
k0nserv, line 1369: I'm not sure we should use the public API here. The public API runs in a different context from creating a transceiver because we apply an SDP answer

webrtc/src/peer_connection/mod.rs
```diff
@@ -1676,34 +1651,21 @@ impl RTCPeerConnection {
...
-                    }
+                if !t.stopped.load(Ordering::SeqCst) && t.kind == track.kind() {
+                    let sender = t.sender().await;
+                    if sender.track().await.is_none() {
+                        if let Err(err) = sender.replace_track(Some(track)).await {
+                            let _ = sender.stop().await;
+                            return Err(err);
+                        }
 
-                    self.internal.trigger_negotiation_needed().await;
```
k0nserv, line 1704: Why did this get removed?

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -425,33 +421,20 @@ impl PeerConnectionInternal {
...
+            .map(|value| value.direction)
+            .unwrap_or(RTCRtpTransceiverDirection::Sendrecv);
 
         let t = match direction {
-            RTCRtpTransceiverDirection::Sendonly | RTCRtpTransceiverDirection::Sendrecv => {
-                let codecs = self.media_engine.get_codecs_by_kind(kind).await;
-                if codecs.is_empty() {
-                    return Err(Error::ErrNoCodecsAvailable);
-                }
-                let track = Arc::new(TrackLocalStaticSample::new(
```
k0nserv, line 446: The fact that we no longer implicitly create this track is an important breaking change. It needs to be called out in the changelog

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -425,33 +421,20 @@ impl PeerConnectionInternal {
...
-                    codecs[0].capability.clone(),
-                    math_rand_alpha(16),
-                    math_rand_alpha(16),
-                ));
-
-                self.new_transceiver_from_track(direction, track).await?
-            }
-            RTCRtpTransceiverDirection::Recvonly => {
+            RTCRtpTransceiverDirection::Sendonly
```
k0nserv, line 435: The spec doesn't mention any special handling for direction https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-addtransceiver

Even for inactive the steps are the same

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -494,40 +489,28 @@ impl PeerConnectionInternal {
             .ok_or(Error::ErrInterceptorNotBind)?;
 
         let (r, s) = match direction {
-            RTCRtpTransceiverDirection::Sendrecv => {
-                let r = Some(Arc::new(RTCRtpReceiver::new(
+            RTCRtpTransceiverDirection::Sendrecv
```
k0nserv, line 492: Same as above with regards to special handling for direction

k0nserv (commented):
Nice! I left a few comments.

Did you test this with at least some of the examples to make sure it hasn't broken anything?

webrtc/src/peer_connection/mod.rs
rogurotus: > Nice! I left a few comments.
> 
> Did you test this with at least some of the examples to make sure it hasn't broken anything?

Yes, I tested with 
cargo test,
using internal example,

and on a side project.

Maybe I didn't cover every possible case, but nothing broke

rogurotus:
> Nice! I left a few comments.
> 
> Did you test this with at least some of the examples to make sure it hasn't broken anything?

Yes, I tested with 
cargo test,
using internal example,

and on a side project.

Maybe I didn't cover every possible case, but nothing broke

k0nserv:
> Yes, I tested with
> cargo test,
> using internal example,

> and on a side project.

> Maybe I didn't cover every possible case, but nothing broke

Cool I can test on our project too after you address some of these comments

rogurotus:
I will try to check as soon as possible

webrtc/src/peer_connection/mod.rs
rogurotus: fixed

webrtc/src/peer_connection/peer_connection_internal.rs
k0nserv: @rogurotus you resolved this but didn't change the code, what's your thinking here?

rogurotus: what you mean hasn't changed.

I made it possible to create an inactive transceiver.

I'm just throwing an error

```
        if direction == RTCRtpTransceiverDirection::Unspecified {
            return Err(Error::ErrPeerConnAddTransceiverFromTrackSupport);
        }
```

Maybe I somehow formatted it wrong, but everything is in commit `fix`

k0nserv: Hmm, I must've been misreading then. I see it in the `fix` commit.

k0nserv:
Can you rebase or merge again and then we can merge?

k0nserv (approved):

k0nserv:
Thanks for this @rogurotus 🎉

Issue #317, opened 2022-10-11
Move outbound sequence number generation to transport layer


k0nserv:
Currently sequence numbers for outbound RTP packets are created internally inside the track(in the case of `TrackLocalStaticSample`) and by the user in case of `TrackLocalStaticRTP`. There are several problems with this:

1. If a track is bound to two different `RTCRtpSender`s the sequence numbers used by each RTP stream will be the same. This weakens SRTP and makes known-plaintext attacks easier. 
2. It breaks `RTCRtpSender::replace_track` which should allow replacing the current track with a different track with a matching envelope. Since the other track will have its own random starting sequence number this will not work in many instances(the sequence numbers between the two tacks should be consecutive).
3. When pausing one of several `RTCRtpSender` that a track is attached to the resulting gap in sequence numbers can cause SRTP validation to fail at the remote side(if the sender is paused for long enough). An attempt to address this is made in https://github.com/webrtc-rs/webrtc/pull/316

## Suggestion

> Move sequence number generation out of the track layer and into transport.

For `TrackLocalStaticRTP` this might mean that the underlying transport layer overwrites the sequence number of packets passed to `write_rtp` or perhaps we have a way to signal in `TrackLocalStaticRTP` whether sequence numbers should be respected or not

k0nserv:
One problem with this suggestion is that `TrackLocalStaticSample` has a way to propagate dropped packets via sequence number gaps which would break https://github.com/webrtc-rs/webrtc/blob/5612d59869e1a4b8d991ab129ab27103c2975f5b/webrtc/src/track/track_local/track_local_static_sample.rs#L55-L60

algesten:
Another part to this is when we implement RTX resend properly. Each RTX channel have separate SSRC and sequence number counter to the one it's repairing. It would seem logical that all sequence number generation is an internal concern and only done just before pushing the packet to the wire.

PR #378, opened 2022-12-27
Outbound sequence number overriding

logist322:
Move outbound sequence number overriding on the transport level to prevent `SRTP` crashing on the `RTCRtpSender::replace_track()` call.


melekes:
Is there an open issue? If not, @logist322 could you describe the problem in detail? Under what conditions does `SRTP` crash? How we can replicate?

k0nserv:
@melekes more details are available here https://github.com/webrtc-rs/webrtc/issues/317

@logist322 I think this is a good start, but we need to address this concern

> For TrackLocalStaticRTP this might mean that the underlying transport layer overwrites the sequence number of packets passed to write_rtp or perhaps we have a way to signal in TrackLocalStaticRTP whether sequence numbers should be respected or not

From #317.

In particular, the code as written here only works when the RTP packets are guaranteed to be sent in order, an RTP forwarder would break with this code since it might forward packets out of order.

logist322:
@k0nserv I guess it's ready for review :)

webrtc/src/rtp_transceiver/rtp_sender/mod.rs
```diff
@@ -479,6 +487,18 @@ impl RTCRtpSender {
         self.internal.read_rtcp(self.receive_mtu).await
     }
 
+    /// Enables overriding outcoming `RTP` packets' `sequence number`s.
```
melekes, line 490:
```suggestion
    /// Enables overriding outgoing `RTP` packets' `sequence number`s.
```

webrtc/src/rtp_transceiver/srtp_writer_future.rs
```diff
@@ -12,6 +12,70 @@ use interceptor::{Attributes, RTCPReader, RTPWriter};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Weak};
 use tokio::sync::Mutex;
+use util;
+
+/// RTP packet sequence number manager.
+pub(crate) struct SequenceTransformer(util::sync::Mutex<SequenceTransformerInner>);
+
+struct SequenceTransformerInner {
+    offset: u16,
```
melekes, line 21: would be good to add some comments here

webrtc/src/rtp_transceiver/srtp_writer_future.rs
```diff
@@ -12,6 +12,70 @@ use interceptor::{Attributes, RTCPReader, RTPWriter};
...
+            return None;
+        }
+
+        let offset = guard
+            .reset_needed
+            .then(|| {
+                guard.reset_needed = false;
+                let offset = guard.last_sq.overflowing_sub(raw_sn.overflowing_sub(1).0).0;
+                guard.offset = offset;
+                offset
```
melekes, line 70: we don't need to reset guard.last_sq here, right?

webrtc/src/rtp_transceiver/rtp_sender/mod.rs
logist322: Oh, you are right

webrtc/src/rtp_transceiver/srtp_writer_future.rs
logist322: I'll do it

webrtc/src/rtp_transceiver/srtp_writer_future.rs
logist322: Right, here we just getting or calculating `offset`.

webrtc/src/rtp_transceiver/srtp_writer_future.rs
logist322: @melekes I have added some docs

melekes (approved):
👍

logist322:
@melekes as I can understand, you have approved this PR. So, can you merge it? :sweat_smile:

logist322:
@melekes @k0nserv anyone, pls, merge it :sweat_smile:

k0nserv:
Sorry @logist322 I've been super busy recently. Can you rebase on main so clippy and formatting corrects itself?

logist322:
@k0nserv done!

webrtc/src/rtp_transceiver/srtp_writer_future.rs
```diff
@@ -12,6 +12,92 @@ use interceptor::{Attributes, RTCPReader, RTPWriter};
...
+        self.0.lock().reset_needed = true;
+    }
+
+    /// Gets [`Some`] consistent `sequence number` if this [`SequenceTransformer`] is
+    /// enabled or [`None`] if it is not.
+    ///
+    /// Once this method is called, considers data sending began.
+    fn seq_number(&self, raw_sn: u16) -> Option<u16> {
+        let mut guard = self.0.lock();
+        guard.data_sent = true;
```
k0nserv, line 80: Shouldn't this be after `guard.enabled` below?

logist322: No, it shouldn't.
We have already discussed it with you in discord. If data is sent we are not allowed to enable `SequenceTransformer`.

logist322: So, we should register data sending even if `SequenceTransformer` is not `enabled`.

k0nserv: Ah, right I recall

PR #381, opened 2023-01-02
sctp: remove unnecessary drops and use precise padding

melekes:
Refs https://github.com/webrtc-rs/webrtc/pull/364 and https://github.com/webrtc-rs/webrtc/pull/367

cc @KillingSpark

### DIFF
--- a/sctp/src/packet.rs
+++ b/sctp/src/packet.rs
@@ -165,7 +165,7 @@ impl Packet {
             let padding_needed = get_padding_size(writer.len());
             if padding_needed != 0 {
                 // padding needed if < 4 because we pad to 4
-                writer.extend_from_slice(&[0u8; 16][..padding_needed]);
+                writer.extend_from_slice(&[0u8; PADDING_MULTIPLE][..padding_needed]);
             }
         }
 
--- a/sctp/src/queue/pending_queue.rs
+++ b/sctp/src/queue/pending_queue.rs
@@ -1,17 +1,28 @@
+use tokio::sync::{Mutex, Semaphore};
+use util::sync::RwLock;
+
 use std::{
     collections::VecDeque,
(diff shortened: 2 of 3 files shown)


algesten (changes requested):
I don't believe this change will behave correctly.

Since all  the functions are `&self` rather than `&mut self`, we can potentially have multiple threads calling these functions at the same time.

The `let sem_lock` is a RAII guard that ensure the entire data structure is locked for the entire insertion. This is important in a case like this for loop.

https://github.com/webrtc-rs/webrtc/blob/daaf05d1021d8079ea1be4aa539e968726d18271/sctp/src/queue/pending_queue.rs#L120-L135

If two threads were competing in this loop, we would not be in "direct sequence". The code doc explains the relationship between the two locks.

>  The chunks of one fragmented message need to be put in direct sequence into the queue which the lock guarantees

Edit: provide correct link to loop

melekes:
> The let sem_lock is a RAII guard that ensure the entire data structure is locked for the entire insertion. This is important in a case like this for loop.

I agree, but my point is we don't need to drop them since it will be done by Rust automatically.

algesten:
> I agree, but my point is we don't need to drop them since it will be done by Rust automatically.

@melekes this is true. They do need to live for the entire scope though. So you can replace with `let _sem_lock = ` and just remove the `drop`.

KillingSpark:
If you are fixing the padding anyways, there are a few some more places where get_padding_size is used alongside a vec![] call.

I think generalizing what I did in the packet marshal code would be a good idea:

```rust
static PADDING_BYTES: [u8; PADDING_MULTIPLE] = [0;PADDING_MULTIPLE];
pub(crate) fn get_padding(len: usize) -> &'static [u8] {
    &PADDING_BYTES[..get_padding_size()]
}
```

k0nserv (approved):

PR #382, opened 2023-01-02
chores: min-version, deprecations etc

algesten:
* Fixes the min-version check in github actions.
* Run the same version aem, aem-gcm, sha1, sha2 and hmac across the board.
* Fix deprecations arising in chrono


algesten:
:'( this friggin PR…

melekes (approved):

PR #383, opened 2023-01-03
Improve the markdown of our README files

regexident:
See the commit messages for the individual improvements.


regexident:
> Increases project coverage by +1.25% 🎉

@codecov, you're drunk. I only touched markdown files.

dtls/examples/certificates/README.md
```diff
@@ -26,31 +26,31 @@ $ rm "${EXTFILE}" "${SERVER_NAME}.csr" "${CLIENT_NAME}.csr"
 ```
 
 in pion/examples/util/util.go, convert ECPrivateKey to PKCS8PrivateKey
-```
+
+```go
```
k0nserv, line 30: Hmm, we should remove this code and replace it with a TODO I think.

ice/CHANGELOG.md
```diff
@@ -6,25 +6,25 @@
 
 * Increased minimum support rust version to `1.60.0`.
 
-### Breaking changes  
+### Breaking changes
 
 * Make functions non-async [#338](https://github.com/webrtc-rs/webrtc/pull/338):
-    - `Agent`:
```
k0nserv, line 12: This change from `-` to `*` doesn't seem consequential

regexident: The CommonMark spec [says the following](https://spec.commonmark.org/0.30/#lists) about lists (emphasis mine):

> A [list](https://spec.commonmark.org/0.30/#list) is a sequence of one or more list items [**of the same type**](https://spec.commonmark.org/0.30/#of-the-same-type). The list items may be separated by any number of blank lines.
> 
> Two list items are [of the same type](https://spec.commonmark.org/0.30/#of-the-same-type) if they begin with a [list marker](https://spec.commonmark.org/0.30/#list-marker) **of the same type**. Two list markers are of the same type if (a) they are **bullet list markers using the same character** (`-`, `+`, or `*`) or (b) they are ordered list numbers with the same delimiter (either `.` or `)`).

As such a list with mixed markers (i.e. `*` and `-`) is invalid.

dtls/examples/certificates/README.md
regexident: Where would that TODO go and what would it say?

ice/CHANGELOG.md
k0nserv: Hmm, doesn't this constitute a new list since it's a sublist though? So this would be fine:

```markdown
* Hello
  - Start of sublist
  - Other item
* Bye
```

**Rendered:**

* Hello
  - Start of sublist
  - Other item
* Bye

dtls/examples/certificates/README.md
k0nserv: Right here and it should probably say `TODO: Port code example from Pion <link>`

ice/CHANGELOG.md
regexident: Hm, upon further inspection the example [Principle of uniformity](https://spec.commonmark.org/0.30/#principle-of-uniformity) does indeed show …

```plain
* I need to buy
    - new shoes
    - a coat
    - a plane ticket
```

… as a supposedly correct use. 🤔 

The popular [markdownlint](https://github.com/DavidAnson/markdownlint) linter (and its corresponding [VSCode extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint)) does produce a warning for it though.

dtls/examples/certificates/README.md
regexident: That piece of code is actually nowhere to be found in [pion/dtls](https://github.com/pion/dtls) at this point.

The closest I could find is [this implementation](https://github.com/pion/dtls/blob/0fbf8d103737d247a36377c739eb34bea6d790a6/examples/util/util.go#L95-L124) of a `func LoadKey(path string) (crypto.PrivateKey, error)`, which was later removed in [this commit](https://github.com/pion/dtls/commit/5c4fb0e2211ae28500bd0a3bf52f2ce5abe66a74) from Apr 26, 2022.

k0nserv: Thanks for investigating. Perhaps we just write something to the effect of `TODO: write some example code to convert an EC private key to a PKCS#8 key` 🤷🏼

regexident: Looks like there exists a crate for this!

[topk8](https://crates.io/crates/topk8) – Convert private keys to PKCS#8 format in pure Rust

Added a rust example based on it!

k0nserv: Aight

ice/CHANGELOG.md
regexident: @k0nserv I reversed the `-` -> `*` change, keeping only the indentation fix (to make things uniform with other markdown files in the project).

k0nserv (approved):

PR #385, opened 2023-01-04
Unify `assert_eq!(…)`/`assert_ne!(…)` to take args in order of `…!(actual, expected)`

regexident:
While there isn't an explicit convention for this in Rust the stdlib docs all use `assert_eq!(actual, expected)`/`assert_ne!(actual, expected)` —as do the [Rust by Example](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html) and [The Rust Programming Language](https://doc.rust-lang.org/book/ch11-01-writing-tests.html) <sup>([*](https://github.com/rust-lang/book/pull/3497))</sup> books— making it a sort of implicit convention.

The project contains over 2000 occurrences of the pattern `/assert_(ne|eq|ok|err)!/`, so this PR is clearly a best-effort attempt to unify the arg order. There are no doubt instances that I missed either due to inattentiveness after having gone through thousands of lines of code, or ambiguous argument names.

### DIFF
--- a/data/src/data_channel/data_channel_test.rs
+++ b/data/src/data_channel/data_channel_test.rs
@@ -450,12 +450,12 @@ async fn test_data_channel_buffered_amount() -> Result<()> {
     }
 
     let n = dc0.write(&Bytes::new()).await?;
-    assert_eq!(0, n, "data length should match");
-    assert_eq!(1, dc0.buffered_amount(), "incorrect bufferedAmount");
+    assert_eq!(n, 0, "data length should match");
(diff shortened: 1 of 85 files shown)


algesten (approved):
I personally don't care much about the argument order of asserts, but I'm happy to have cleaner code. Let's merge it! :)

regexident:
Definitely not a must-have, but a consistent arg ordering makes debugging test failures easier in my experience by not having first figure out which is which.

Issue #386, opened 2023-01-04
Initial packets lost during unknown SSRC probing


k0nserv:
We seem to lose the first packet when doing unknown SSRC detection. For example, when identifying simulcast tracks via rid + mid extension the first packet is never delivered to the track.

## Example

This image showcases the issue. On the left is a packet capture from Chrome's perspective and on the right is the result of querying a packet capture taken from an interceptor. Note that the initial packet, sequence number `24698`, is missing in the capture while the second packet, sequence number `24699`, does appear in the capture.

![image](https://user-images.githubusercontent.com/1333960/210596755-d876fc2d-f97a-49ce-a164-714b3880d902.png)

## Notes

The read that consumes the initial packet happens here https://github.com/webrtc-rs/webrtc/blob/e60607fa269f20fc002347f32e0d5b756f5b92e3/webrtc/src/peer_connection/peer_connection_internal.rs#L964 I believe

k0nserv:
A PR that attempts to fix this https://github.com/webrtc-rs/webrtc/pull/387

k0nserv:
The above screen shot if flawed since the RTP captures we use are based on interceptors. Currently it will always miss the first packet, even with #387 I verified via `read_rtp` and reading the code that we do still miss at least one packet

k0nserv:
Done in #387

PR #387, opened 2023-01-04
Prevent loss of initial packet(s) in RTP streams

k0nserv:
The method by which we detected non-signaled SSRCs would silently drop a
few packets. This is particularly problematic for a video stream since
these initial packets usually contain the first part of a keyframe to
start the stream.

See: #386 

**Note:** The first packet read for each RTP stream is not processed by interceptors unfortunately.

### DIFF
--- a/webrtc/src/dtls_transport/mod.rs
+++ b/webrtc/src/dtls_transport/mod.rs
@@ -582,10 +582,10 @@ impl RTCDtlsTransport {
         stream_info: &StreamInfo,
         interceptor: &Arc<dyn Interceptor + Send + Sync>,
     ) -> Result<(
-        Option<Arc<srtp::stream::Stream>>,
-        Option<Arc<dyn RTPReader + Send + Sync>>,
-        Option<Arc<srtp::stream::Stream>>,
-        Option<Arc<dyn RTCPReader + Send + Sync>>,
+        Arc<srtp::stream::Stream>,
+        Arc<dyn RTPReader + Send + Sync>,
+        Arc<srtp::stream::Stream>,
+        Arc<dyn RTCPReader + Send + Sync>,
     )> {
         let srtp_session = self
             .get_srtp_session()
@@ -608,10 +608,10 @@ impl RTCDtlsTransport {
         let rtcp_interceptor = interceptor.bind_rtcp_reader(rtcp_stream_reader).await;
 
         Ok((
-            Some(rtp_read_stream),
-            Some(rtp_interceptor),
-            Some(rtcp_read_stream),
-            Some(rtcp_interceptor),
+            rtp_read_stream,
+            rtp_interceptor,
+            rtcp_read_stream,
+            rtcp_interceptor,
         ))
     }
 }
--- a/webrtc/src/peer_connection/peer_connection_internal.rs
+++ b/webrtc/src/peer_connection/peer_connection_internal.rs
@@ -1,3 +1,4 @@
+use bytes::Bytes;
 use tokio::time::Instant;
 
 use super::*;
@@ -10,6 +11,7 @@ use crate::stats::{
 use crate::track::TrackStream;
 use crate::{SDES_REPAIR_RTP_STREAM_ID_URI, SDP_ATTRIBUTE_RID};
 use arc_swap::ArcSwapOption;
+use std::collections::VecDeque;
 use std::sync::atomic::AtomicIsize;
 use std::sync::Weak;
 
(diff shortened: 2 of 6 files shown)


webrtc/src/dtls_transport/mod.rs
```diff
@@ -608,10 +608,10 @@ impl RTCDtlsTransport {
         let rtcp_interceptor = interceptor.bind_rtcp_reader(rtcp_stream_reader).await;
 
         Ok((
-            Some(rtp_read_stream),
```
k0nserv, line 611: There was no reason for this to return `Option`s since we always used `Some`

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -924,154 +933,156 @@ impl PeerConnectionInternal {
...
-            if let Some(parsed) = &rd.parsed {
-                // If the remote SDP was only one media section the ssrc doesn't have to be explicitly declared
-                let handled = self.handle_undeclared_ssrc(ssrc, parsed).await?;
-                if handled {
-                    return Ok(());
-                }
+        let parsed = match self.remote_description().await.and_then(|rd| rd.parsed) {
+            Some(r) => r,
+            None => return Err(Error::ErrPeerConnRemoteDescriptionNil),
+        };
```
k0nserv, line 939: Pion includes new code to bail out of signalled SSRCs here, we should port this

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -924,154 +933,156 @@ impl PeerConnectionInternal {
...
+                uri: SDES_REPAIR_RTP_STREAM_ID_URI.to_owned(),
+            })
+            .await;
+
+        let mut buf = vec![0u8; self.setting_engine.get_receive_mtu()];
+        // Packets that we read as part of simulcast probing that we need to make available
+        // if we do find a track later.
+        let mut buffered_packets: VecDeque<(Bytes, Attributes)> = VecDeque::default();
+
+        let n = rtp_stream.read(&mut buf).await?;
```
k0nserv, line 980: This read does not go through the interceptor chain hence is missed by all interceptors :( Unsure if we can solve this

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -924,154 +933,156 @@ impl PeerConnectionInternal {
...
+
+        let (mut mid, mut rid, mut rsid, payload_type) = handle_unknown_rtp_packet(
+            &buf[..n],
+            mid_extension_id as u8,
+            sid_extension_id as u8,
+            rsid_extension_id as u8,
+        )?;
+        // TODO: Can we have attributes on the first packets?
```
k0nserv, line 988: This is because we don't read through the interceptor chain

webrtc/src/track/track_remote/mod.rs
```diff
@@ -207,40 +207,37 @@ impl TrackRemote {
         handlers.on_unmute = Some(Box::new(handler));
     }
 
-    /// Read reads data from the track.
+    /// Reads data from the track.
+    ///
+    /// **Cancel Safety:** This method is not cancel safe. Dropping the resulting [`Future`] before
```
k0nserv, line 212: This isn't new. It wasn't cancel safe before either, I've just documented it

webrtc/src/peer_connection/peer_connection_internal.rs
k0nserv: Idle though: Maybe we can open the SRTP stream earlier in which case we can optimistically configure the interceptor chain...

webrtc/src/peer_connection/peer_connection_internal.rs
```diff
@@ -924,154 +933,156 @@ impl PeerConnectionInternal {
...
+                    );
                 }
-
-                return Err(Error::ErrPeerConnSimulcastIncomingSSRCFailed);
+                return Ok(());
             }
         }
 
-        Err(Error::ErrPeerConnRemoteDescriptionNil)
+        let _ = rtp_read_stream.close().await;
```
melekes, line 1080: ❤️

webrtc/src/peer_connection/peer_connection_internal.rs
melekes: what are the downsides of this read not going through interceptors?

k0nserv: We have an interceptor that dumps all ingress RTP, it misses this initial packet. In general any interceptors that reads RTP will miss the first RTP packet

k0nserv: It's a bit of a chicken and egg problem however, we only have the necessary information to create the interceptor at line 1001. This information is derived from the payload type which we read out of the first packet.

We'd need a way to fake a read through the interceptor chain to alleviate this I think

PR #389, opened 2023-01-08
Prepare an initial release for the new 'webrtc-constraints' crate

regexident:
- [x] Add `README.md` file
- [x] Add `CHANGELOG.md` file
- [x] Add example files
- [x] Do some final clean up

---

Other than adding a `constraints-v0.1.0`git tag for the release commit, are there any other release steps I should be aware of when doing an initial release of `webrtc-constraints`?

### DIFF
--- a/constraints/Cargo.toml
+++ b/constraints/Cargo.toml
@@ -26,3 +26,8 @@ serde_json = { version = "1.0.91", features = ["preserve_order"] }
 [features]
 default = ["serde"]
 serde = ["dep:serde", "indexmap/serde"]
+
+[[example]]
+name = "json"
+path = "examples/json.rs"
+required-features = ["serde"]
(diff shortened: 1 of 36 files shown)


algesten (approved):

constraints/Cargo.toml
```diff
@@ -24,5 +24,10 @@ lazy_static = "1.4.0"
 serde_json = { version = "1.0.91", features = ["preserve_order"] }
 
 [features]
-default = []
+default = ["serde"]
 serde = ["dep:serde", "indexmap/serde"]
+
+[[example]]
```
melekes, line 30: what about `macros` and `native` examples? should we add them here too?

melekes (approved):
👍

constraints/Cargo.toml
regexident: The reason for only the `json` example being listed is that it requires specific features (`serde`).

k0nserv (approved):

PR #392, opened 2023-01-14
Fix `SettingEngine.set_interface_filter` not working on Windows

smilecc:
I'm using `SettingEngine.set_interface_filter` and found it doesn't work on Windows.

I read `util/ifaces/ffi/windows` source code and found that the value of `Interface.name` is `"".to_string()` when initialized.

I found that the object returned after calling `GetAdaptersAddresses` contains two members `FriendlyName` and `AdapterName`, so I read the source code of Pion. It calls the standard library `net.Interfaces()` of go, and in version 1.19.4 of go, it uses `FriendlyName`.

```go
// https://cs.opensource.google/go/go/+/master:src/net/interface_windows.go;l=61
ifi := Interface{
    Index: int(index),
    Name:  windows.UTF16PtrToString(aa.FriendlyName),
}
```

So to be consistent with Pion, I also used `FriendlyName` to fix this.

In Windows, `wchar_t` and `WCHAR` are both 16bit, so I used `wcslen` and `OsString` to deal with this problem.

Currently it works fine on Windows:

```
[2023-01-14T22:39:30Z INFO  webrtc_actix::engine::rtc] interface vEthernet (以太网 3)
[2023-01-14T22:39:30Z INFO  webrtc_actix::engine::rtc] interface vEthernet (以太网)
[2023-01-14T22:39:30Z INFO  webrtc_actix::engine::rtc] interface 以太网 3
[2023-01-14T22:39:30Z INFO  webrtc_actix::engine::rtc] interface vEthernet (WSL)
[2023-01-14T22:39:30Z INFO  webrtc_actix::engine::rtc] interface 本地连接
[2023-01-14T22:39:30Z INFO  webrtc_actix::engine::rtc] interface 以太网
[2023-01-14T22:39:30Z INFO  webrtc_actix::engine::rtc] interface Loopback Pseudo-Interface 1
[2023-01-14T22:39:30Z INFO  webrtc_actix::engine::rtc] interface vEthernet (Default Switch)
```

I think this is a small problem, so I didn't create Issues. Thanks a lot.

### DIFF
--- a/util/src/ifaces/ffi/windows/mod.rs
+++ b/util/src/ifaces/ffi/windows/mod.rs
@@ -1,5 +1,6 @@
 #![allow(unused, non_upper_case_globals)]
 
+use winapi::ctypes::wchar_t;
 use winapi::shared::basetsd::{UINT32, UINT8, ULONG64};
 use winapi::shared::guiddef::GUID;
 use winapi::shared::minwindef::{BYTE, DWORD, PULONG, ULONG};
@@ -14,9 +15,11 @@ const MAX_DNS_SUFFIX_STRING_LENGTH: usize = 256;
 pub const IP_ADAPTER_IPV4_ENABLED: DWORD = 0x0080;
 pub const IP_ADAPTER_IPV6_ENABLED: DWORD = 0x0100;
 
+use std::ffi::OsString;
 use std::io;
 use std::mem;
 use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
+use std::os::windows::prelude::OsStringExt;
 use std::ptr;
 
 use winapi::shared::winerror::{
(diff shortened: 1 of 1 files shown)


smilecc:
@k0nserv I changed the code to be more secure and added `// SAFETY: ...` comment. please review again. thx :)

util/src/ifaces/ffi/windows/mod.rs
```diff
@@ -316,6 +320,44 @@ unsafe fn local_ifaces_with_buffer(buffer: &mut Vec<u8>) -> io::Result<()> {
...
+        /// determine the length of a fixed-size wide-character string.
+        ///
+        /// In windows, wchar_t is 16bit, in GNU LibC it is 32bit,
+        /// So in Windows, the length of both wchar_t and WCHAR is equal.
+        /// See https://learn.microsoft.com/en-us/cpp/cpp/char-wchar-t-char16-t-char32-t
+        fn wcsnlen(buf: *const wchar_t, size: size_t) -> usize;
+    }
+
+    // # SAFETY
+    // Caller has provided a pointer of valid C String, And confirm the maximum length of the buffer.
```
k0nserv, line 341: Given this is placing requirements on the caller shouldn't the whole function be `unsfafe` it seems possible to misuse this function from safe code but trigger undefined behaviour. WDYT?

smilecc:
@k0nserv 
Sorry, I didn't think well before, this function should be unsafe, if the pointer is not a valid c str pointer, its behaviour is undefined, you are right.

Before this, I didn't fully consider that others called this function, maybe this function should be an internal function of `map_adapter_addresses` when my first commit, because my original intention of writing this function is to make the code for handling ipv4 and ipv6 `friendly_name` not duplicate.

Anyway, maybe this function will really help other ffi of Windows , so I add `unsafe` keyword and more SAFETY 
comment instructions to this function.

Hope I didn't misunderstand your suggestion, thanks again!

k0nserv:
That sounds good. Even if the function is only internal we should make it `unsafe`, people using it can still make mistakes and need to understand the requirements.

Please add a 


```
# Safety
```

section to the documentation for the function itself

PR #395, opened 2023-01-17
Custom `RTCRtpTransceiver` creating

logist322:
This PR provides API for creating custom `RTCRtpSender`, `RTCRtpReceiver` and `RTCRtpTransceiver` and adding `RTCRtpTransceiver` to `RTCPeerConnection`.

### DIFF
--- a/webrtc/src/api/mod.rs
+++ b/webrtc/src/api/mod.rs
@@ -168,6 +168,14 @@ impl API {
         )
         .await
     }
+
+    pub fn setting_engine(&self) -> Arc<SettingEngine> {
+        Arc::clone(&self.setting_engine)
+    }
+
+    pub fn media_engine(&self) -> Arc<MediaEngine> {
+        Arc::clone(&self.media_engine)
+    }
 }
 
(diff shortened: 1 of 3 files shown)


k0nserv:
Can you talk a bit about why this is necessary? The spec compliant API should be enough I think

logist322:
In general, it provides more advanced ways for working with `RTCRtpTransceiver`s. For example, in our project we need to pass the unique `Interceptor` for every `RTCRtpSender`.

logist322:
@k0nserv it is ready for review

logist322:
@k0nserv ping

k0nserv:
> For example, in our project we need to pass the unique Interceptor for every RTCRtpSender.

Out of interest why is this needed? This isn't something that Pion provides atm so I'm thinking maybe there's another solve for your use case

k0nserv:
@logist322 Can you talk a bit more about why this requires opening up this API? You can still register custom interceptors via the interceptor registry

rainliu (approved):

PR #396, opened 2023-01-18
Fix a MAX_RETRY_ATTEMPTS bug for relay_conn

623637646:
Should stop attempting if successful.


fzdy1914:
Add related golang code to support this. In case of nil error, the check will pass and break the loop.
image

PR #407, opened 2023-02-01
general: apply clippy lints

jxs:
updated the crates to the 2021 edition due to the new formatting lints in `panic!` and `assert!` macros.  
If preferred I can try to isolate the fixes to only assert #405 passes.

### DIFF
--- a/constraints/Cargo.toml
+++ b/constraints/Cargo.toml
@@ -2,7 +2,7 @@
 name = "webrtc-constraints"
 version = "0.1.0"
 authors = ["Vincent Esche <regexident@gmail.com>"]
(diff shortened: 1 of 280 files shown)


algesten:
Thanks!

@k0nserv is 2021 ok or are we more conservative on the MSRV?

k0nserv:
Yeah that's fine, current minimum target is 1.60.0 so should be fine

algesten:
@jxs i think it wants a `cargo fmt`

jxs:
thanks @algesten, updated!

algesten (approved):

PR #411, opened 2023-02-27
Remove `derive_builder` dependency

thomaseizinger:
This is quite a heavy dependency and it is only in use in one place which IMO doesn't need it. A struct with all public fields an easily be built in a single expression using `Config { negotiated: true, ..Default::default() }`.

Here is the output of `cargo tree` for us. In this case, it is also the only place where we pull in `darling` which by itself is quite heavy:

image

### DIFF
--- a/data/CHANGELOG.md
+++ b/data/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Remove builder pattern from `data_channel::Config`.
+
 ## v0.6.0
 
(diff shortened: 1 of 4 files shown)


data/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
```
yngrtc, line 4: add "##v0.7.0" since it is just released.

yngrtc (approved):

Issue #422, opened 2023-03-21
Types `RTCRtpReceiver` and `RTCRtpSender` is not `Send`


kolserdav:
At attempt to close peer_connection gives such errors:
```sh
kol @ ~/Projects/werift-sfu-react $ RUST_LOG=info cargo run
   Compiling uyem-server v0.1.0 (/home/kol/Projects/werift-sfu-react/packages/server-rs)
error: future cannot be sent between threads safely
   --> packages/server-rs/src/ws/mod.rs:115:26
    |
115 |             tokio::spawn(this.handle_mess(stream));
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^ future returned by `handle_mess` is not `Send`
    |
    = help: within `impl futures::Future<Output = ()>`, the trait `std::marker::Send` is not implemented for `std::sync::MutexGuard<'_, Arc<RTCRtpSender>>`
note: future is not `Send` as this value is used across an await
   --> /home/kol/.cargo/registry/src/github.com-1ecc6299db9ec823/webrtc-0.7.1/src/rtp_transceiver/mod.rs:428:26
    |
427 |             let sender = self.sender.lock();
    |                 ------ has type `webrtc::webrtc_util::sync::MutexGuard<'_, Arc<RTCRtpSender>>` which is not `Send`
428 |             sender.stop().await?;
    |                          ^^^^^^ await occurs here, with `sender` maybe used later
429 |         }
    |         - `sender` is later dropped here
note: required by a bound in `tokio::spawn`
   --> /home/kol/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.26.0/src/task/spawn.rs:163:21
    |
163 |         T: Future + Send + 'static,
    |                     ^^^^ required by this bound in `tokio::spawn`

error: future cannot be sent between threads safely
   --> packages/server-rs/src/ws/mod.rs:115:26
    |
115 |             tokio::spawn(this.handle_mess(stream));
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^ future returned by `handle_mess` is not `Send`
    |
    = help: within `impl futures::Future<Output = ()>`, the trait `std::marker::Send` is not implemented for `std::sync::MutexGuard<'_, Arc<RTCRtpReceiver>>`
note: future is not `Send` as this value is used across an await
[...]

PR #423, opened 2023-03-22
fixed send mutex guard between threads bug

kolserdav:
This pull request solves this issue https://github.com/webrtc-rs/webrtc/issues/422. Implementation example of `Send` for `MutexGuard` I took from here https://github.com/rust-lang/rust/blob/98815742cf2e914ee0d7142a02322cf939c47834/library/std/src/sync/mutex.rs#L204 .

### DIFF
--- a/util/src/sync/mod.rs
+++ b/util/src/sync/mod.rs
@@ -27,6 +27,8 @@ impl<T> Mutex<T> {
 /// dropped (falls out of scope), the lock will be unlocked.
 pub struct MutexGuard<'a, T>(sync::MutexGuard<'a, T>);
 
+unsafe impl<'a, T> Send for MutexGuard<'a, T> {}
+
 impl<'a, T> ops::Deref for MutexGuard<'a, T> {
     type Target = T;
 


yngrtc (approved):

alexlapa:
This should not be merged. `MutexGuard` is `!Send` by design. The example from `std::sync::Mutex` clearly says that:

```rust
impl<T: ?Sized> !Send for MutexGuard<'_, T> {}

unsafe impl<T: ?Sized + Sync> Sync for MutexGuard<'_, T> {}
```

kolserdav:
> This should not be merged. `MutexGuard` is `!Send` by design. The example from `std::sync::Mutex` clearly says that:
> 
> ```rust
> impl<T: ?Sized> !Send for MutexGuard<'_, T> {}
> 
> unsafe impl<T: ?Sized + Sync> Sync for MutexGuard<'_, T> {}
> ```

The example here is not for verbatim copying, but for an example that the implementation of Send is the place to be for MutexGuard.  The implementation that I proposed only works in the existing variation.  The rest of the type parameter restrictions from the "example" are inadequate in this implementation.

PR #429, opened 2023-03-27
Stop pc bug

kolserdav:
This PR solved this issue https://github.com/webrtc-rs/webrtc/issues/422 .
This fix repeats behavior in other places like: https://github.com/webrtc-rs/webrtc/blob/5859550cac046376bd4c7a252e79199c54f44bb8/webrtc/src/rtp_transceiver/mod.rs#L399  and https://github.com/webrtc-rs/webrtc/blob/5859550cac046376bd4c7a252e79199c54f44bb8/webrtc/src/rtp_transceiver/mod.rs#L446
In this case, the call goes not to `MutexGuard` but to `Arc`.

### DIFF
--- a/webrtc/src/rtp_transceiver/mod.rs
+++ b/webrtc/src/rtp_transceiver/mod.rs
@@ -399,7 +399,7 @@ impl RTCRtpTransceiver {
         }
 
         {
-            let receiver = self.receiver.lock().await.clone();
+            let receiver = self.receiver.lock().await;
             let pause_receiver = !current_direction.has_recv();
 
             if pause_receiver {
@@ -427,11 +427,11 @@ impl RTCRtpTransceiver {
         self.stopped.store(true, Ordering::SeqCst);
 
         {
-            let sender = self.sender.lock().await.clone();
+            let sender = self.sender.lock().await;
             sender.stop().await?;
         }
         {
-            let r = self.receiver.lock().await.clone();
+            let r = self.receiver.lock().await;
             r.stop().await?;
         }
 
@@ -446,7 +446,7 @@ impl RTCRtpTransceiver {
     ) -> Result<()> {
         let track_is_none = track.is_none();
         {
-            let sender = self.sender.lock().await.clone();
+            let sender = self.sender.lock().await;
             sender.replace_track(track).await?;
         }
 
--- a/webrtc/src/rtp_transceiver/mod.rs
+++ b/webrtc/src/rtp_transceiver/mod.rs
@@ -175,7 +175,7 @@ pub type TriggerNegotiationNeededFnOption =
 
 /// RTPTransceiver represents a combination of an RTPSender and an RTPReceiver that share a common mid.
 pub struct RTCRtpTransceiver {
-    mid: OnceCell<SmallStr>,                  //atomic.Value
+    mid: OnceCell<SmallStr>,              //atomic.Value
     sender: Mutex<Arc<RTCRtpSender>>,     //atomic.Value
     receiver: Mutex<Arc<RTCRtpReceiver>>, //atomic.Value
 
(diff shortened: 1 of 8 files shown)


kolserdav (commented):
Unintentional change from failed PR https://github.com/webrtc-rs/webrtc/pull/423 , this is fixed in the third commit.

webrtc/src/rtp_transceiver/mod.rs
```diff
@@ -424,11 +424,11 @@ impl RTCRtpTransceiver {
         self.stopped.store(true, Ordering::SeqCst);
 
         {
-            let sender = self.sender.lock();
+            let sender = self.sender.lock().clone();
```
yngrtc, line 427: I think the usage of SyncMutex for self.sender and self.receiver is wrong here.

SyncMutex must not be used in await.

Could change self.sender and self.receiver from SyncMutex to tokio::sync::Mutex?

yngrtc: something like
```
   sender: Arc<tokio::sync::Mutex<RTCRtpSender>>,     //atomic.Value
    receiver: Arc<tokio::sync::Mutex<RTCRtpReceiver>>, //atomic.Value
```

kolserdav: I tried this and it works even without `clone()` https://github.com/webrtc-rs/webrtc/pull/429/commits/ef0f6c5f332e094e7ac4a2270e23086cce2f13cf

kolserdav: > @yngrtc ,
> 
> > Could you try to change self.sender and self.receiver from SyncMutex to tokio::sync::Mutex?
> 
> This would revert #402 and introduce unnecessary async in public API.
> 
> > SyncMutex must not be used in across await
> 
> Mutex guard won't be held across await points if the `Arc` behind mutex is cloned. Its:
> 
> ```rust
> let sender: MutexGuard<_> = self.sender.lock().unwrap();
> let sender: Arc<_> = sender.clone();
> ```
> 
> Perhaps its better to explicitly use `Arc::clone(&*self.sender.lock().unwrap())` so it would be more clear that `stop()` is called on an `Arc` and not on a `MutexGuard`

Change it https://github.com/webrtc-rs/webrtc/pull/429/commits/b15f1cb4aa3408755c352137dd0740a9ba0d9831

chengyuhui:
Is there anything blocking this PR? I'm currently working around this using `spawn_local`, but that's not very ideal.

Perondas:
@chengyuhui can you share your workaround?

webrtc/src/rtp_transceiver/mod.rs
```diff
@@ -398,7 +399,7 @@ impl RTCRtpTransceiver {
         }
 
         {
-            let receiver = self.receiver.lock().clone();
+            let receiver = self.receiver.lock().await.clone();
```
yngrtc, line 402: does it still need .clone()?

webrtc/src/rtp_transceiver/mod.rs
```diff
@@ -426,11 +427,11 @@ impl RTCRtpTransceiver {
         self.stopped.store(true, Ordering::SeqCst);
 
         {
-            let sender = self.sender.lock();
+            let sender = self.sender.lock().await.clone();
```
yngrtc, line 430: same here.

yngrtc, line 430: same here.

webrtc/src/rtp_transceiver/mod.rs
```diff
@@ -426,11 +427,11 @@ impl RTCRtpTransceiver {
...
             sender.stop().await?;
         }
         {
-            let r = self.receiver.lock();
+            let r = self.receiver.lock().await.clone();
```
yngrtc, line 434: same here

yngrtc (approved):

PR #432, opened 2023-04-01
Update x25519-dalek

gmorenz:
Without this webrtc fails to compile because x25519 2.0.0-rc.2 introduced a breaking change by moving `StaticSecret` behind a feature flag, and cargo selects the latest version of the dependency.

```
error[E0433]: failed to resolve: could not find `StaticSecret` in `x25519_dalek`
  --> dtls/src/curve/named_curve.rs:57:44
   |
57 |             let secret_key = x25519_dalek::StaticSecret::new(OsRng);
   |                                            ^^^^^^^^^^^^
   |                                            |
   |                                            could not find `StaticSecret` in `x25519_dalek`
   |                                            help: a struct with a similar name exists: `SharedSecret`
```

This updates the dependency to point to the latest version (so the feature flag will always exist), and enables it. It also fixes a deprecation warning where `StaticSecret::new` was renamed to `StaticSecret::random_from_rng`

### DIFF
--- a/dtls/Cargo.toml
+++ b/dtls/Cargo.toml
@@ -35,7 +35,7 @@ aes-gcm = "0.10.1"
 ccm = "0.3.0"
 tokio = { version = "1.19", features = ["full"] }
 async-trait = "0.1.56"
-x25519-dalek = "2.0.0-pre.1"
+x25519-dalek = { version = "2.0.0-rc.2", features = ["static_secrets"] }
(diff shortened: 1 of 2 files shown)


yngrtc (approved):

gmorenz:
At first glance I don't believe either CI failure is related to this PR.

xxsanekxx (approved):

PR #437, opened 2023-04-11
test for RTCPeerConnection::close

Horusiath:
This is really an issue report. It's basically not allowed to use `RTCPeerConnection` inside of tokio tasks (via `spawn`), because the `close` method does not implement `Send` trait. Hence once opened, peer connection cannot be closed in async/await method context. The test below produces following compiler error:

```
error: future cannot be sent between threads safely
    --> webrtc\src\peer_connection\mod.rs:2101:36
     |
2101 |         let handle  = tokio::spawn(async move { peer().await });
     |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
     |
     = help: within `[async block@webrtc\src\peer_connection\mod.rs:2101:36: 2101:63]`, the trait `Send` is not implemented for `std::sync::MutexGuard<'_, std::sync::Arc<rtp_sender::RTCRtpSender>>`
note: future is not `Send` as this value is used across an await
    --> webrtc\src\rtp_transceiver\mod.rs:428:26
     |
427  |             let sender = self.sender.lock();
     |                 ------ has type `webrtc_util::sync::MutexGuard<'_, std::sync::Arc<rtp_sender::RTCRtpSender>>` which is not `Send`
428  |             sender.stop().await?;
     |                          ^^^^^^ await occurs here, with `sender` maybe used later
429  |         }
     |         - `sender` is later dropped here
note: required by a bound in `tokio::spawn`
     |
163  |         T: Future + Send + 'static,
     |                     ^^^^ required by this bound in `tokio::spawn`
```

### DIFF
--- a/webrtc/src/peer_connection/mod.rs
+++ b/webrtc/src/peer_connection/mod.rs
@@ -2084,3 +2084,54 @@ impl RTCPeerConnection {
         self.internal.add_rtp_transceiver(t).await
     }
 }
+
+#[cfg(test)]
+mod test {
+    use std::sync::Arc;
+    use interceptor::registry::Registry;
+    use crate::api::APIBuilder;
+    use crate::api::interceptor_registry::register_default_interceptors;
+    use crate::api::media_engine::MediaEngine;
+    use crate::Error;
+    use crate::ice_transport::ice_server::RTCIceServer;
+    use crate::peer_connection::configuration::RTCConfiguration;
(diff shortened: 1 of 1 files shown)


yngrtc:
Thanks for the test. I merged the test in https://github.com/webrtc-rs/webrtc/pull/429.

PR #440, opened 2023-04-25
Surfaces io::Error from H264Reader read.

nathansizemore:
This is useful for custom streams, that aren't blocking and implement a non-blocking IO model, where `io::ErrorKind::WouldBlock` is returned, along with other custom `io::Error` return values that are valid, but right now not exposed once the reader is wrapped within the `H264Reader`.

### DIFF
--- a/media/src/io/h264_reader/mod.rs
+++ b/media/src/io/h264_reader/mod.rs
@@ -154,7 +154,7 @@ impl<R: Read> H264Reader<R> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


nathansizemore:
I have a project that this would be nice to use in. When this gets merged, can this issue a new release, please and thank you!

EricHenry:
My project would benefit from this too!

yngrtc (approved):

PR #442, opened 2023-05-08
Switched mid and rid to SmolStr

### LINKED ISSUE #314: Consider smol_str for mid/rid
The specification recommends that `mid` and `rid` only use about 3 bytes of ASCII hence we'd benefit from storing these as [`smol_str::Smolstr`](https://docs.rs/smol_str/0.1.23/smol_str/struct.SmolStr.html) which can allocate up to 22 bytes inline, chance are we'd always avoid heap allocation for these values with this approach

gautamprikshit1:
Fixes #314 maybe

### DIFF
--- a/webrtc/Cargo.toml
+++ b/webrtc/Cargo.toml
@@ -40,6 +40,7 @@ bytes = "1"
 thiserror = "1.0"
 waitgroup = "0.1.2"
 regex = "1.7.1"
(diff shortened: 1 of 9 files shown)


yngrtc:
please re-base it to fix cargo test

rainliu (approved):

PR #444, opened 2023-05-11
Allow setting local SDP for `RTCSignalingState::HaveLocalOffer`

logist322:
This PR makes setting local `SDP` spec compliant (as described in **RFC8829** [section 5.5](https://datatracker.ietf.org/doc/html/rfc8829#section-5.5)).


logist322:
@rainliu readu for review. CI fails, but it is not related to this PR.

yngrtc (approved):

PR #445, opened 2023-05-12
Adds configurable capacity to H264Reader.

nathansizemore:
Chunk reading size should be left to consumer to configure based on source.


nathansizemore:
A publish on this change would be greatly appreciated :)

yngrtc (approved):

PR #453, opened 2023-06-04
Use native serde impl for SmolStr

lolPants:
https://lib.rs/crates/smol_str has implementations of serde `Serialize` and `Deserialize` behind a feature flag.


yngrtc:
please fix clippy

yngrtc (approved):

PR #459, opened 2023-06-18
Update cargo.yml

yngrtc:


.github/workflows/cargo.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: ["ubuntu-latest", "macos-latest"]
         toolchain:
-          - 1.65.0 # min supported version (https://github.com/webrtc-rs/webrtc/#toolchain)
+          # - 1.65.0 # min supported version (https://github.com/webrtc-rs/webrtc/#toolchain)
```
melekes, line 23: curious what's the problem with min supported version?

rainliu (approved):

PR #461, opened 2023-06-19
yank interceptor 0.8.1 & 0.8.2 and rtcp 0.7.1 & 0.7.2 versions and re…

### LINKED ISSUE #412: Bad semver bump in latest release
I haven't found the exact problem yet but one of your users posted that they have started hitting compilation errors originating in webrtc: https://github.com/libp2p/rust-libp2p/discussions/3526

I think what might have happened is that in the recent release, the version of one of the crates did not receive a correct semver bump. `rust-libp2p` is - similar to `webrtc` - a workspace with many crates. What happened to us previously is that we accidentally released something as a patch version despite one of the workspace dependencies having received a minor bump AND that dependency was part of the public API. I don't know how many crates participated in the last release but that might be a good starting point to look for the issue :)

melekes:
…lease interceptor 0.9.0 and rtcp 0.8.0

Fixes #412

### DIFF
--- a/data/CHANGELOG.md
+++ b/data/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ## v0.7.0
 
+* Increased required `webrtc-sctp` version to `0.8.0`.
+
 ## v0.6.0
 
 * Increased minimum support rust version to `1.60.0`.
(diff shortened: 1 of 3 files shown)


yngrtc (approved):

melekes:
@yngrtc why did you bump rtcp version https://github.com/webrtc-rs/webrtc/commit/3e5997420726078d68331f597dcd8dc5f2236845 if no changes to it were made? Also I think you forgot to add changelog entries for those crates.

PR #465, opened 2023-06-30
Add as_any to Conn trait for accessing underlying Conn

jaredwolff:
This allows you to access the inner connection state of the DTLS connection. Having the connection identity is helpful in server code.


rainliu (changes requested):
Could you consider to add as_any in Conn trait?

`    fn as_any(&self) -> &dyn std::any::Any;`

so, you can use as_any().downcast_ref
```
if let Some(dtls) = dtls_conn.as_any().downcast_ref::<DTLSConn>() {
    dtls.connection_state().await.identity_hint
}
```

rainliu (approved):

PR #470, opened 2023-07-25
Error clarification

rogurotus:
I get `failed to verify rtsp auth tag` (`RtcpFailedToVerifyAuthTag`) errors intermittently but the description of the error is not sufficient to identify the problem.

it could be:
different encryption methods,
broken package,
wrong auth tag


rainliu (approved):
LGTM

Issue #473, opened 2023-08-02
IPv6?


jamilbk:
Hey folks, thanks so much for the amazing library. We're using it over in firezone/firezone with good results so far.

However, we noticed that testing with IPv6 addresses seems to fail completely, and we suspect perhaps it is not supported.

Is this the case? If so, any hint as to how much work would be required to support it? We'd like to help out if possible.

Thanks in advance!

cc @thomaseizinger @conectado

jamilbk:
Closed via #475

thomaseizinger:
Depending on the scope, I'd say this isn't fully resolved because `webrtc-rs` doesn't seem to use the `RequestedAddressFamily` and `AdditionalAddressFamily` headers to make IPv6 allocations when necessary.

thomaseizinger:
@rainliu Thanks for landing #495. Maybe I am missing something but that only seems to address the server-side right? Can we re-open this to also track handling IPv6 on the client side, i.e. mkaing IP6 allocations?

ris-work:
I am also interested in IPv6-only deployments. I'd love to see it supported as well.

PR #475, opened 2023-08-02
fix linux ipv6 handling

conectado:
Hi!

In Linux, the last 8 octets from Ipv6 were being discarded when gathering local interfaces, making Ipv6 bindings fail.

This PR further simplifies Ip parsing.

(Relates to #473 )


rainliu (approved):
Thanks for fix

rainliu (changes requested):
Please fix clippy/ fmt error

conectado:
fixed!

rainliu (approved):

PR #479, opened 2023-08-14
Add useful derives to RTCIceServer and RTCIceCredentialType

alexipeck:


alexipeck:
If you would rather I implement Serialize and Deserialize behind a feature flag, let me know and I will implement it.

rainliu (approved):

PR #484, opened 2023-08-22
fix(android): Favor getifaddrs vs ifaddrs for android 11

jamilbk:
Remove android-specific codepath for getifaddrs

This is not supported in Android 11+. Instead, use libc getifaddr which *allegedly* works with Android 7+.

See https://github.com/messense/if-addrs/issues/14


jamilbk:
Please feel free to ignore.

PR #489, opened 2023-09-01
Support RFC 3890

skyuplam:
Support RFC 3890 (TIAS) for unmarshaling bandwidth.

### DIFF
--- a/sdp/src/description/description_test.rs
+++ b/sdp/src/description/description_test.rs
@@ -402,6 +402,15 @@ m=audio 54400 RTP/SAVPF 0 96\r\n\
 b=X-YZ:128\r\n\
 b=AS:12345\r\n";
 
+const MEDIA_TRANSPORT_BANDWIDTH_SDP: &str = "v=0\r\n\
+o=jdoe 2890844526 2890842807 IN IP4 10.47.16.5\r\n\
+s=SDP Seminar\r\n\
+t=2873397496 2873404696\r\n\
+m=video 51372 RTP/AVP 99\r\n\
+m=audio 54400 RTP/SAVPF 0 96\r\n\
+b=AS:12345\r\n\
+b=TIAS:12345\r\n";
+
 const MEDIA_ENCRYPTION_KEY_SDP: &str = "v=0\r\n\
 o=jdoe 2890844526 2890842807 IN IP4 10.47.16.5\r\n\
 s=SDP Seminar\r\n\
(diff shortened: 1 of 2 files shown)


rainliu (approved):

zhiyong0804:
Hi skyuplam,
     
    as rfc3890#section-6.2.2 descriped, TIAS has an integer bit-rate value in bits per second. in rfc4566#section-5.8, CT and AS is interpreted as kilobits per second by default, so the bandwidth value should be divided by 1000 if b type is TIAS。

`
fn unmarshal_bandwidth(value: &str) -> Result<Bandwidth> {
    let mut parts: Vec<&str> = value.split(':').collect();
    if parts.len() != 2 {
        return Err(Error::SdpInvalidSyntax(format!("`b={value}`")));
    }

    let experimental = parts[0].starts_with("X-");
    if experimental {
        parts[0] = parts[0].trim_start_matches("X-");
    } else {
        // Set according to currently registered with IANA
        // https://tools.ietf.org/html/rfc4566#section-5.8 and
        // https://datatracker.ietf.org/doc/html/rfc3890
        let i = index_of(parts[0], &["CT", "AS", "TIAS"]);
        if i == -1 {
            return Err(Error::SdpInvalidValue(parts[0].to_owned()));
        }
    }

    let bandwidth = parts[1].parse::<u64>()?;

    Ok(Bandwidth {
        experimental,
        bandwidth_type: parts[0].to_owned(),
        bandwidth,
    })
}
`

zhiyong0804:
> Hi @skyuplam,
> 
> as rfc3890 section-6.2.2 descriped, TIAS has an integer bit-rate value in bits per second. in rfc4566#section-5.8, CT and AS is interpreted as kilobits per second by default, so the bandwidth value should be divided by 1000 if b type is TIAS.
> 
> ` fn unmarshal_bandwidth(value: &str) -> Result {
> 
> ```
> // ...
> 
> let bandwidth = parts[1].parse::<u64>()?;
> 
> Ok(Bandwidth {
>     experimental,
>     bandwidth_type: parts[0].to_owned(),
>     bandwidth,
> })
> ```
> 
> } `

the correction ref : https://github.com/zhiyong0804/webrtc/commit/e38f873dcde53a0303cd93011c2d61c9c82491d8

skyuplam:
Hi @zhiyong0804 , I think this is not necessary as the bandwidth type has already implied the unit of the value according to the spec whereas the consumer of this sdp crate should be aware when using the bandwidth.

PR #497, opened 2023-09-21
allow image media type and udptl protocol in sdp parser

1HPorange:


sdp/src/description/session.rs
```diff
@@ -1231,7 +1233,7 @@ fn unmarshal_media_description<'a, R: io::BufRead + io::Seek>(
         let i = index_of(
             proto,
             &[
-                "UDP", "RTP", "AVP", "SAVP", "SAVPF", "TLS", "DTLS", "SCTP", "AVPF",
+                "UDP", "RTP", "AVP", "SAVP", "SAVPF", "TLS", "DTLS", "SCTP", "AVPF", "udptl"
```
rainliu, line 1236: Shouldn’t it be “UDPTL”?

1HPorange: Hi, your link refers to UDPTL over DTLS, which is something different (and something I've never seen in the wild).

Every implementation I have ever encountered used lowercase "udptl", which seems to be confirmed here: https://www.rfc-editor.org/rfc/rfc5347.html#section-2.5.2

The RFC however seems to recommend matching case-insensitive. Like I said, I have *never* seen anything other than "udptl" in lowercase, so I'll let you decide what you prefer.

rainliu: Got it. 

Please fix clippy/fmt error.

rainliu (approved):

PR #508, opened 2023-10-18
Remove netlink `getifaddrs` implementation on Android in favor of `libc::getifaddrs`

### LINKED ISSUE #272: Interface list is broken on Android 11+
See https://github.com/messense/if-addrs/issues/14

TLDR: on Android 11, `getifaddrs` is working, but the method used in `webrtc-util` doesn't. Since `getifaddrs` doesn't work on old Android versions (seemingly 6 and below?), perhaps both methods should be tried?

Weirdly, the current method works just fine for me on an emulator, but not on a physical device

jamilbk:
Fixes #272 
Refs #484 

On Android >= 11, the netlink method of getting `ifaddrs` fails because of restrictions introduced in Android 11+:

https://developer.android.com/training/articles/user-data-ids#mac-11-plus

As such, `webrtc-rs` is essentially broken on Android 11+. Standard `libc::getifaddrs` works fine however on Android >= 7.

This PR removes the netlink based `getifaddrs` override so that the libc version in stdlib is used instead.

We have these changes running successfully on Android 11 - 14 over at https://github.com/firezone/firezone -- see https://github.com/firezone/webrtc/pull/1

Some more reading about similar projects hitting the Android 11+ restriction:

https://github.com/golang/go/issues/40569
https://github.com/arvidn/libtorrent/issues/6251
https://github.com/Catfriend1/syncthing-android/issues/735


rainliu:
Should the build-dependencies in https://github.com/webrtc-rs/webrtc/blob/13c490c3e78303be4845a14a66df2b838d52b7cf/util/Cargo.toml#L45 be removed too?

jamilbk:
@rainliu Ah yes, good call! Updated.

rainliu:
Could you fix style due to typo check? Thanks

rainliu (approved):

PR #516, opened 2023-11-21
Implement pion commit 2184 - mux: drop packets when buffer is full

stuqdog:
This PR implements [pion commit 2184](https://github.com/pion/webrtc/pull/2184). 

Currently it's assumed that the read buffer will never fill, and so any failure to `write` should be cause to return an `Err`. This isn't always true however, particularly in high load scenarios. Per `pion` fixes, we should instead log and drop packets when the buffer is full, and continue to return an error in other failure cases.

See https://github.com/pion/webrtc/issues/2152 and https://github.com/pion/webrtc/issues/2180 for more details.

### DIFF
--- a/sctp/src/association/association_test.rs
+++ b/sctp/src/association/association_test.rs
@@ -1991,22 +1991,18 @@ async fn test_assoc_reset_close_both_ways() -> Result<()> {
     let done_ch_tx0 = Arc::clone(&done_ch_tx);
     tokio::spawn(async move {
         let mut buf = vec![0u8; 32];
-        loop {
-            log::debug!("s.read_sctp begin");
-            match s0.read_sctp(&mut buf).await {
-                Ok((0, PayloadProtocolIdentifier::Unknown)) => {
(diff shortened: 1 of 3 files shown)


sctp/src/association/association_test.rs
```diff
@@ -1991,22 +1991,18 @@ async fn test_assoc_reset_close_both_ways() -> Result<()> {
...
+            Ok((0, PayloadProtocolIdentifier::Unknown)) => {
+                log::debug!("s0.read_sctp EOF");
+                let _ = done_ch_tx0.send(Some(Error::ErrEof)).await;
+            }
+            Ok(_) => {
+                panic!("must be error");
+            }
+            Err(err) => {
+                log::debug!("s0.read_sctp err {:?}", err);
+                let _ = done_ch_tx0.send(Some(err)).await;
```
stuqdog, line 2005: (flyby) clippy was complaining that this `loop` never actually looped, so I removed it.

rainliu (approved):

PR #523, opened 2024-01-01
unify add/remove_track arguments for peer connections

jumbeldliam:
Previously, RTCPeerConnection's remove_track took a `&Arc<RTCRtpSender>` while its add_track took an `Arc<dyn TrackLocal + Send + Sync>` for its argument, and semantically it makes sense for these two methods to have the same parameters. 

This would delegate checking whether the track exists on a given sender to the user before removing the track (as seen in the changes to the example), but I think its rare that a user would want to remove a track based on the sender rather than explicitly specifying the track.


rainliu:
why didn't this RP trigger github actions Checks?

rainliu:
why didn't this RP trigger github actions Checks?

PR #534, opened 2024-02-14
Fix simulcast for firefox

haaspors:
This is a patchset which at least improves using webrtc-rs on a SFU with simulcast and firefox as a client

I found that the simulcast example didn't work for Firefox. It might not really work now either, but this should be a step on the way.
Please let me know what you think :)


haaspors:
Ping @rainliu

Please leave some feedback on what you think about this. Thanks!

rainliu (approved):

Issue #542, opened 2024-03-06
If channel is inactive, it times out and is never re-established


marcbrevoort-cyberhive:
As it says in the title, channels appear to time out after some time of inactivity (~15 minutes). Which is probably fine.

However, when traffic occurs again later on, this always occurs using SendIndication/DataIndication (i.e. plain permissions based packets) and another channel is never re-established.

While data can still flow between two peers without a channel, it is less efficient.

marcbrevoort-cyberhive:
When this occurs, the following can be seen in the logs
```
121625	2024-03-07 08:52:05.834850	192.168.0.100	18.132.207.60	STUN	198	Channel-Bind Request XOR-PEER-ADDRESS: 90.206.45.61:18025 ChannelNumber=0x4001 user: 1709887601:74134842-07f3-4d16-a5c3-33ea4c52ab03 realm: eturnal.net with nonce
121626	2024-03-07 08:52:05.857553	18.132.207.60	192.168.0.100	STUN	126	Channel-Bind Error Response realm: eturnal.net with nonce error-code: 438 (Stale Nonce) Stale Nonce
```
and there is no effort made to recover.

The client should retry, using the NONCE provided in the response (as per RFC5389).

marcbrevoort-cyberhive:
This issue is still ongoing - is there any timeline for its resolution please?

PR #543, opened 2024-03-07
Update relay_conn.rs

marcbrevoort-cyberhive:
If a channel bind (refresh) fails with a stale nonce error, the nonce should be refreshed and the attempt should be retried.

This seeks to fix https://github.com/webrtc-rs/webrtc/issues/542

### DIFF
--- a/turn/src/client/relay_conn.rs
+++ b/turn/src/client/relay_conn.rs
@@ -567,6 +567,17 @@ impl<T: RelayConnObserver + Send + Sync> RelayConnInternal<T> {
 
         let res = tr_res.msg;
 
+        if res.typ.class == CLASS_ERROR_RESPONSE {
+            let mut code = ErrorCodeAttribute::default();
+            let result = code.get_from(&res);
+            if result.is_err() {
+                return Err(Error::Other(format!("{}", res.typ)));
+            } else if code.code == CODE_STALE_NONCE {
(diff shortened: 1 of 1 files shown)


rainliu (approved):

rainliu (approved):

rainliu (changes requested):
.

rainliu (changes requested):
make sure all tests are passed.

marcbrevoort-cyberhive:
Okay, while this demonstrates what needs to happen, `bind` is actually an associated function rather than a method so it doesn't have access to `self` so the line
```
self.set_nonce_from_msg(&res);
```
throws an error.

Now, this line doesn't _have_ to be run during the bind; it can also be excuted by the caller. To avoid confusion, instead of a TryAgain I could define a new error StaleNonce(Message) which allows communicating back the nonce to be set in an error.
However since the caller is itself an async block, it too doesn't have access to `self` (unless I move it into the block which of course causes all sorts of borrow checker complaints.

_What_ needs to happen here is clear enough: if the channel bind fails with "stale nonce", it needs to be retried with a refreshed nonce, but at this point my knowledge of the webrtc source is insufficient to figure out _how_ the missing re-bind without upsetting the architecture of the crate, so I'll leave #542 open as a bug.

PR #548, opened 2024-03-26
Code improvements, part 0: Clippy

ProgramCrafter:


ProgramCrafter:
It seems I fixed warnings but in other branch, which should've gone into part 1: doc improvements. I'll open new PR for that branch, then.

PR #549, opened 2024-03-26
Code improvement, part 1: modules documentation

ProgramCrafter:
In particular, several doc-links which pointed nowhere were fixed. This code should also pass Clippy's and rustfmt's judgement.


ProgramCrafter:
Those four lints failing Clippy check are fixed! Or, more precisely, one lint was incorrect so it was locally-allowed.

rainliu (approved):

PR #557, opened 2024-04-19
fix OBS-studio DTLS failed

a-wing:
We found use OBS-studio WHIP cannot PeerConnected

package DTLS have problem

We found after this commit have this bug

This PR https://github.com/webrtc-rs/webrtc/pull/496 and this commit after https://github.com/webrtc-rs/webrtc/pull/496/commits/935283abc719b2c3a6925cd8bb90ffb3809bccc5

Finally we found the problem. But, I'm not sure this have bug 

https://github.com/RustCrypto/utils/blob/master/block-padding/src/lib.rs#L176-L216

Maybe, webrtc-rs is using this library incorrectly

https://github.com/binbat/live777/issues/39
https://github.com/binbat/live777/issues/58

### DIFF
--- a/dtls/src/crypto/crypto_cbc.rs
+++ b/dtls/src/crypto/crypto_cbc.rs
@@ -8,12 +8,13 @@
 
 // https://github.com/RustCrypto/block-ciphers
 
-use aes::cipher::{block_padding::Pkcs7, BlockDecryptMut, BlockEncryptMut, KeyIvInit};
+use aes::cipher::{BlockDecryptMut, BlockEncryptMut, KeyIvInit};
 use p256::elliptic_curve::subtle::ConstantTimeEq;
 use rand::Rng;
 use std::io::Cursor;
 use std::ops::Not;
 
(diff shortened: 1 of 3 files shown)


yngrtc:
please fix formatting

yngrtc (approved):

edaniels:
@rainliu, can you comment on why https://github.com/webrtc-rs/webrtc/commit/935283abc719b2c3a6925cd8bb90ffb3809bccc5 changed the padding in the first place? I'm not sure I understand the reasoning and this was particularly hard to track down. Thank you!

rainliu:
dtls dependent crates upgrade to new version, however, incorrect usage of new APIs of those crates causes incompatible PADDING.

aes = "0.8"
cbc = { version = "0.1", features = [ "block-padding", "alloc"] }
ccm = "0.5"

edaniels:
I see, but how is PKCS7 relevant to DTLS? Does the RFC say somewhere that the padding schemes are equivalent?

PR #568, opened 2024-05-14
Match group bundle from offer

haaspors:
Implemented like in pion

https://github.com/pion/webrtc/pull/2564


rainliu:
please fix two checks. Thanks.

Issue #577, opened 2024-06-16
Modify the webrtc RTCConfiguration and the get/setconfiguration for  fixing "whipinto and whepfrom iceserver" in our project


udrs:
we hope to modify this issue in our project(https://github.com/binbat/live777/issues/128), so we modified your "mod.rs" and "get/setconfiguration" function.
[ModifyRTCConfiguration.zip](https://github.com/user-attachments/files/15858131/ModifyRTCConfiguration.zip)

udrs:
Thank you for you great job, look forward to your check and receive.

rainliu:
thanks @udrs , the changes are merged as commit https://github.com/webrtc-rs/webrtc/commit/ae93e81a493c004b0270603d539ad9a40b7b3064

PR #581, opened 2024-06-17
resolve conflict & add setconfiguration function

udrs:
Thank you for your great project.
As I mentioned in this issue, we need to modify RTCConfiguration.
https://github.com/webrtc-rs/webrtc/issues/577

Based on the code previously provided, I have replaced println! statements with assertions.
Looking forward to your check and receive.


rainliu:
I think I already merged your changes into master based on your provide .zip file in https://github.com/webrtc-rs/webrtc/issues/577

rainliu:
feature is also included in master.

Issue #55, opened 2021-04-04
[All] add Documentations for all crates


rainliu:

morajabi:
Should this be renamed to just: Improve docs?

k0nserv:
> Should this be renamed to just: Improve docs to reflect the monorepo change.

I don't think so, but perhaps that'd make for a good new issue

overheat:
How to test STUN crate? 
Since it's related to documents I put it here.
Do I need to make a new issue?

sahilwep:
"We require comprehensive documentation detailing the proper implementation of **DTLS-SRTP** for encryption."

sahilwep:
We require comprehensive documentation detailing the proper implementation of **DTLS-SRTP** for encryption.

PR #591, opened 2024-07-01
Add spec links to public items

Serial-ATA:
This adds a `Specifications` section to a bunch of public items, linking to MDN, W3C, and RFCs.

This doesn't cover all public items, but it is a good start, since #55 is still in progress.


rainliu (approved):
Thanks for improving the documentation.

Issue #295, opened 2022-09-19
Add support for RTX via distinc SSRC


k0nserv:
We don't support RTP retransmissions(RTX) when the retransmitted packets are sent over a separate RTP stream(with a distinct SSRC). These packets are simply black holed at the moment.

For now RTX based on NACKs emitted by webrtc-rs work, but the RTX packets will be sent within the established RTP stream.

Related issues in Pion:

* https://github.com/pion/webrtc/issues/1675
* https://github.com/pion/webrtc/issues/2272

k0nserv:
I'm removing the default registration of RTX codecs in https://github.com/webrtc-rs/webrtc/pull/294

PR #607, opened 2024-08-26
Configuration of outgoing RTX streams

anders-avos:
Makes `RTCRtpSender` capable of configuring RTX streams.

Incoming RTX streams are already handled at the SDP/SDES layer and then processed by `RTCRtpReceiver::receive_for_rtx`. This means interceptors can bind to remote RTX streams. This PR aims to do the same for local streams. This is related to, but not a full fix for https://github.com/webrtc-rs/webrtc/issues/295 since it doesn't implement the retransmission, but it allows the separate RTX SSRCs to be configured, negotiated, and then bound to interceptors that can do the rest.

First, a field is added to `StreamInfo` to pass information about associated streams from the SDP to the interceptor chain. This is populated for both senders and receivers. Typically the interceptor can check if the MIME type is `*/rtx` and know that the associated stream is the original stream, but this concept can also used for other RTP extensions.

The main commit adds support for establishing RTX streams in `RTCRtpSender`. Since it's easy to unknowingly have `*/rtx` codecs configured, this feature is gated behind a new `enable_sender_rtx` setting (this can be removed if it's thought safe to always enable). Whenever an encoding is added, the sender checks if the media engine contains a suitable RTX codec. If so, a SSRC is allocated, a flow identification `ssrc-group` is added to the SDP, and interceptors are bound for the RTX streams on `start()`.

The last commit fixes an issue where `repair_ssrc` is only populated if the `a=ssrc-group:FID` line appears before the `a=ssid:` lines.

Let me know what you think about the approach!

### DIFF
--- a/interceptor/src/stream_info.rs
+++ b/interceptor/src/stream_info.rs
@@ -20,6 +20,15 @@ pub struct StreamInfo {
     pub channels: u16,
     pub sdp_fmtp_line: String,
     pub rtcp_feedback: Vec<RTCPFeedback>,
+    pub associated_stream: Option<AssociatedStreamInfo>,
+}
+
+/// AssociatedStreamInfo provides a mapping from an auxiliary stream (RTX, FEC,
+/// etc.) back to the original stream.
+#[derive(Default, Debug, Clone)]
+pub struct AssociatedStreamInfo {
+    pub ssrc: u32,
+    pub payload_type: u8,
 }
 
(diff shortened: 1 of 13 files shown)


anders-avos:
Do you think the new setting is necessary? If so I'm open to suggestions on how to get the setting into `RTCRtpSender` without tripping clippy parameter count lint. Maybe a new `RTCRtpSenderInit` struct?

rainliu (approved):

PR #609, opened 2024-09-03
webrtc: Add support for unordered unreliable data channels

### LINKED ISSUE #371: Making an unordered unreliable datachannel doesn't seem possible
Creating a datachannel with `ordered = false` and (`max_retransmits = 0` or `max_packet_lifetime = 0`) seems to fall back to making a reliable channel.

The receiving other client has `maxPacketLifeTime: null` and `maxRetransmits: null`

PaulOlteanu:
Fixes #371. Most of the change is in `RTCDataChannel::open` to match https://github.com/pion/webrtc/blob/4ef00e6e5f78647305cdd08568d16f49ae6190eb/datachannel.go#L130-L152

### DIFF
--- a/webrtc/src/data_channel/mod.rs
+++ b/webrtc/src/data_channel/mod.rs
@@ -60,8 +60,8 @@ pub struct RTCDataChannel {
     pub(crate) stats_id: String,
     pub(crate) label: String,
     pub(crate) ordered: bool,
-    pub(crate) max_packet_lifetime: u16,
-    pub(crate) max_retransmits: u16,
+    pub(crate) max_packet_lifetime: Option<u16>,
+    pub(crate) max_retransmits: Option<u16>,
     pub(crate) protocol: String,
     pub(crate) negotiated: bool,
     pub(crate) id: AtomicU16,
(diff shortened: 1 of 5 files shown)


webrtc/src/data_channel/mod.rs
```diff
@@ -137,27 +137,35 @@ impl RTCDataChannel {
...
+                (None, Some(max_retransmits)) => {
+                    reliability_parameter = max_retransmits as u32;
+                    if self.ordered {
+                        channel_type = ChannelType::PartialReliableRexmit;
+                    } else {
+                        channel_type = ChannelType::PartialReliableRexmitUnordered;
+                    }
                 }
+
+                (Some(_), Some(_)) => unreachable!(),
```
rainliu, line 168: unreachable! is not acceptable here, since DataChannel library user may pass both  max_packet_lifetime and max_retransmits with values.

please keep the same logic as before, but with Option check.

rainliu (changes requested):

webrtc/src/data_channel/mod.rs
```diff
@@ -137,26 +137,32 @@ impl RTCDataChannel {
...
+                    }
                 }
-            } else {
-                reliability_parameter = self.max_packet_lifetime as u32;
-                if self.ordered {
-                    channel_type = ChannelType::PartialReliableTimed;
-                } else {
-                    channel_type = ChannelType::PartialReliableTimedUnordered;
+
+                (None, Some(max_packet_lifetime)) => {
```
rainliu, line 159: change this line to 
```
_ => {
```

PaulOlteanu: All the cases are covered (and I compiled and ran tests locally fine). The CI failures aren't related to my changes

rainliu (approved):

PR #611, opened 2024-09-05
dtls: fix race condition in DTLSConn::close

feschber:
a context switch to a simultaneous close operation between `closed.load()` and `closed.store()` would lead to an additional unwanted close operation on the inner connection


rainliu:
please fix cargo clippy/fmt check. Thanks.

feschber:
> please fix cargo clippy/fmt check. Thanks.
looks like that is unrelated to this PR

rainliu (approved):

PR #617, opened 2024-10-01
AV1: Correctly parse 5-byte leb128 size value

nwolverson:
The parsing of leb128 size value a 2-pass approach rather than following the pseudocode in the AV1 spec, presumably for performance reasons. However while the maximum permitted value is indeed 32 bits, the input bytes accumulated in a `u32` are packed 7 bits per byte, thus losing relevant bits from the first byte for 5-byte values. (I believe the leb128 value can be up to 8 bytes with leading zeroes but perhaps that case was fine).

I imagine this doesn't come up where the the only large OBU in a binary presented to webrtc-rs has the size omitted, we hit this with an encoder which sets the size on every OBU.


rainliu:
Please fix clippy/fmt warnings

rainliu (approved):

PR #623, opened 2024-10-16
Don't send receiver reports for SSRC during simulcast probe

anders-avos:
libwebrtc will stop sending SDES headers after the first receiver report on an SSRC[^1]. This breaks simulcast probing if the receiver report is sent before probing is completed. This change delays receiver reports for probing SSRCs until the first non-probe packet is received.

I didn't find any easy way to get this information into the interceptor, so I opted for using attributes. Let me know if you prefer some other means.

[^1]: https://github.com/webrtc-sdk/webrtc/blob/0ae5688d4d490274ea47fee3f87f622ef3f13113/modules/rtp_rtcp/source/rtp_sender.cc#L509

### DIFF
--- a/interceptor/src/lib.rs
+++ b/interceptor/src/lib.rs
@@ -24,6 +24,9 @@ pub mod twcc;
 
 pub use error::Error;
 
+/// Attribute indicating the stream is probing incoming packets.
+pub const ATTR_READ_PROBE: usize = 2295978936;
+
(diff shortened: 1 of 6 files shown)


anders-avos:
Any updates on this? I'm happy to make changes if there are other ideas on how to propagate the probe status into the interceptor!

PR #625, opened 2024-11-02
Bugfix/133- fix util unit test memory leak

mutexd:
I use this crate in my project which is sort of a combination of `rtp-to-webrtc` and `rtp-forwarder`. I came to realize that there's memory leak issue and I wish to tackle it. I bumped into Issue#133 and I guess I could start with it and see if some memory leaks can be resolved by tackling it.

According to Issue#133. There is memory leak in Util/Turn/Ice/Interceptor detected by nightly-rusts sanitizer. 
This PR targets the memory leak issue in Util and hopefully we can tackle other modules in the near future.

- `vnet` is used in unit test and several cyclic dependencies are discovered and resolved by using `Weak<>`
- `util::UdpConn` owns a map of UdpConn and itself is also added into the map, which creates a cyclic dependency of `Arc<>` and thus the memory will never be released. The way I see it, Listener already owns the table and should take care of clean-up. The only operation `util::UdpConn` do to the map is clean-up the map upon `close`. In my opinion it's superfluous and we should just remove the map from `util::UdpConn`.

`RUSTFLAGS="-Z sanitizer=leak" cargo test` in util shows no sign of memory leak anymore.

### DIFF
--- a/util/src/conn/conn_udp_listener.rs
+++ b/util/src/conn/conn_udp_listener.rs
@@ -211,7 +211,7 @@ impl ListenConfig {
             }
         }
 
-        let udp_conn = Arc::new(UdpConn::new(Arc::clone(pconn), Arc::clone(conns), raddr));
+        let udp_conn = Arc::new(UdpConn::new(Arc::clone(pconn), raddr));
         {
             let accept_ch = accept_ch_tx.lock().await;
             if let Some(tx) = &*accept_ch {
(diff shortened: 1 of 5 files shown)


rainliu:
Could you fix clippy warning?

mutexd:
clippy warning fixed.

rainliu (approved):

PR #673, opened 2025-05-19
Move dtls encryption into blocking async tasks

wdouglass:
This should help prevent the runtime from holding up other tasks. I'm not sure if `spawn_blocking` should be used, or if `block_in_place` is more appropriate, but these seem like computationally expensive blocks. Also, I added some preallocation of vectors to reduce the number of internal memcpys that have to happen. I appreciate any review and feedback, and if I've broken the project workflow with this PR, please point me in the right direction! Thank you.

### DIFF
--- a/dtls/src/conn/mod.rs
+++ b/dtls/src/conn/mod.rs
@@ -713,7 +713,8 @@ impl DTLSConn {
 
             //p.record.record_layer_header = record_layer_header;
 
-            let mut raw_packet = vec![];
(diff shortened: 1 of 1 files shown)


wdouglass:
This has negatively affected throughput, and i am continuing to investigate. As part of that investigation, i have enabled `console_subscriber`, and found a possible deadlock in the SCTP implementation. I'm investigating this now.

PR #675, opened 2025-05-23
Implement ChaCha20_Poly1305 cipher suite

### LINKED ISSUE #674: Support ChaCha20 as a dtls cipher suite
I have a system that would benefit from ChaCha20 (as opposed to AES) for DTLS encryption, and plan on attempting to implement it, per RFC9147. I'm adding this issue for discussion, and to track my progress in this task.

wdouglass:
This is a cipher suite supported by Chrome, and preferred on some low-end processors. I've also added an argument to the dial_selfsign example, which forces a cipher suite, and used that argument to test against the DTLS implementation in OpenSSL. Any feedback is appreciated, thank you!

Fixes #674


rainliu:
please fix clippy/fmt warning

rainliu (approved):

PR #687, opened 2025-06-05
turn: use ok_or() to simplify get_allocation() call

cppcoffee:
Simplify the logic of `if let Some(a) = a { ... } else { ... }`


cppcoffee:

rainliu (approved):

PR #689, opened 2025-06-08
turn: use ok_or() to simplify command_tx

cppcoffee:
use `command_tx.clone().ok_or(Error::ErrClosed)?` to replace `if let Some(..) { ... } else { return Err(Error::ErrClosed) }`


cppcoffee:

PR #693, opened 2025-06-11
mdns: simplify resource pack function

cppcoffee:


cppcoffee:
hi @rainliu , can you revew these PR?

PR #702, opened 2025-06-30
mdns: allow reuse_port on non-Unix platforms when feature is enabled

cppcoffee:
Change conditional compilation from requiring both unix target family and reuse_port feature to allowing either condition.


cppcoffee:

CI Reporting error have been fixed in the following PR:

https://github.com/webrtc-rs/webrtc/pull/703

rainliu (approved):

Issue #705, opened 2025-07-10
SCTP: Bug: Fixed sctp ports of 5000 breaks datachannels if any other port is negotiated


KillingSpark:
I noticed that after negotiating sctp ports different from 5000 (to get nicer filtering in wireshark) this library stopped opening datachannels. This is caused by [this function](https://github.com/webrtc-rs/webrtc/blob/f296fc7295e64804a1bd57b7eed82f61fc4210b4/sctp/src/association/association_internal.rs#L208
): 

```rust
    pub(crate) fn send_init(&mut self) -> Result<()> {
        if let Some(stored_init) = self.stored_init.clone() {
            log::debug!("[{}] sending INIT", self.name);

            self.source_port = 5000; // Spec??
            self.destination_port = 5000; // Spec??

            let outbound = Packet {
                source_port: self.source_port,
                destination_port: self.destination_port,
                verification_tag: 0,
                chunks: vec![Box::new(stored_init)],
            };

            self.control_queue.push_back(outbound);
            self.awake_write_loop();

            Ok(())
        } else {
            Err(Error::ErrInitNotStoredToSend)
        }
    }
```

Which causes an init on the wrong ports, which results in an abort later on.

I think this can be fixed rather easily by passing the negotiated ports to this function. This is (outside of tests) only called from `start_sctp` which in turn is only called here:

```rust
    if let Some(parsed) = &remote_desc.parsed {
            if have_application_media_section(parsed) {
                self.start_sctp().await;
            }
        }
```

Instead of just checking if there is an application section in the remote description this could look at both the remote and local description to get the correct sctp ports and pass them to this function down to the sctp transport via two new fields in the `sctp::association::Config` struct.

I'd be willing to make a PR, just wanted to get your opinions first.

PR #706, opened 2025-07-16
do not used fixed sctp-ports 5000, but consult the negotiated ports

KillingSpark:
AS described in #705 this library currently fixes the sctp port used for datachannels as 5000 for both parties even if the negotiation says otherwise. This PR fixes that behaviour, consulting the negotiated ports and passing them to the sctp subsystem.


yngrtc (approved):
nit: please fix clippy & fmt warnings

KillingSpark:
The CI is stuck on a specific test, but I was able to confirm that this test sometimes hangs, even on commit 4327e6ea without any of the changes in this PR.

PR #719, opened 2025-08-22
[ice] Fix unhandled overflow panic in `listen_udp_in_port_range`

lucamuscat:
When `port_start` is assigned the value `u16::MAX`, and the socket cannot be bound to port `u16::MAX`, an unhandled overflow will occur.

When building webrtc-rs using the debug profile, this overflow will lead to a panic, and worse yet in the release profile, a silent overflow and wrap around will occur, allowing `listen_udp_in_port_range` to bind to a port outside of the specified port range.

### DIFF
--- a/ice/src/util/mod.rs
+++ b/ice/src/util/mod.rs
@@ -163,7 +163,8 @@ pub async fn listen_udp_in_port_range(
             Err(err) => log::debug!("failed to listen {laddr}: {err}"),
         };
 
-        port_current += 1;
+        port_current = port_current.checked_add(1).unwrap_or(i);
+
         if port_current > j {
             port_current = i;
         }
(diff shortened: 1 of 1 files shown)


ice/src/util/mod.rs
```diff
@@ -163,7 +163,8 @@ pub async fn listen_udp_in_port_range(
             Err(err) => log::debug!("failed to listen {laddr}: {err}"),
         };
 
-        port_current += 1;
+        port_current = port_current.checked_add(1).unwrap_or(i);
```
lucamuscat, line 166: If an overflow were to occur when incrementing `port_current`, we should wrap around to `port_min` instead of panicing on debug or wrapping around to zero on release.

rainliu (approved):

PR #722, opened 2025-08-25
Make use of the `max-message-size` SDP attribute in datachannels

lucamuscat:
Currently, webrtc-rs does not react to the `max-message-size` attribute that may be attached to an SDP.

RFC8841 introduces the `max-message-size` SDP attribute:
> The attribute can be associated with an "m=" line to indicate the maximum SCTP user message size (indicated in bytes) that an SCTP endpoint is willing to receive on the SCTP association associated with the "m=" line. Different attribute values can be used in each direction [Source](https://datatracker.ietf.org/doc/html/rfc8841#name-general-4). Browsers such as chrome typically advertise a max message size of 256KiB. Some user agents may also request a max-message-size smaller than the default 64KiB.

### DIFF
--- a/webrtc/src/api/setting_engine/mod.rs
+++ b/webrtc/src/api/setting_engine/mod.rs
@@ -54,6 +54,29 @@ pub struct ReplayProtection {
     pub srtcp: usize,
 }
 
+#[derive(Clone)]
+pub enum SctpMaxMessageSize {
+    Bounded(u32),
+    Unbounded,
+}
+
+impl SctpMaxMessageSize {
+    pub fn as_u32(&self) -> u32 {
+        match self {
+            Self::Bounded(result) => *result,
+            Self::Unbounded => 0,
+        }
+    }
+}
+
+impl Default for SctpMaxMessageSize {
+    fn default() -> Self {
+        // https://datatracker.ietf.org/doc/html/rfc8841#section-6.1-4
+        // > If the SDP "max-message-size" attribute is not present, the default value is 64K.
+        Self::Bounded(65536)
+    }
+}
+
 /// SettingEngine allows influencing behavior in ways that are not
 /// supported by the WebRTC API. This allows us to support additional
 /// use-cases without deviating from the WebRTC API elsewhere.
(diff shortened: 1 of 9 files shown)


webrtc/src/api/setting_engine/mod.rs
```diff
@@ -54,6 +54,30 @@ pub struct ReplayProtection {
     pub srtcp: usize,
 }
 
+#[derive(Clone)]
+pub enum SctpMaxMessageSize {
```
lucamuscat, line 58: I am personally not a huge fan of how the crate makes use of `0` as a sentinel value for some configuration being unset, so I decided to make use of an enum to give the user a clear distiction between having a bounded `SctpMaxMessageSize`, or an unbounded one (which is typically represented as a zero in this code base).

webrtc/src/data_channel/data_channel_test.rs
```diff
@@ -1375,6 +1376,203 @@ async fn test_data_channel_non_standard_session_description() -> Result<()> {
...
+    // A buffer with a size greater than the default size of 64KB.
+    let buffer = vec![0; 68000];
+    let bytes = bytes::Bytes::copy_from_slice(buffer.as_slice());
+    data_channel.send(&bytes).await.unwrap();
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_given_remote_max_message_size_is_none_when_data_channel_can_send_max_message_size_respected_on_send(
```
lucamuscat, line 1453: The following tests whether the `max-message-size` attribute, or the `sctp_max_message_size_can_send` attribute in the SettingsEngine is actually respected (in an end to end manner).

webrtc/src/sctp_transport/mod.rs
```diff
@@ -338,9 +342,9 @@ impl RTCSctpTransport {
             .store(Some(Arc::new(Mutex::new(f))));
     }
 
-    fn calc_message_size(remote_max_message_size: usize, can_send_size: usize) -> usize {
+    fn calc_message_size(remote_max_message_size: u32, can_send_size: u32) -> u32 {
```
lucamuscat, line 345: I decided to use `u32` instead of `usize` here, since the method is only called by the code added in this PR. It feels pretty pointless to convert all u32s into usizes and then convert the resulting usize back into a u32.

rainliu (approved):

PR #751, opened 2025-11-03
Make psk callback async-capable

mdelete:
With this change it is possible to request the **psk** from an external system using the hint, e.g. a database, by making the callback return a _future_, thus being able to use _await_ in the callback.


mdelete:
Please reject. I have to amend this, I did run the tests but didn't fix and run all the examples involving psk.

PR #771, opened 2026-01-17
sdp: add UnmarshalConfig for deserialization

DanglingPointer:
Sometimes an SDP contains proprietary protocols not listed in https://datatracker.ietf.org/doc/html/rfc8866#section-5.14 and/or not registered with IANA. To be able to parse such SDPs, this change adds `UnmarshalConfig` which allows to specify a custom list of allowed protocols in addition to those accepted by default.


yngrtc:
no longer accept feature development on master branch yet.

Issue #778, opened 2026-02-25
Fix localhost ip 127.0.0.1 takes too long to recv RTCIceGatheringState::Complete,  when stun:stun.l.google.com:19302 is set


yngrtc:

rainliu:
should be fixed in commit a6600fb86b9b9b58c31e63fb9c8adc5faf3c789a

Issue #776, opened 2026-02-12
RTCDataChannel transitions to Closed state after a single failed delivery with max_retransmits(0)


tasramazan625-bit:
### **Description**
I am reporting a behavior in `webrtc-rs` that seems to violate the expected behavior of unreliable Data Channels as defined in **RFC 8831**. 

When a Data Channel is configured with `ordered = false` and `max_retransmits = 0`, a single failed transmission attempt (due to transient network loss) causes the channel to transition to the `Closed` state immediately, rather than simply dropping the packet.

### **RFC References**
1.  **[RFC 8831, Section 6.1](https://datatracker.ietf.org/doc/html/rfc8831#section-6.1):**
    > *"Limiting the number of retransmissions to zero, combined with unordered delivery, provides a UDP-like service..."*
    
    A UDP-like service should inherently tolerate packet loss. Closing the entire association due to a single dropped packet contradicts the core purpose of "Partial Reliability".

2.  **[RFC 4960 (SCTP), Section 8](https://datatracker.ietf.org/doc/html/rfc4960#section-8):**
    SCTP should only fail an association after reaching `Association.Max.Retransmits` thresholds. A packet that is not retransmitted because of the **PR-SCTP** policy (as per **RFC 3758**) should be silently discarded by the sender's stack once the policy is met, without terminating the SCTP association or closing the Data Channel.

### **Observed Behavior**
In `webrtc-rs`, if I simulate a network drop (e.g., `sudo iptables -A OUTPUT -o lo -j DROP`) and call `data_channel.send()`, the channel state moves to `Closed` immediately. 

* **Comparison with other stacks (C++, C#, aiortc):** The channel stays `Open` and simply discards the data (or waits for SCTP timeouts), allowing the service to resume once the network is restored without requiring a new handshake.
* **Current `webrtc-rs` behavior:** Single failure leads to terminal state.

### **Steps to Reproduce**
1. Initialize a Data Channel with the following settings:
   ```rust
   let mut options = RTCDataChannelInit::default();
   options.ordered = Some(false);
   options.max_retransmits = Some(0);
2. Establish a successful Peer Connection and wait for the Data Channel to be Open.
3. Block network traffic to simulate a loss of connectivity.
4. Attempt to send a message via data_channel.send_text("test").

### Expected Result
The Data Channel should remain in the Open state. The packet should be discarded by the SCTP stack due to the 0-retransmission policy, but the channel should remain active for future attempts.

### Actual Result
[...]

Issue #774, opened 2026-02-10
Fix IPV6 ICE gather failure issue


yngrtc:
[2026-02-10T00:23:39Z DEBUG webrtc::ice_gatherer] Resolved STUN server stun.l.google.com:19302 to 74.125.250.129:19302
[2026-02-10T00:23:39Z DEBUG webrtc::ice_gatherer] STUN client bound to 0.0.0.0:46260
[2026-02-10T00:23:39Z DEBUG webrtc::ice_gatherer] Resolving STUN server: stun.l.google.com:19302
[2026-02-10T00:23:39Z DEBUG webrtc::ice_gatherer] Resolved STUN server stun.l.google.com:19302 to [2001:4860:4864:5:8000::1]:19302
[2026-02-10T00:23:39Z DEBUG webrtc::ice_gatherer] STUN client bound to [::]:45307
[2026-02-10T00:23:39Z TRACE webrtc::peer_connection_driver] Sent 20 bytes to 74.125.250.129:19302 from 0.0.0.0:46260
[2026-02-10T00:23:39Z ERROR webrtc::peer_connection_driver] Failed to send to [2001:4860:4864:5:8000::1]:19302 from [::]:45307: Network is unreachable (os error 101)
[2026-02-10T00:23:39Z TRACE webrtc::peer_connection_driver] Received 32 bytes from 74.125.250.129:19302
[2026-02-10T00:23:39Z ERROR webrtc::peer_connection] I/O error: the DTLS transport has not started yet
test test_stun_gathering_with_google_stun has been running for over 60 seconds

PR #785, opened 2026-04-01
fix: resolve ICE gathering delays, IPv6 failures, socket recv crashes, and DataChannel close bug (#774 #776 #777 #778)

nightness:
## Summary

This PR fixes four related bugs across the ICE gatherer, UDP mux, and RTCDataChannel layers.

---

### #777 — Better handling on socket recv error

**Root cause:** `ice/src/udp_mux/mod.rs` — the `start_conn_worker` recv loop breaks permanently on *any* non-timeout error, orphaning all ICE connections multiplexed over that socket.

**Fix:** Classify errors as transient vs. fatal. `ConnectionReset`, `ConnectionAborted`, `NetworkUnreachable`, and `WouldBlock` are now logged as warnings and the loop continues. Only genuinely fatal errors (e.g. socket closed, permission denied) stop the worker.

---

### #776 — RTCDataChannel transitions to Closed after a single failed delivery with max_retransmits(0)

**Root cause:** `data/src/data_channel/mod.rs` — `read_data_channel()` calls `self.close()` (which sends an SCTP stream reset to the remote) on *any* error returned by `stream.read_sctp()`, including the non-fatal `ErrShortBuffer`. This incorrectly closes the channel when the receive buffer is simply too small for a received message — a condition that should never terminate an unreliable data channel per [RFC 8831 §6.6](https://datatracker.ietf.org/doc/html/rfc8831#section-6.6).

**Fix (two layers):**
- `data/src/data_channel/mod.rs` — `ErrShortBuffer` is now returned to the caller without triggering a stream reset. Only errors that indicate a fatal stream condition call `self.close()`.
- `webrtc/src/data_channel/mod.rs` — The read_loop `continue`s on `ErrShortBuffer` (logging a warning) instead of breaking out of the loop and transitioning the channel to `Closed`.

---

### #778 — Fix localhost IP 127.0.0.1 takes too long to receive RTCIceGatheringState::Complete when a STUN server is set

**Root cause:** `ice/src/agent/agent_gather.rs` — `gather_candidates_internal()` uses a `WaitGroup` that blocks until *all* spawned gathering tasks finish. Each STUN URL spawns a task with a 5-second `STUN_GATHER_TIMEOUT`. If the STUN server is slow or unreachable, `GatheringState::Complete` is delayed by `N_urls × 5s` even though host candidates (including loopback) were ready immediately.

**Fix:** Added `AgentConfig::candidate_gather_timeout: Option<Duration>` (defaults to 10 seconds) and exposed it via `SettingEngine::set_candidate_gather_timeout()`. The `wg.wait()` call is now wrapped with `tokio::time::timeout`, so gathering always completes within the configured deadline regardless of STUN server responsiveness.

```rust
// Example: faster completion in localhost-only environments
let mut se = SettingEngine::default();
se.set_candidate_gather_timeout(Some(Duration::from_secs(3)));
```

---

### #774 — Fix IPv6 ICE gather failure

**Root cause:** `gather_candidates_srflx()` spawns one task per `(network_type × STUN URL)` combination. On systems without IPv6 or with a misconfigured IPv6 stack, the `net.resolve_addr()` call for IPv6 STUN URLs can hang for an extended period, blocking the WaitGroup and delaying `GatheringState::Complete` by the full `STUN_GATHER_TIMEOUT` per URL.

**Fix:** Added `STUN_DNS_TIMEOUT = 3s` around `net.resolve_addr()`. A hanging or slow IPv6 DNS lookup now fails fast with a warning log, releasing the WaitGroup worker promptly.

---

## Files Changed

| File | Change |
|------|--------|
| `ice/src/udp_mux/mod.rs` | Transient recv error classification (#777) |
| `data/src/data_channel/mod.rs` | Don't close stream on `ErrShortBuffer` (#776) |
| `webrtc/src/data_channel/mod.rs` | Continue read_loop on `ErrShortBuffer` (#776) |
| `ice/src/agent/agent_config.rs` | Add `candidate_gather_timeout` field (#778) |
| `ice/src/agent/mod.rs` | Store and pass through `candidate_gather_timeout` (#778) |
| `ice/src/agent/agent_gather.rs` | Wrap `wg.wait()` with timeout; add DNS timeout (#778 #774) |
| `webrtc/src/api/setting_engine/mod.rs` | Expose `set_candidate_gather_timeout()` to users (#778) |
| `webrtc/src/ice_transport/ice_gatherer.rs` | Wire `candidate_gather_timeout` into `AgentConfig` (#778) |
| `examples/examples/regression-checks/` | Self-contained in-process regression checks (new) |

## Test plan

- [x] `cargo build` — clean build, no warnings introduced
- [x] `cargo test -p webrtc-ice -p webrtc-data -p webrtc-sctp` — all 124 tests pass
- [x] `cargo run --example regression-checks` — all three in-process checks pass:
  - **#776**: DataChannel with `ordered=false, max_retransmits=0` sends 20 messages; `on_close` fires 0 times during the test window
  - **#778**: ICE gathering with an unreachable STUN server (`192.0.2.1`) completes in ≤4 s with `candidate_gather_timeout(4s)`
  - **#774**: ICE gathering with a real STUN URL completes in <1 s even when IPv6 DNS may be slow (DNS timeout = 3 s)

### DIFF
--- a/data/src/data_channel/mod.rs
+++ b/data/src/data_channel/mod.rs
@@ -174,8 +174,14 @@ impl DataChannel {
                     return Ok((0, false));
                 }
                 Ok((n, ppi)) => (n, ppi),
+                Err(err @ sctp::Error::ErrShortBuffer { .. }) => {
+                    // Non-fatal: the caller's buffer was too small for this message.
+                    // Return the error without resetting the stream so the channel
+                    // stays open for future messages.
+                    return Err(err.into());
+                }
                 Err(err) => {
-                    // Shutdown the stream and send the reset request to the remote.
+                    // Fatal stream error – shut down and send a reset request.
                     self.close().await?;
                     return Err(err.into());
                 }
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -151,3 +151,8 @@ bench = false
 name = "ice-restart"
 path = "examples/ice-restart/ice-restart.rs"
 bench = false
+
+[[example]]
+name = "regression-checks"
+path = "examples/regression-checks/regression-checks.rs"
+bench = false
--- a/examples/examples/regression-checks/regression-checks.rs
+++ b/examples/examples/regression-checks/regression-checks.rs
@@ -1,0 +1,333 @@
+//! Regression checks for issues #774, #776, #777, #778.
+//!
+//! Run with:
+//!   cargo run --example regression-checks
+//!
+//! All three checks run in-process; no browser or external signaling needed.
+
+use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use anyhow::Result;
+use tokio::sync::Notify;
+use webrtc::api::interceptor_registry::register_default_interceptors;
+use webrtc::api::media_engine::MediaEngine;
+use webrtc::api::setting_engine::SettingEngine;
+use webrtc::api::APIBuilder;
+use webrtc::data_channel::data_channel_init::RTCDataChannelInit;
+use webrtc::data_channel::data_channel_message::DataChannelMessage;
+use webrtc::ice_transport::ice_server::RTCIceServer;
+use webrtc::peer_connection::configuration::RTCConfiguration;
+use webrtc::peer_connection::RTCPeerConnection;
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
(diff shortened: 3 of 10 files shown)


nightness:
Closing — this PR targeted v0.17.x which is no longer the development base. The core ICE/DataChannel stack has been rewritten as Sans-IO in the `rtc` submodule on master, which addresses these bug classes architecturally. Future fixes will target master.

PR #790, opened 2026-04-01
fix(runtime/smol): eliminate lost-wakeup race and mutex-poison panic

nightness:
## Summary

Two correctness bugs in the smol runtime backend:

### 1. `SmolJoinHandle` — mutex poison panic
`detach()` and `abort()` used `.lock().unwrap()` on a `std::sync::Mutex`. If another thread panicked while holding the lock, this would trigger a **double-panic** and abort the process. Fixed with `.unwrap_or_else(|e| e.into_inner())` which recovers the data even from a poisoned mutex.

### 2. `SmolNotify` — lost-wakeup race (permanent task hang)
The previous implementation used `try_lock()` in `notify_one()` / `notify_waiters()`. If the lock was held by a concurrent `notified()` call registering itself as a waiter, the notification was **silently discarded** — causing any task waiting on `notified()` to hang permanently.

**New design** adds an `AtomicBool` "pending" flag:
- `notify_*` sets the flag (`Release`) **before** attempting to acquire the waiter list — so even if the lock is contended, a concurrent `notified()` will see the flag.
- `notified()` checks the flag (`AcqRel` swap) **before and after** acquiring the lock, so it cannot miss a notification that arrived between the two checks.

Also eliminates the duplicated `AsyncNotify` impl body by delegating to the `SmolNotify` inherent methods.

## Test Plan

- [x] `cargo build` (tokio, default) passes
- [x] `cargo build --no-default-features --features runtime-smol` passes

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "rtc"]
 	path = rtc
-	url = https://github.com/webrtc-rs/rtc
+	url = https://github.com/Brainwires/webrtc-rs-rtc
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -145,3 +145,8 @@ bench = false
 name = "swap-tracks"
 path = "examples/swap-tracks/swap-tracks.rs"
 bench = false
+
+[[example]]
+name = "data-channel-with-video"
+path = "examples/data-channel-with-video/data-channel-with-video.rs"
+bench = false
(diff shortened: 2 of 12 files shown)


nightness:
Superseded by #795 which consolidates all driver robustness fixes into a single PR.

nightness:
Superseded by #795 which is a properly scoped PR for these driver/runtime fixes.

PR #792, opened 2026-04-01
fix(driver): configurable mDNS socket-creation error handling

nightness:
## Summary

- Adds `with_mdns_fail_on_socket_error(bool)` to `PeerConnectionBuilder`
- When `true`, a failure to create the mDNS multicast socket is returned as an error from `build()` instead of being silently demoted to a `warn!` log
- Defaults to `false` — existing behaviour is completely unchanged
- Allows production deployments that depend on mDNS to detect misconfiguration at startup rather than silently operating without it

## Test plan

- [ ] `cargo build` passes
- [ ] `cargo test --workspace` passes
- [ ] Verify that with `with_mdns_fail_on_socket_error(false)` (default), a socket failure still produces a `warn!` and `build()` succeeds
- [ ] Verify that with `with_mdns_fail_on_socket_error(true)`, a socket failure causes `build()` to return `Err`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "rtc"]
 	path = rtc
-	url = https://github.com/webrtc-rs/rtc
+	url = https://github.com/Brainwires/webrtc-rs-rtc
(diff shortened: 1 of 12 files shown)


nightness:
Superseded by #795 which consolidates all driver robustness fixes into a single PR.

nightness:
Superseded by #795 which is a properly scoped PR for these driver/runtime fixes.

Issue #754, opened 2025-11-24
[SCTP] Expose MTU in `sctp::association::Config`


oteffahi:
Hello,

Is there a reason why the MTU is not configurable on the `webrtc-sctp` crate?

Correct me if I'm wrong, but the crate and its supported connections don't seem to implement PMTUD, and therefore the connection remains stuck on the default `INITIAL_MTU=1228`. I understand that the currently hardcoded value works for WebRTC purposes, but as it stands it is just too small for our uses-cases.

it also seems that the Pion implementation exposes this: https://github.com/pion/sctp/blob/b1a66a4b0fe20254da90495abf4013feac6815fd/association.go#L275

Thank you for your time.

PR #807, opened 2026-06-10
sctp: lower INITIAL_MTU to 1191 and expose the MTU in association::Config

### LINKED ISSUE #806: bug: hardcoded `INITIAL_MTU = 1228` emits packets too big for 1280 byte paths, causing a silent, permanent data-channel stall
## Summary

`webrtc-sctp` sizes outgoing packets against a hardcoded `INITIAL_MTU = 1228` (`sctp/src/association/mod.rs`). There is no PMTUD, no ICMP feedback handling, and retransmissions reuse the original chunk sizes, so the value never adapts.

A full-size SCTP packet becomes a 1265 byte UDP payload after DTLS (1228 + 13 header + 8 nonce + 16 tag, measured). That is a 1293 byte IPv4 packet or a 1313 byte IPv6 packet. Both exceed 1280, the IPv6 minimum MTU and the tunnel MTU of WireGuard-family VPNs (Tailscale etc.).

Delivery of every message large enough to fill a chunk then depends on IP fragmentation surviving the path. Where it does not ([RFC 8900](https://datatracker.ietf.org/doc/html/rfc8900) documents how common that is) we end up in a bad failure mode:

1. Fragments are dropped. Nothing tells the sender.
2. SCTP retransmits the same-size packet into the same hole forever.
3. On an ordered channel, every later message head-of-line blocks behind the wedged one.

From the application's view: `send()` succeeds, `buffered_amount` climbs and freezes, the peer receives nothing, no error is ever raised on either side.

## Real-world case

We hit this in production: a webrtc-rs sender streaming HTTP responses over a data channel to an iPad whose ICE-nominated pair was Tailscale over IPv6. Tailscale drops all IPv6 fragments by design (its filter classifies the v6 Fragment extension header as an unmatchable protocol), so the stall was deterministic on that pair: small control frames arrived, every full-size DATA chunk vanished, `buffered_amount` froze at ~8 KB, and the session died and reconnected in a loop. 

Full evidence (pcaps, receiver-side counters, writeup): https://github.com/phact/mtu-webrtc-bug

libwebrtc avoids this class of failure by keeping its SCTP packet budget near 1200 bytes, which keeps even the IPv6 worst case under 1280. webrtc-rs is 28 bytes less conservative and cannot detect when that matters.

## Reproduction

https://github.com/phact/mtu-webrtc-bug: `./repro.sh` runs the full A/B on Linux with no VPN or extra devices. A small UDP relay models a 1280 byte path; the webrtc-rs sender wedges after the first small message while the relay logs the same 1265 byte datagram retransmitted into the hole. A control run with the cap lifted delivers everything in seconds. A pcap of the real Tailscale stall is included under `diagnostics/`.

Observed with `webrtc = 0.13` / `webrtc-sctp = 0.12`. The constant and its derivation are unchanged in `webrtc-rs/rtc` master (`rtc-sctp/src/config.rs`).

## Relation to #754

#754 asked to expose the MTU in `sctp::association::Config` and was closed as not planned. That request wanted a larger value for throughput on big-MTU paths. This report is the inverse and is about correctness: the current default is too large for common real paths, and without either a config knob or PMTUD the failure is silent and permanent. Worth reconsidering with that framing.

## Suggested fixes, smallest first

1. Lower the default so the worst-case wire packet fits 1280: `1280 - 40 (IPv6) - 8 (UDP) - 37 (DTLS record overhead) = 1195` byte SCTP packet budget. One-line change, matches libwebrtc's posture, costs ~3% per-chunk overhead.
2. Expose the MTU / `max_payload_size` in `Config` (pion/sctp does).
3. DPLPMTUD (RFC 8899): probe with padded HEARTBEATs and converge on the real path MTU. The long-term correct answer; pion has prior art.

Happy to PR 1 and/or 2 if maintainers are open to it.

Filed a related issue with tailscale here https://github.com/tailscale/tailscale/issues/20083
Did a [writeup of our incident](https://www.p2claw.com/blog/2026-06-09-the-ipad-was-on-tailscale/) if that's of interest.

phact:
Fixes #806. Two commits, separable:

*Part 1 - Lower INITIAL_MTU from 1228 to 1191:*

At 1228 a full-size SCTP packet is a 1265 byte UDP payload after DTLS, i.e. a 1313 byte IPv6 packet, over the 1280 byte IPv6 minimum MTU and WireGuard-family tunnel MTU.

Prior art:

1191 is libwebrtc's kMaxSafeMTUSize, same derivation: 1280 - 40  (IPv6) - 8 (UDP) - 4 (TURN ChannelData) - 13 (DTLS record header) - 24 (GCM cipher).

The current 1228 (inherited from pion) looks like the same derivation with the DTLS term left out: 1280 - 40 (IPv6) - 8 (UDP) - 4 (TURN) = 1228.

A conservative default matters because PMTUD isn't realistically coming: the ask has been open in the parent project since 2018 (pion/sctp#12).

*Part 2 - Expose the MTU in association::Config as mtu:*

Re-opens the ask from #754, this time motivated by correctness rather than throughput. Follows the existing zero-means-default convention of max_receive_buffer_size and max_message_size; includes a regression test.

Prior art:

MTU has been configurable in pion since April 2025 https://github.com/pion/sctp/commit/dfcf6c4a8f9b13cee62de27f34b9970b7e1e6b52.

**Note**: commit 2 is technically breaking for struct-literal constructors of Config (they need mtu: 0 for the previous behavior); changelog has a Breaking entry.


sctp/src/association/mod.rs
```diff
@@ -53,7 +53,14 @@ use crate::util::*;
...
 /// MTU for inbound packet (from DTLS)
-pub(crate) const INITIAL_MTU: u32 = 1228;
+// The largest safe SCTP packet: starting from the IPv6 minimum MTU of
+// 1280 (where on-path fragmentation is unavailable), subtract
+// 40 (IPv6 header) + 8 (UDP) + 4 (TURN ChannelData) + 13 (DTLS record
+// header) + 24 (GCM cipher) = 1191, the same derivation and value as
+// libwebrtc's kMaxSafeMTUSize. There is no PMTUD, so this initial
+// value never adapts; anything larger makes delivery depend on IP
+// fragmentation surviving the path (#806).
+pub(crate) const INITIAL_MTU: u32 = 1191;
```
rainliu, line 63: I think we should keep  INITIAL_MTU: u32 = 1228.

If you need to handle small size, then use the config to change.

rainliu (changes requested):

sctp/src/association/mod.rs
phact: Okay, I will update the default back up to 1228

rainliu (approved):

PR #813, opened 2026-07-09
perf: eliminate tokio scheduler overhead on the data-channel send path (#101)

StefanoD:
## Summary

Two changes that together roughly **double** in-process data-channel throughput on the default multi-threaded tokio runtime, by removing the tokio *scheduler* overhead on the SCTP send/wake hot path. Addresses #101.

Both changes live in the `webrtc` (async wrapper) crate — the sans-io `rtc` core is unchanged. The headline change (dedicated reactor thread) is **opt-in, off by default**, so existing behaviour is preserved unless you ask for it.

## The bottleneck

For a bulk transfer, every `dc.send()` did one **blocking** `driver_event_tx.send(WriteNotify).await` on a bounded channel (a scheduler wake + a sender yield per ~1 KB message), and the single per-connection driver task was migrated across tokio worker threads on every wake (cache-cold DTLS/SCTP state + cross-worker futex wakeups).

A worker-count sweep of the flow-control example (16-core Linux, steady-state, 1 KB messages) makes the cost concrete — throughput peaks at 2 workers and *collapses* as the scheduler thrashes:

| `TOKIO_WORKER_THREADS` | throughput | ctx-switches | cpu-migrations |
|---|---|---|---|
| 1 | 141 Mbps | 333 | 12 |
| 2 | 276 Mbps | 10.9k | 284 |
| 4 | 195 Mbps | 1.26M | 14.2k |
| 8 | 177 Mbps | 1.23M | 16.1k |
| 16 (default) | 190 Mbps | 1.0M | 10.8k |

pion, measured identically, sustains **~361 Mbps** and is not sensitive to core count — Go keeps a woken goroutine core-local, and its write loop is woken by a non-blocking, coalescing poke.

## Change 1 — coalesce the driver wake (pion `awakeWriteLoop`)

Replace the per-message blocking send with an `AtomicBool` gate: only the `false→true` transition drops a single **non-blocking** `WriteNotify`; the driver clears the flag at the top of its loop and drains the core unconditionally, so a burst of N sends produces at most one wake (`PeerConnectionRef::wake_writes`).

Because the gate keeps that channel near-empty, the implicit backpressure the old per-message send provided (once the 256-deep channel filled) is gone — on a runtime without a cooperative poll budget (smol) a hot sender then starves the driver at the shared core mutex. So the wake path cooperatively yields once every 128 coalesced sends, which mimics tokio's budget and restores large drain batches. Adds a runtime-agnostic `yield_now`.

## Change 2 — opt-in dedicated reactor thread

`PeerConnectionBuilder::with_dedicated_reactor_thread(true)` runs a connection's driver on its own OS thread with a single-threaded reactor (new `Runtime::spawn_reactor`; the default impl falls back to `spawn`). The driver is pinned to one core, so the runtime never migrates it — which **eliminates the worker-count sensitivity entirely** (cpu-migrations 10k+ → ~120, and throughput becomes flat across 2/4/8/16 workers).

tokio I/O resources bind to the reactor that wraps them, so `PeerConnectionImpl::new` now binds the std sockets up front (for local addrs / SDP) but defers `wrap_udp_socket`/`wrap_tcp_listener` into the reactor future, where the event loop also runs. A `Drop` impl best-effort-sends `Close` for the dedicated variant so the thread can't leak if a connection is dropped without `close()`.

It is **off by default**: one OS thread per connection is ideal for latency/throughput-sensitive deployments with a modest number of connections, but not for e.g. an SFU with thousands. With it enabled, event-handler callbacks run on the reactor thread and must not block.

## Results

Steady-state loopback throughput, 16-core Linux, 1 KB messages, matched to pion's 1 MB flow-control window:

(percentages are relative to baseline)

| | baseline | + coalesced wake | + dedicated reactor |
|---|---|---|---|
| **tokio, default workers** | 178 Mbps | 215 (+21%) | **~300 (+69%)** |
| **smol** | 120 Mbps | 166 (+38%) | **270 (+125%)** |

At the default runtime this is **178 → ~300 Mbps (+69%)**, now independent of worker count, closing the gap to pion (~361 Mbps, measured identically) from ~2.0× to ~1.2×. The remaining difference is the two surviving app↔driver cross-thread hops per round trip; removing them would mean fusing the application and driver onto one thread, which is out of scope here.

## Testing

- New `tests/dedicated_reactor_data_channel.rs` exercises the reactor path end-to-end on both runtimes, including the UDP **and** TCP-listener socket-wrapping-on-reactor-thread paths and clean shutdown.
- All existing default-path integration tests pass (tokio and smol).
- `cargo clippy --lib` is clean on both runtimes.

## Notes

- The flow-control example's thresholds were lowered to match pion's (1 MB / 512 KB, from 100 MB / 5 MB) so throughput is directly comparable and a fast sender can't flood the SCTP send buffer. It also gained `FLOW_WARMUP_MB` / `FLOW_STOP_MB` env knobs for deterministic, bounded `perf stat` runs and a `FLOW_DEDICATED_REACTOR` toggle.
- No `rtc` submodule change is required; this builds against the current pin.

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -174,6 +174,7 @@ where
     mdns_mode: MulticastDnsMode,
     udp_addrs: Vec<A>,
     tcp_addrs: Vec<A>,
+    dedicated_reactor: bool,
 }
 
 impl<A: ToSocketAddrs> Default for PeerConnectionBuilder<A, NoopInterceptor> {
@@ -185,6 +186,7 @@ impl<A: ToSocketAddrs> Default for PeerConnectionBuilder<A, NoopInterceptor> {
             mdns_mode: MulticastDnsMode::Disabled,
             udp_addrs: vec![],
             tcp_addrs: vec![],
+            dedicated_reactor: false,
         }
     }
 }
@@ -234,6 +236,7 @@ where
             mdns_mode: self.mdns_mode,
             udp_addrs: self.udp_addrs,
             tcp_addrs: self.tcp_addrs,
+            dedicated_reactor: self.dedicated_reactor,
         }
     }
 
@@ -261,6 +264,24 @@ where
         self
     }
 
+    /// Run this peer connection's driver on its own dedicated OS thread with a
+    /// single-threaded reactor, instead of on the shared async runtime.
+    ///
+    /// This pins the driver (and thus its SCTP/DTLS/SRTP state and I/O reactor)
+    /// to one core so the runtime never migrates it across worker threads, which
+    /// markedly improves in-process data-channel throughput on multi-threaded
+    /// runtimes (issue #101). It costs one OS thread per peer connection, so it
+    /// is **off by default**: enable it for latency/throughput-sensitive
+    /// deployments with a modest number of connections, and leave it off for
+    /// large-scale servers (e.g. SFUs) with thousands of connections.
+    ///
+    /// Note: with this enabled, event-handler callbacks run on the dedicated
+    /// reactor thread, so they must not block.
+    pub fn with_dedicated_reactor_thread(mut self, enabled: bool) -> Self {
+        self.dedicated_reactor = enabled;
+        self
+    }
+
     /// Builds the [`PeerConnection`] and starts the background event loop driver.
     pub async fn build(self) -> Result<impl PeerConnection> {
         let runtime = if let Some(runtime) = self.runtime {
@@ -279,6 +300,7 @@ where
             self.mdns_mode,
             self.udp_addrs,
             self.tcp_addrs,
+            self.dedicated_reactor,
         )
         .await
     }
@@ -390,6 +412,10 @@ where
 {
     inner: Arc<PeerConnectionRef<I>>,
     driver_handle: Mutex<Option<JoinHandle>>,
+    /// Whether the driver runs on a dedicated reactor thread. When true, a
+    /// best-effort `Close` is sent on drop so the thread does not leak if the
+    /// connection is dropped without an explicit `close()`.
+    dedicated_reactor: bool,
 }
 
 pub(crate) struct PeerConnectionRef<I = NoopInterceptor>
@@ -480,30 +506,34 @@ where
         mdns_mode: MulticastDnsMode,
         udp_addrs: Vec<A>,
         tcp_addrs: Vec<A>,
+        dedicated_reactor: bool,
     ) -> Result<Self> {
-        let async_mdns_socket = if mdns_mode != MulticastDnsMode::Disabled {
-            let socket = MulticastSocket::new().into_std()?;
-            Some(runtime.wrap_udp_socket(socket)?)
+        // Bind the std sockets up front (synchronous, and needed to compute the
+        // local addresses used for ICE gathering / SDP). Wrapping them into async
+        // I/O resources is deferred so it can happen on whichever runtime actually
+        // drives the event loop: with a dedicated reactor thread, tokio I/O
+        // resources must be created on the reactor that polls them, so wrapping is
+        // done inside the reactor future (see `run_driver`) rather than here.
+        let std_mdns_socket = if mdns_mode != MulticastDnsMode::Disabled {
+            Some(MulticastSocket::new().into_std()?)
         } else {
             None
         };
 
-        let mut async_udp_sockets = HashMap::new();
+        let mut std_udp_sockets = Vec::new();
         for addr in udp_addrs {
             let socket = std::net::UdpSocket::bind(addr)?;
             socket.set_nonblocking(true)?;
             let local_addr = socket.local_addr()?;
-            let async_udp_socket = runtime.wrap_udp_socket(socket)?;
-            async_udp_sockets.insert(local_addr, async_udp_socket);
+            std_udp_sockets.push((local_addr, socket));
         }
 
-        let mut async_tcp_listeners = HashMap::new();
+        let mut std_tcp_listeners = Vec::new();
         for addr in tcp_addrs {
             let listener = std::net::TcpListener::bind(addr)?;
             listener.set_nonblocking(true)?;
             let local_addr = listener.local_addr()?;
-            let async_tcp_listener = runtime.wrap_tcp_listener(listener)?;
-            async_tcp_listeners.insert(local_addr, async_tcp_listener);
+            std_tcp_listeners.push((local_addr, listener));
         }
 
         let configuration = core.get_configuration();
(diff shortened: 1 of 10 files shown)


src/peer_connection/mod.rs
```diff
@@ -466,35 +551,88 @@ where
...
+                for (local_addr, socket) in std_udp_sockets {
+                    async_udp_sockets.insert(local_addr, driver_runtime.wrap_udp_socket(socket)?);
+                }
+                let mut async_tcp_listeners = HashMap::new();
+                for (local_addr, listener) in std_tcp_listeners {
+                    async_tcp_listeners
+                        .insert(local_addr, driver_runtime.wrap_tcp_listener(listener)?);
+                }
+
+                let mut driver = PeerConnectionDriver::new(
```
rainliu: build() now silently swallows init errors. run_driver wraps socket wrapping + PeerConnectionDriver::new + the event loop in one async block whose Err is only error!-logged (mod.rs:602). Pre-PR, wrap/init errors propagated out of build(). Now a failed wrap (e.g. exhausted fd limit) yields an apparently-healthy PeerConnection with a dead driver. 

Fix: send the init Result back over a oneshot that build() awaits (init is fast; the event loop then continues fire-and-forget). 

This is one must-fix before merge.

src/peer_connection/mod.rs
```diff
@@ -466,35 +551,88 @@ where
...
+        // A dedicated reactor thread only exits when its event loop returns, so
+        // a connection dropped without an explicit `close()` would leak the
+        // thread. Best-effort signal it to stop. (Task-based drivers detach
+        // harmlessly, so this is limited to the dedicated-reactor case to avoid
+        // changing the default lifecycle.)
+        if self.dedicated_reactor {
+            let _ = self
+                .inner
+                .driver_event_tx
+                .try_send(PeerConnectionDriverEvent::Close);
```
rainliu, line 681: Drop leak path: drop → try_send(Close) (mod.rs:627-631) can fail on a full channel; the flag-gated design makes fullness rare, but if it happens the reactor thread + runtime + sockets leak forever (the driver holds a strong Arc<PeerConnectionRef>, so there's no drop-based fallback). 

Options: driver holds Weak and exits when upgrade fails on wake; or Drop retries via a detached std-thread blocking send. Also worth noting: the new test would not catch a leaked thread — it close()s properly and process exit reaps threads; a stronger test would drop without close and assert is_finished() within a timeout.

src/runtime/tokio.rs
```diff
@@ -31,6 +31,25 @@ impl Runtime for TokioRuntime {
...
     }
 
+    fn spawn_reactor(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) -> super::JoinHandle {
+        let join = std::thread::Builder::new()
+            .name("webrtc-pc-reactor".into())
+            .spawn(move || {
+                // A single-threaded runtime: its I/O + timer drivers live on this
+                // one thread, so the driver task is never migrated. enable_all()
+                // is required for sleep()/recv_from() to work.
+                match ::tokio::runtime::Builder::new_current_thread()
```
rainliu, line 45: zero affinity code in the tree (grep affinity/sched_set — nothing). It's thread-confinement; the OS may still migrate the thread.

Offer real pinning (core_affinity) as a follow-up PR with //TODO.

src/peer_connection/mod.rs
rainliu, line 691: close() doesn't join the reactor thread.

abort() no-op + handle dropped ⇒ close() returning doesn't guarantee socket/thread release;

rainliu (commented):
Thank you very much for this PR to improve the performance.

Leave a few comments, you can take a look whether you can address them.

src/peer_connection/mod.rs
```diff
@@ -260,6 +264,24 @@ where
...
+    /// to one core so the runtime never migrates it across worker threads, which
+    /// markedly improves in-process data-channel throughput on multi-threaded
+    /// runtimes (issue #101). It costs one OS thread per peer connection, so it
+    /// is **off by default**: enable it for latency/throughput-sensitive
+    /// deployments with a modest number of connections, and leave it off for
+    /// large-scale servers (e.g. SFUs) with thousands of connections.
+    ///
+    /// Note: with this enabled, event-handler callbacks run on the dedicated
+    /// reactor thread, so they must not block.
+    pub fn with_dedicated_reactor_thread(mut self, enabled: bool) -> Self {
```
rainliu, line 286: I think we may use similar API like https://github.com/retty-io/retty/blob/master/src/executor/mod.rs

something like thread_per_core API
```
pub fn with_dedicated_reactor_thread_per_core(mut self, core_id: CoreId) -> Self {
   self.dedicated_reactor = Some(core_id);
   self
}
```

src/peer_connection/mod.rs
StefanoD: Fixed in 08f5a304. `run_driver` now reports its init `Result` over a one-shot channel that `new()` awaits, so `wrap_udp_socket` / `wrap_tcp_listener` / `PeerConnectionDriver::new` errors propagate out of `build()` again; only after init succeeds does the event loop run fire-and-forget. A failed wrap now returns `Err` from `build()` instead of yielding a healthy-looking `PeerConnection` behind a dead driver.

src/peer_connection/mod.rs
StefanoD: Fixed in 08f5a304. Added an infallible `closing: AtomicBool` on `PeerConnectionRef`, checked at the top of the driver event loop. `close()`/`Drop` `store` it (cannot fail) and then best-effort `try_send(Close)` purely as a fast wake — so a full channel can drop the wake but cannot leak the thread: the flag guarantees the loop exits, and a full channel means the driver is actively draining and re-checks the flag on its next iteration. I went with the flag rather than `Weak`-in-driver to avoid an `upgrade()` on the hot path.

Also added `tests/dedicated_reactor_drop.rs` per your suggestion: it drops a dedicated-reactor connection **without** `close()` and asserts the reactor OS thread terminates within a timeout — on Linux by watching it disappear from `/proc/self/task`, a stronger check than `is_finished()` since it confirms the real OS thread is gone (each `tests/*.rs` is its own process, so the count is not polluted by other tests).

src/runtime/tokio.rs
StefanoD: Agreed — it is thread *confinement*, not core affinity. Reworded the builder and `Runtime::spawn_reactor` docs to say so explicitly (the async runtime no longer migrates the driver across its worker pool, but the OS may still move the thread between cores), and added `TODO(#101)` markers in `tokio.rs`, `smol.rs`, and the builder for optional real pinning via `core_affinity` as a follow-up. (08f5a304)

src/peer_connection/mod.rs
StefanoD: Fixed in 08f5a304. For the dedicated-reactor case `close()` now waits (bounded: 1 ms poll, 2 s cap) for the reactor `JoinHandle::is_finished()` before returning, so the socket and OS thread are released by the time `close()` resolves. It exits promptly once it observes the `closing` flag / `Close`; the bound only guards against a wedged reactor, after which the handle is detached as a last resort. Non-dedicated drivers still `abort()` as before.

src/peer_connection/mod.rs
StefanoD: Good idea — `with_dedicated_reactor_thread_per_core(CoreId)` (backed by `dedicated_reactor: Option<CoreId>`) is the natural home for real pinning. Since actual affinity is deferred to the `core_affinity` follow-up (the `TODO(#101)`s), I kept the boolean `with_dedicated_reactor_thread(bool)` for this PR rather than introduce a `CoreId` that is not wired to anything yet — so the API and the behavior can ship together in that follow-up. Happy to add the per-core overload in this PR instead if you would prefer it here.

StefanoD:
Thanks for the thorough review, @rainliu! Pushed 08f5a304 addressing all five comments (details inline in each thread):

- **Init errors (must-fix)** — `build()` again propagates socket-wrap / driver-init errors, via a one-shot `Result` channel that the constructor awaits; the event loop then runs fire-and-forget.
- **Drop leak** — an infallible `closing: AtomicBool` checked at the top of the event loop guarantees the reactor thread exits even if the `Close` wake is dropped on a full channel; added `tests/dedicated_reactor_drop.rs`, which drops **without** `close()` and asserts the reactor OS thread terminates.
- **`close()` doesn't join** — for the dedicated case `close()` now waits (bounded) for the reactor thread to finish before returning, so its socket/thread are released by then.
- **Pinning claims** — reworded to thread *confinement* (not core affinity) with `TODO(#101)` markers for real `core_affinity` pinning as a follow-up.
- **`thread_per_core` API** — kept the boolean for now; the `CoreId` overload is the natural home for the affinity follow-up (glad to add it here if you'd rather).

While addressing these I also caught and fixed a race the `closing` flag introduced: `close()` no longer treats a failed `Close` send as an error (the flag already guarantees shutdown), so a clean shutdown can't spuriously return `Err`. Re-ran the throughput benchmark — no regression (~261 Mbps before/after, within noise).

rainliu (approved):

PR #127, opened 2021-10-28
Integrate UDP mux support for ICE

k0nserv:
Sibling PR https://github.com/webrtc-rs/ice/pull/16

PR #817, opened 2026-07-14
feat(datachannel): opt-in send back-pressure (writable / try_send) to bound peak RSS

StefanoD:
# feat(datachannel): opt-in send back-pressure (writable / try_send) to bound peak RSS

> **v3 — redesigned again per @rainliu's review, now mirroring `tokio::mpsc`.** The send-buffer
> limit is **opt-in**: the default is `usize::MAX` (unbounded), so there is **zero behaviour change**
> unless an application configures a limit. When a limit *is* set, `send()`/`send_text()` **block**
> until there is room (like `mpsc::Sender::send`), new `try_send()`/`try_send_text()` **fail fast**
> with `ErrSendBufferFull` (like `mpsc::Sender::try_send`), and new `writable()` awaits capacity.
> (Evolution: v1 blocking-only → v2 non-blocking-only → v3 both, opt-in. See the thread.)

## Dependency: rtc#127 is merged ✅ — CI no longer blocked

The sans-io counter plumbing ([rtc#127](https://github.com/webrtc-rs/rtc/pull/127)) merged into
upstream `rtc` on 2026-07-14 as `4d902e9c`. This branch's `rtc` submodule pointer **already equals
upstream `master`'s**, so the PR builds standalone. This PR carries only the webrtc-side send
back-pressure API + tests + bench.

## Summary

Adds **opt-in send back-pressure** to data channels, so an application can bound peak RSS under a
fast producer and a slow reactor/peer. The unbounded send path (found while comparing webrtc-rs
against rustrtc, pinned down by an N-pair stress test) lets `dc.send()` grow send-side memory
without limit as the transfer grows — an OOM/DoS risk under a slow reactor **or** a slow/malicious
peer advertising a tiny receive window. Configuring a per-channel limit bounds it.

**The default is unbounded (`usize::MAX`)** — no behaviour change unless you opt in. Rationale
(per review): only Chromium hard-codes a 16 MiB send-queue cap; Safari and Firefox impose none, and
the W3C flow-control model is app-driven (`bufferedAmountLowThreshold` + `onbufferedamountlow`). So
the library keeps the historical non-blocking behaviour by default and lets the application choose a
cap and a blocking-vs-fail-fast policy.

## API

All new trait methods are `async fn` with **default bodies**, so an external `impl DataChannel`
keeps compiling (`writable` → `Ok(())`, `try_send`/`try_send_text` → delegate to `send`/`send_text`).

- **`PeerConnectionBuilder::with_data_channel_send_buffer_limit(bytes)`** — per-channel send-buffer
  limit, applied connection-wide. **Default `usize::MAX` (unbounded); `0` is also unbounded.** A
  finite limit opts into back-pressure. `16 * 1024 * 1024` (Chromium's `MaxSendQueueSize`) is a
  reasonable browser-like value, well above the ~1 MiB SCTP receive window.
- **`DataChannel::send` / `send_text`** — with a limit configured, **block** until the channel's
  `outstanding_bytes` is below it, then enqueue (mirrors `mpsc::Sender::send`). With no limit (the
  default) they never block, exactly as before.
- **`DataChannel::try_send` / `try_send_text`** *(new)* — **never block**: over the limit they return
  `Err(ErrSendBufferFull)` immediately (mirrors `mpsc::Sender::try_send`). `data` is **consumed** on
  rejection — retain a clone to retry.
- **`DataChannel::writable() -> Result<()>`** *(new)* — resolves once `outstanding_bytes` is below the
  limit (immediately when unbounded); `ErrDataChannelClosed` if the channel/connection is closing.
  This is the await-for-capacity primitive `send`/`send_text` use internally; call it directly to
  pace your own sends.
- **`DataChannel::outstanding_bytes() -> Result<usize>`** — query primitive (from rtc#127). Unlike
  `bufferedAmount` (post-packetization) it also counts bytes still queued in the app→core→SCTP
  pipeline — a slightly more complete back-pressure signal.
- **Errors:** `ErrSendBufferFull` (from `try_send`/`try_send_text` only; retryable once the buffer
  drains — pace on `OnBufferedAmountLow`, don't spin). `ErrDataChannelClosed` is checked first, so
  after `close()`/`Drop` every send method — and `writable()` — fails **terminally** rather than
  hanging or spinning on the retryable error; a send already **parked** in back-pressure is woken and
  returns `ErrDataChannelClosed` whenever the driver stops for any reason — `close()`/`Drop` **or** an
  abnormal error exit (so a stalled peer + a dropped network can't hang a parked sender).

## Semantics: level-triggered, not a reserved permit

`writable()` resolves on `outstanding_bytes < limit` — it is **not** a reserved permit like
`mpsc::Sender::reserve`. For the normal single-sender-per-channel case this is exact. With multiple
concurrent senders on one channel the limit is a **soft** bound (each admitted sender may add one
in-flight message over it) — the same property as `bufferedAmount`-based flow control. A lone message
larger than the limit still goes out (`writable()` returns as soon as `outstanding` hits 0), so a big
send never deadlocks.

## A/B: peak RSS under a naive flood

A naive sender that ignores `bufferedAmount` floods the blocking `send()` over **30 pairs × 32 MB**
(unordered / no-retransmit, dedicated reactor). Same binary, only the limit differs; `poop`, 3 runs
each. Opting into a limit bounds peak RSS (and, because the pipeline no longer balloons, runs
faster):

| measurement | unbounded (default) | 16 MiB limit | 4 MiB limit |
|---|---|---|---|
| wall_time | 5.07 s ± 113 ms | 4.31 s ± 121 ms — ⚡ **−14.9%** | 3.71 s ± 1.3 ms — ⚡ **−26.8%** |
| peak_rss | 2.36 GB ± 78 MB | 1.17 GB ± 34 MB — ⚡ **−50.5%** | 576 MB ± 17 MB — ⚡ **−75.6%** |
| cpu_cycles | 122 G ± 2.5 G | 109 G ± 1.8 G — ⚡ −10.1% | 96.2 G ± 0.1 G — ⚡ −21.0% |
| instructions | 109 G ± 2.0 G | 97.5 G ± 1.1 G — ⚡ −10.8% | 85.2 G ± 0.4 G — ⚡ −22.0% |
| cache_references | 8.47 G ± 235 M | 7.65 G ± 162 M — ⚡ −9.8% | 6.76 G ± 47 M — ⚡ −20.2% |
| cache_misses | 1.25 G ± 58 M | 1.12 G ± 26 M — ⚡ −10.1% | 1.02 G ± 28 M — ⚡ −18.1% |
| branch_misses | 312 M ± 12 M | 285 M ± 9 M — −8.8% | 256 M ± 3.9 M — ⚡ −18.1% |

The bounded arms' peak RSS stays **flat** as the transfer grows, whereas the unbounded arm scales
with total bytes in flight. Reproduce with `examples/data-channels-stress-bench` (see its header).

### Happy path is unaffected (flow-control A/B)

The complementary check: a well-behaved, `bufferedAmount`-paced sender (the
[#101](https://github.com/webrtc-rs/webrtc/issues/101#issuecomment-4950542200) throughput harness —
1 KB messages, 512 KB/1 MB watermarks, ordered/reliable, dedicated reactor) is **throughput- and
resource-neutral** vs `master`, because the opt-in machinery is dormant on the default (unbounded)
path — `writable()` short-circuits before locking, and the driver skips the per-iteration wake
entirely when no limit is set. `poop`, master vs this branch, 3 runs each:

| | master | this branch | Δ |
|---|---|---|---|
| throughput, N=1 | 722 Mbps | 726 Mbps | +0.5% (noise) |
| throughput, N=10 aggregate | 3261 Mbps | 3306 Mbps | +1.4% (noise) |

Every `poop` counter is within measurement noise (no significant markers): N=1 wall `−0.8%`, peak_rss
`+0.3%`, instructions `+0.9%`; N=10 wall `+1.0%`, peak_rss `−11.2%` (σ±19%), instructions `+0.7%`.

## Tests

All green on **tokio and smol**:

- `data_channel_send_blocking.rs` — blocking `send()` with a limit: `outstanding_bytes` stays bounded
  throughout **and** every byte is delivered (reliable/ordered ⇒ drain-to-0 proves SCTP acked all).
- `data_channel_send_backpressure.rs` — non-blocking `try_send`/`try_send_text` cap: bounded +
  reject-engaged + conservation (incl. `max_retransmits: 0` abandonment).
- `data_channel_close_wakes_send.rs` — a `send()` **parked** in back-pressure is woken by `close()`
  and returns `ErrDataChannelClosed` (liveness; without the `closing` check + `close`/`Drop` wake it
  hangs).
- `data_channel_send_after_close.rs` — `send`/`send_text`/`try_send`/`try_send_text`/`writable` all
  fail **terminally** after `close()`.
- `data_channel_send_unbounded.rs` — `0`/default = unbounded escape hatch; reliable end-to-end
  delivery with the gate disabled.
- unit tests — the trait default bodies (`outstanding_bytes`→0, `writable`→`Ok`, `try_send*`→delegate).

## Design evolution / addressing the review (@rainliu)

- **v1** parked `send()` in a loop → *"strongly against the parking loop."*
- **v2** made `send()` non-blocking (`ErrSendBufferFull`), 16 MiB default.
- **v3 (this)** mirrors `tokio::mpsc`, per the latest review:
  1. **Default `usize::MAX`** — zero behaviour change unless opted in (Safari/Firefox have no cap).
  2. **`async fn try_send`/`try_send_text`** (not sync) — the counter lives behind the async core
     mutex, so a sync `fn try_send` would need `try_lock()` and spuriously fail under contention;
     `async fn` is consistent with the rest of the object-safe trait and has no such failure mode.
  3. **Blocking `send`/`send_text` when a limit is set** — the complexity lives in `writable()`, and
     `send()` is just `self.writable().await?` then enqueue.
  4. **"Silent drop" clarified** — today's `send()` never drops *payload*; only the internal unit
     `WriteNotify` wake is droppable at channel capacity, and that is harmless (the driver runs
     `poll_writes()` unconditionally each loop). No data-loss bug; the new APIs are ergonomics.

### DIFF
--- a/src/data_channel/mod.rs
+++ b/src/data_channel/mod.rs
@@ -477,3 +477,83 @@ where
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::runtime::block_on;
+
+    /// A minimal external `DataChannel` implementation that intentionally does **not**
+    /// override [`DataChannel::outstanding_bytes`], so it exercises the trait's default body.
+    ///
+    /// This locks in the compatibility contract the send back-pressure work relies on: adding
+    /// `outstanding_bytes()` to the public trait must not break a downstream `impl DataChannel`
+    /// that predates it, and the default must report `0` — i.e. "no back-pressure signal" — so
+    /// those channels keep compiling and behaving. Every other method is unreachable in this
+    /// test and left `unimplemented!()`.
+    struct DefaultDataChannel;
+
+    #[async_trait::async_trait]
+    impl DataChannel for DefaultDataChannel {
+        async fn label(&self) -> Result<String> {
+            unimplemented!()
+        }
+        async fn ordered(&self) -> Result<bool> {
+            unimplemented!()
+        }
+        async fn max_packet_life_time(&self) -> Result<Option<u16>> {
+            unimplemented!()
+        }
+        async fn max_retransmits(&self) -> Result<Option<u16>> {
+            unimplemented!()
+        }
+        async fn protocol(&self) -> Result<String> {
+            unimplemented!()
+        }
+        async fn negotiated(&self) -> Result<bool> {
+            unimplemented!()
+        }
+        fn id(&self) -> RTCDataChannelId {
+            unimplemented!()
+        }
+        async fn ready_state(&self) -> Result<RTCDataChannelState> {
+            unimplemented!()
+        }
+        async fn buffered_amount_high_threshold(&self) -> Result<u32> {
+            unimplemented!()
+        }
+        async fn set_buffered_amount_high_threshold(&self, _threshold: u32) -> Result<()> {
+            unimplemented!()
+        }
+        async fn buffered_amount_low_threshold(&self) -> Result<u32> {
+            unimplemented!()
+        }
+        async fn set_buffered_amount_low_threshold(&self, _threshold: u32) -> Result<()> {
+            unimplemented!()
+        }
+        // outstanding_bytes(): deliberately NOT overridden — this is the whole point of the test.
+        async fn send(&self, _data: BytesMut) -> Result<()> {
+            unimplemented!()
+        }
+        async fn send_text(&self, _text: &str) -> Result<()> {
+            unimplemented!()
+        }
+        async fn poll(&self) -> Option<DataChannelEvent> {
+            unimplemented!()
+        }
+        async fn close(&self) -> Result<()> {
+            unimplemented!()
+        }
+    }
+
+    #[test]
+    fn outstanding_bytes_trait_default_is_zero() {
+        let dc = DefaultDataChannel;
+        let n =
+            block_on(dc.outstanding_bytes()).expect("default outstanding_bytes() must return Ok");
+        assert_eq!(
+            n, 0,
+            "the DataChannel::outstanding_bytes default must report 0 outstanding bytes"
+        );
+    }
+}
--- a/src/data_channel/mod.rs
+++ b/src/data_channel/mod.rs
@@ -50,6 +50,7 @@ use futures::FutureExt;
 use rtc::interceptor::{Interceptor, NoopInterceptor};
 use rtc::shared::error::{Error, Result};
 use std::sync::Arc;
+use std::sync::atomic::Ordering;
 use std::time::Duration;
 
 pub use rtc::data_channel::{
@@ -90,13 +91,16 @@ fn send_limits() -> (usize, usize) {
                 None => default,
             }
         };
-        (
-            resolve("WEBRTC_SEND_HIGH_WATER_BYTES", DATA_CHANNEL_SEND_HIGH_WATER),
-            resolve(
-                "WEBRTC_SEND_HARD_CEILING_BYTES",
-                DATA_CHANNEL_SEND_HARD_CEILING,
-            ),
-        )
+        let high_water = resolve("WEBRTC_SEND_HIGH_WATER_BYTES", DATA_CHANNEL_SEND_HIGH_WATER);
+        let hard_ceiling = resolve(
+            "WEBRTC_SEND_HARD_CEILING_BYTES",
+            DATA_CHANNEL_SEND_HARD_CEILING,
+        );
+        // The hard ceiling must never sit below the high-water mark: the admit check
+        // (`outstanding + len <= high_water`) runs before the reject check, so a ceiling
+        // below the mark would admit sends it was meant to reject, silently defeating the
+        // backstop. Clamp so `hard_ceiling >= high_water` regardless of misconfiguration.
+        (high_water, hard_ceiling.max(high_water))
     })
 }
 
@@ -135,7 +139,12 @@ pub trait DataChannel: Send + Sync + 'static {
     /// that SCTP has not yet released (acknowledged or abandoned) — the true
     /// outstanding send-side memory, including bytes still queued in the send
     /// pipeline (unlike `bufferedAmount`, which counts only post-packetization).
-    async fn outstanding_bytes(&self) -> Result<usize>;
+    ///
+    /// Defaults to `0` so external implementors of this trait keep compiling; the
+    /// built-in channel overrides it with the real counter that drives send back-pressure.
+    async fn outstanding_bytes(&self) -> Result<usize> {
+        Ok(0)
+    }
     /// Sends raw binary data on this data channel.
     async fn send(&self, data: BytesMut) -> Result<()>;
     /// Sends text data on this data channel.
@@ -425,6 +434,14 @@ where
         let (high_water, hard_ceiling) = send_limits();
         let len = data.len();
         loop {
+            // Bail if the connection is closing/dropped. Once the driver stops (on the
+            // `Close` event) it no longer drains `outstanding_bytes` or wakes us, and the
+            // channel is not removed from the core map on close, so the re-check below
+            // stays `Some`. Without this a sender parked at the high-water mark would spin
+            // on the 50 ms liveness timer forever; `close()`/`Drop` wake us immediately.
+            if self.inner.closing.load(Ordering::Acquire) {
+                return Err(Error::ErrDataChannelClosed);
+            }
             {
                 let mut peer_connection = self.inner.core.lock().await;
                 let mut dc = peer_connection
@@ -470,6 +487,10 @@ where
         let (high_water, hard_ceiling) = send_limits();
         let len = text.len();
         loop {
+            // See `send`: bail if closing so a parked sender can't hang past teardown.
+            if self.inner.closing.load(Ordering::Acquire) {
+                return Err(Error::ErrDataChannelClosed);
+            }
             {
                 let mut peer_connection = self.inner.core.lock().await;
                 let mut dc = peer_connection
--- a/src/data_channel/mod.rs
+++ b/src/data_channel/mod.rs
@@ -46,63 +46,14 @@
 use crate::peer_connection::PeerConnectionRef;
 use crate::runtime::{Mutex, Receiver};
 use bytes::BytesMut;
-use futures::FutureExt;
 use rtc::interceptor::{Interceptor, NoopInterceptor};
 use rtc::shared::error::{Error, Result};
 use std::sync::Arc;
 use std::sync::atomic::Ordering;
-use std::time::Duration;
 
 pub use rtc::data_channel::{
     RTCDataChannelId, RTCDataChannelInit, RTCDataChannelMessage, RTCDataChannelState,
 };
-
-/// Default high-water mark for synchronous data-channel send back-pressure:
-/// [`DataChannel::send`] blocks once this many bytes are outstanding (handed to
-/// `send`/`send_text` but not yet acknowledged or abandoned by SCTP) on the channel.
-/// Chosen well above the ~1 MiB SCTP receive window so steady-state throughput is
-/// unaffected, yet far below the tens-of-MiB/connection an unbounded send path can
-/// otherwise accumulate under a slow reactor or a slow/malicious peer advertising a
-/// tiny receive window. Override with `WEBRTC_SEND_HIGH_WATER_BYTES` (`0` = unbounded).
-const DATA_CHANNEL_SEND_HIGH_WATER: usize = 4 * 1024 * 1024;
-
-/// Default hard ceiling backstop: above this, [`DataChannel::send`] returns
-/// [`Error::ErrSendBufferFull`] instead of blocking, so a counter that ever drifts
-/// upward (a bug, or a torn-down channel) can never hang a sender indefinitely.
-/// Override with `WEBRTC_SEND_HARD_CEILING_BYTES` (`0` = unbounded).
-const DATA_CHANNEL_SEND_HARD_CEILING: usize = 16 * 1024 * 1024;
-
-/// Resolved `(high_water, hard_ceiling)` send back-pressure limits, read once from the
-/// environment (falling back to the defaults above). A value of `0` in either env var
-/// means "unbounded" and is mapped to `usize::MAX`, disabling that gate — this is the
-/// escape hatch for the send-backpressure A/B benchmark and for callers that do their
-/// own flow control. `WEBRTC_SEND_HIGH_WATER_BYTES` / `WEBRTC_SEND_HARD_CEILING_BYTES`.
-fn send_limits() -> (usize, usize) {
-    use std::sync::OnceLock;
-    static LIMITS: OnceLock<(usize, usize)> = OnceLock::new();
-    *LIMITS.get_or_init(|| {
-        let resolve = |key: &str, default: usize| -> usize {
-            match std::env::var(key)
-                .ok()
-                .and_then(|s| s.parse::<usize>().ok())
-            {
-                Some(0) => usize::MAX, // 0 = unbounded
-                Some(n) => n,
-                None => default,
-            }
-        };
-        let high_water = resolve("WEBRTC_SEND_HIGH_WATER_BYTES", DATA_CHANNEL_SEND_HIGH_WATER);
-        let hard_ceiling = resolve(
-            "WEBRTC_SEND_HARD_CEILING_BYTES",
-            DATA_CHANNEL_SEND_HARD_CEILING,
-        );
-        // The hard ceiling must never sit below the high-water mark: the admit check
-        // (`outstanding + len <= high_water`) runs before the reject check, so a ceiling
-        // below the mark would admit sends it was meant to reject, silently defeating the
-        // backstop. Clamp so `hard_ceiling >= high_water` regardless of misconfiguration.
-        (high_water, hard_ceiling.max(high_water))
-    })
-}
 
 /// Object-safe trait exposing all public DataChannel operations.
 ///
@@ -237,22 +188,6 @@ where
             id,
             inner,
             evt_rx: Mutex::new(evt_rx),
-        }
-    }
-
-    /// Park until the channel's send buffer has room under the high-water mark, called
-    /// only on the slow path when a `send` finds the buffer already over the mark.
-    ///
-    /// Holds no lock while waiting: the driver applies SCTP buffer releases
-    /// (acknowledged or abandoned bytes) to the per-channel `outstanding_bytes`
-    /// counter and then wakes `data_channel_backpressure`, so a blocked sender
-    /// re-checks and proceeds as soon as the peer acknowledges data. The 50 ms
-    /// timeout is a lost-wakeup / liveness backstop (and, for a peer that never
-    /// acks, the sender correctly stays blocked rather than buffering unboundedly).
-    async fn await_send_capacity(&self) {
-        futures::select! {
-            _ = self.inner.data_channel_backpressure.notified().fuse() => {}
-            _ = crate::runtime::sleep(Duration::from_millis(50)).fuse() => {}
         }
     }
 }
@@ -411,7 +346,27 @@ where
         Ok(())
     }
 
-    /// Send binary data
+    /// Send binary data.
+    ///
+    /// Non-blocking, like the browser `RTCDataChannel.send()`: it queues the message and
+    /// returns; it never waits for the peer to acknowledge. To bound send-side memory it
+    /// applies a **non-blocking** back-pressure gate — see *Errors*.
+    ///
+    /// # Errors
+    ///
+    /// - [`Error::ErrSendBufferFull`] if this message would push the channel's outstanding
+    ///   send bytes (handed to `send`/`send_text` but not yet acknowledged or abandoned by
+    ///   SCTP) past the configured send-buffer limit
+    ///   ([`PeerConnectionBuilder::with_data_channel_send_buffer_limit`], default 16 MiB;
+    ///   set `0` to disable). This mirrors the browser's "throw when the send queue is
+    ///   full". It is **retryable**, but only *after* the buffer drains — an application
+    ///   should pace on the [`OnBufferedAmountLow`](DataChannelEvent::OnBufferedAmountLow)
+    ///   event (see `examples/data-channels-flow-control`) or poll
+    ///   [`outstanding_bytes`](Self::outstanding_bytes) rather than spin. Note `data` is
+    ///   **consumed** on rejection, so retain a clone if you intend to retry.
+    /// - [`Error::ErrDataChannelClosed`] if the channel/connection is closed or closing.
+    ///
+    /// [`PeerConnectionBuilder::with_data_channel_send_buffer_limit`]: crate::peer_connection::PeerConnectionBuilder::with_data_channel_send_buffer_limit
     ///
     /// # Example
     ///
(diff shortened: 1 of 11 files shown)


rainliu (changes requested):
could your rebase to master?

src/data_channel/mod.rs
```diff
@@ -46,14 +46,64 @@
...
+/// upward (a bug, or a torn-down channel) can never hang a sender indefinitely.
+/// Override with `WEBRTC_SEND_HARD_CEILING_BYTES` (`0` = unbounded).
+const DATA_CHANNEL_SEND_HARD_CEILING: usize = 16 * 1024 * 1024;
+
+/// Resolved `(high_water, hard_ceiling)` send back-pressure limits, read once from the
+/// environment (falling back to the defaults above). A value of `0` in either env var
+/// means "unbounded" and is mapped to `usize::MAX`, disabling that gate — this is the
+/// escape hatch for the send-backpressure A/B benchmark and for callers that do their
+/// own flow control. `WEBRTC_SEND_HIGH_WATER_BYTES` / `WEBRTC_SEND_HARD_CEILING_BYTES`.
+fn send_limits() -> (usize, usize) {
```
rainliu: 1. DATA_CHANNEL_SEND_HIGH_WATER and DATA_CHANNEL_SEND_HARD_CEILING are hardcode, it is not flexible enough.  std::env::var should be avoided, instead, make it configurable with default value as above.
2.  send_limits() is kind of utility function, which is not suitable to put it in this file.

src/data_channel/mod.rs
```diff
@@ -342,12 +426,41 @@ where
...
-                .ok_or(Error::ErrDataChannelClosed)?
-                .send(data)?;
+        // Synchronous send back-pressure, folded into the send's own core-lock so the
+        // fast path takes the lock exactly ONCE: check the outstanding-bytes counter
+        // and enqueue atomically. A sender that outruns the drain (slow reactor or
+        // slow/malicious peer) blocks here rather than growing send-side memory without
+        // bound. Only the over-high-water slow path releases the lock and parks.
+        let (high_water, hard_ceiling) = send_limits();
+        let len = data.len();
+        loop {
```
rainliu: it looks like it will introduce possible long-wait due to this loop for send and send_text, which may hurt user-experience.

public APIs like send and sent_text should be executed as fast as possible, although they are async function. Otherwise, user application may be blocked by this loop for very long time, if not handled well in user's application level.

I am strongly against to introduce so loop in these two APIs.

rainliu: I agree "The unbounded send path (found while comparing webrtc-rs against rustrtc,
and pinned down by an N-pair stress test) lets dc.send() grow send-side memory to ~7× the payload
and climb without limit as the transfer grows — an OOM/DoS risk under a slow reactor "

but, it is user's responsibility to avoid such OOM/DoS risk by querying new added API outstanding_bytes before calling send or send_text.

therefore, those new added `synchronous send` logic should be moved to example code, instead of putting inside send and send_txt functions.

StefanoD:
Thanks for the review @rainliu, and for merging rtc#127. I've **rebased onto `master`** — since #127 is now in upstream `rtc`, the submodule points at the merged `4d902e9c` on `webrtc-rs/rtc` (no fork remote needed anymore).

On the two design objections I went to the primary sources, and **I think you're right that `send()`/`send_text()` must not block**. Before I rewrite, here's the evidence and the direction I'd propose — I'd like to confirm it matches your intent first.

### 1. `send()` must not park — agreed; it's what the platform does

The W3C algorithm *queues and returns*; a full buffer is a **synchronous error, not a wait**:

> Queue `data` for transmission on `channel`'s underlying data transport. If queuing data is not possible because not enough buffer space is available, throw an `OperationError`. […] NOTE: The actual transmission of `data` occurs in parallel.

— W3C WebRTC, [§6.2 `send()` algorithm](https://www.w3.org/TR/webrtc/#datachannel-send)

Both reference implementations do exactly this:

- **Chromium** throws synchronously (`OperationError`, *"RTCDataChannel send queue is full"*) once `bufferedAmount + length` would exceed a 16 MiB cap, otherwise enqueues and returns — it never blocks: [`rtc_data_channel.cc#L465-L475`](https://github.com/chromium/chromium/blob/096d4e6ae84f0e2426f4b1dcc5bf205acaaba10b/third_party/blink/renderer/modules/peerconnection/rtc_data_channel.cc#L465-L475), cap = [`MaxSendQueueSize() = 16 MiB`](https://webrtc.googlesource.com/src/+/04964db74ddc63236c30b7e3e38e78f43cdd3536/api/data_channel_interface.cc#48).
- **pion**'s `Send()` is just `ensureOpen()` + `WriteDataChannel()` and returns immediately ([`datachannel.go#L446`](https://github.com/pion/webrtc/blob/v4.2.16/datachannel.go#L446)); its flow-control [example README](https://github.com/pion/webrtc/blob/v4.2.16/examples/data-channels-flow-control/README.md) states: *"The methods return immediately … it is an application's responsibility to control the buffered amount in order not to indefinitely grow the buffer size to eventually exhaust the memory."*

So I'll **remove the parking loop** (which also deletes the `Notify` / 50 ms-timer / close-wake machinery it required).

### 2. Env-var config — agreed, wrong knob

I'll drop `WEBRTC_SEND_*` and the global `OnceLock`, and make any limit a builder / `RTCDataChannelInit` option with a sensible default.

### 3. "App's responsibility / move to example" — mostly agreed

The spec's flow-control model is app-driven ([`bufferedAmountLowThreshold` + `bufferedamountlow`](https://www.w3.org/TR/webrtc/#dom-rtcdatachannel-bufferedamountlowthreshold)), and this crate already ships it (`buffered_amount_low_threshold` + `OnBufferedAmountLow`). So I'll add an **example** that paces sending on `OnBufferedAmountLow`, and keep `outstanding_bytes()` as the query primitive. (It's incremented at app-send time, before packetization, so it also reflects bytes still queued in the app→core→SCTP pipeline — a slightly more complete signal than the post-packetization `bufferedAmount` for this purpose.)

One nuance I'd value your call on: the platform doesn't push *all* protection to the app — browsers keep a **non-blocking** in-library backstop (the 16 MiB → `OperationError` above). So my inclination is to keep a **non-blocking hard cap that returns `ErrSendBufferFull` immediately** (never parks; mirrors `OperationError`), configurable per-channel with a sane default. That preserves default-safety against unbounded send-buffer growth without violating the "send is fast" contract.

**Would you prefer** (a) keep that non-blocking `ErrSendBufferFull` cap as an in-library backstop, or (b) drop it entirely and leave *all* bounding to the app + example? Either works for me — I just want to match your design intent before rewriting. Thanks!

StefanoD:
@rainliu — I've implemented the redesign we discussed and pushed it (tip `23ee776c`). The [PR description](https://github.com/webrtc-rs/webrtc/pull/817) is fully rewritten to match; the short version of how your three points are addressed:

1. **Hard-coded consts + `std::env::var`** → removed. The limit is now a builder option — `PeerConnectionBuilder::with_data_channel_send_buffer_limit(bytes)` — defaulting to 16 MiB (Chromium's `MaxSendQueueSize`); `0` disables it.
2. **`send_limits()` in the wrong file** → deleted. The redesign reads a single limit from the connection config, so the helper is gone entirely.
3. **"Strongly against the parking loop"** → the loop (and the `Notify` / 50 ms-timer / close-wake machinery it required) is gone. `send()`/`send_text()` now return synchronously; over the limit they fail fast with `ErrSendBufferFull` and never wait — matching the W3C `OperationError`, Chromium, and pion (sources in the description).

On the one open question from my previous comment: I went with **(a)** — a **non-blocking** in-library cap as a last-resort backstop (the same shape browsers keep: 16 MiB → synchronous `OperationError`), with `outstanding_bytes()` + the `OnBufferedAmountLow`-paced `data-channels-flow-control` example as the primary, app-driven mechanism. If you'd rather go with **(b)** — drop the in-library cap entirely and leave all bounding to the app + example — I'm happy to switch; it's a small change. Just let me know which you prefer.

For context, the naive-flood A/B (a sender that ignores `bufferedAmount`): peak RSS **−50%** at the 16 MiB default, **−78%** at 4 MiB, with RSS staying flat as the transfer grows instead of climbing toward OOM. A well-behaved, `bufferedAmount`-paced sender never approaches the cap, so the happy path is unaffected.

I've resolved the config / `send_limits()` thread; I left the "move to example / app's responsibility" thread open, since that's the (a)-vs-(b) call above. Thanks again for the review!

src/peer_connection/mod.rs
```diff
@@ -161,6 +161,17 @@ pub trait PeerConnectionEventHandler: Send + Sync + 'static {
...
+/// Default per-channel data-channel send-buffer limit: [`DataChannel::send`] and
+/// [`DataChannel::send_text`] reject with [`Error::ErrSendBufferFull`] once more than
+/// this many bytes are outstanding (handed to send but not yet acknowledged or
+/// abandoned by SCTP) on a single channel. `16 MiB` matches the browser
+/// `RTCDataChannel` send-queue cap (`webrtc::DataChannelInterface::MaxSendQueueSize`),
+/// well above the ~1 MiB SCTP receive window so a well-behaved sender that paces on
+/// `OnBufferedAmountLow` never hits it, yet a bound against the unbounded send-side
+/// growth a naive or throttled sender would otherwise accumulate. Override (or disable
+/// with `0`) via [`PeerConnectionBuilder::with_data_channel_send_buffer_limit`].
+pub const DEFAULT_DATA_CHANNEL_SEND_BUFFER_LIMIT: usize = 16 * 1024 * 1024;
```
rainliu: do we really need it public?

src/data_channel/mod.rs
```diff
@@ -85,6 +86,16 @@ pub trait DataChannel: Send + Sync + 'static {
...
     /// Sets the buffered amount low threshold in bytes.
     async fn set_buffered_amount_low_threshold(&self, threshold: u32) -> Result<()>;
+    /// Returns the bytes handed to [`send`](Self::send)/[`send_text`](Self::send_text)
+    /// that SCTP has not yet released (acknowledged or abandoned) — the true
+    /// outstanding send-side memory, including bytes still queued in the send
+    /// pipeline (unlike `bufferedAmount`, which counts only post-packetization).
+    ///
+    /// Defaults to `0` so external implementors of this trait keep compiling; the
+    /// built-in channel overrides it with the real counter that drives send back-pressure.
+    async fn outstanding_bytes(&self) -> Result<usize> {
```
rainliu, line 98: providing public API outstanding_bytes may not be conventional, instead, just provide conventional 
`async fn writable(&self) -> Result<()>` is more ergonomic for Rust networking library application.
like https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.writable

rainliu (changes requested):

src/data_channel/mod.rs
rainliu: it may even better to provide a pair of API

```
async fn writable(&self) -> Result<()>;
fn try_send(&self, data: BytesMut)-> Result<()>;
fn try_send_text(&self, text: &str) -> Result<()>;
```

and still keep existing 
```
async fn send(&self, data: BytesMut) -> Result<()>;
async fn send_text(&self, text: &str) -> Result<()>
```

rainliu:
Let me summarize my review comments here:

1. Before this PR,  the follow existing APIs are non-block async send fn,
```
async fn send(&self, data: BytesMut) -> Result<()>;
async fn send_text(&self, text: &str) -> Result<()>;
```
when internal driver_event_tx channel reach the PEER_CONNECTION_DRIVER_EVENT_CHANNEL_CAPACITY, the data will be silently drop.

2. In this PR, we should introduce new APIs:
```
async fn writable(&self) -> Result<()>;
fn try_send(&self, data: BytesMut)-> Result<()>;
fn try_send_text(&self, text: &str) -> Result<()>;
```
but maintain the same async send and send_text APIs' default behavior (non-blocking) by initializing data_channel_send_buffer_limit to usize::MAX, unless application uses with_data_channel_send_buffer_limit to change this limit.

if data_channel_send_buffer_limit is not usize::MAX, the async send and send_text API will be blocked to await writable if sending buffer is larger than the limit.

StefanoD:
Thanks @rainliu — clear spec. Before I rewrite I want to check four things against the code, because a couple change the design.

**1. The "silent drop at `PEER_CONNECTION_DRIVER_EVENT_CHANNEL_CAPACITY`" — I don't think `send`/`send_text` can drop *payload* today.**

The payload never travels over `driver_event_tx`. In `send()` the bytes are committed **synchronously into the SCTP core buffer under the core lock** (`dc.send(data)?`, `src/data_channel/mod.rs:413`) *before* the droppable notify. The only thing sent on the bounded (cap 256) `driver_event_tx` is the **unit `WriteNotify` wake**, via `let _ = try_send(WriteNotify)` in `wake_writes()` (`src/peer_connection/mod.rs:552-554`). So at capacity we drop a *wake*, not data.

And that dropped wake is harmless: the driver calls `poll_writes()` **unconditionally at the top of every loop iteration** (`src/peer_connection/driver.rs:203-204`), regardless of whether a `WriteNotify` arrived. A full `driver_event_tx` also *means* 256 events are already queued that will wake the driver and flush. Worst case is the flush waits for the next timer/socket wake (a latency nuance), not lost bytes — and `wake_writes` only sends the notify on the `false→true write_pending` transition anyway. If `dc.send()` itself can't enqueue, it returns `Err` (propagated by `?`), never silent.

So from tracing the path I believe `send`/`send_text` are lossless today. If you have a repro of an actual payload drop I'll dig in — but if not, we don't need the new APIs to fix a data-loss bug (they're still worth adding as ergonomics; see below).

**2. `try_send` / `try_send_text` as synchronous `fn` — the counter lives behind the async core mutex, so a sync signature is problematic.**

`outstanding_bytes` is a plain `usize` inside `RTCPeerConnection` behind the **async** `core: runtime::Mutex` (`rtc/rtc/src/data_channel/internal.rs:27`); reading it *and* enqueuing both require `core.lock().await`. There's no atomic/lock-free mirror. A synchronous `fn try_send` would have to `try_lock()` the runtime mutex — which the driver and every other channel op also hold — so it can **spuriously fail under contention even when the buffer has room**, conflating "lock busy, retry" with "buffer full" (`ErrSendBufferFull`). (Unlike `tokio::mpsc::try_send`, which is sync only because that channel has a lock-free fast path; our counter doesn't.)

Two options: **(a)** `async fn try_send`/`try_send_text` — consistent with the rest of the object-safe async trait (every method but `id()` is already async), no spurious failures; or **(b)** keep them sync but add a distinct `ErrWouldBlock` (lock-busy ≠ buffer-full) and expose `try_lock` on the runtime `Mutex` wrappers. I lean **(a)**. Preference?

**3. Defaulting `data_channel_send_buffer_limit` to `usize::MAX` removes the default safety this PR adds.**

With `usize::MAX` default, an app that never calls `with_data_channel_send_buffer_limit` is back to unbounded send-buffer growth — the exact OOM this PR bounds. Note browsers don't default to unbounded either: Chromium keeps a **16 MiB** in-library cap regardless of app behavior, so `usize::MAX` default is strictly less safe than both this PR and the platform. Fine to ship if strict backward-compat (zero behavior change unless opted in) is the priority — I just want that to be a conscious call vs. keeping the 16 MiB default. Your call?

**4. Blocking `send`/`send_text` when a limit is set — confirming the reversal.**

Last round we agreed `send`/`send_text` "must not park" (W3C queues-or-throws; Chromium throws `OperationError`; pion returns immediately), and I removed the `Notify` / 50 ms-timer / close-wake machinery. The new spec re-introduces blocking (await `writable()`) whenever a limit is configured, bringing that machinery back. Implementable, and it mirrors `tokio::mpsc` (`send().await` awaits capacity, `try_send` fails fast) — I just want to confirm this supersedes the "must not park" conclusion: blocking is acceptable **because** it's opt-in behind a configured limit, with `try_send` as the non-blocking escape.

Confirm 2–4 (and correct me on 1) and I'll implement it in one pass.

rainliu:
Thanks for clarifying. Here are my comments:

1. looks like I misunderstood it. Then, it is good.
2.  (a) async fn try_send/try_send_text is better
3. let's keep zero behavior change unless opted in. in other words, use usize::MAX as default. The reason is only Chromium has such hardcode 16 MiB limitation, but other browsers, like, Safari and Firefox, seem not have such cap.
4. yes, let's mirror tokio::mpsc (send().await awaits capacity, try_send fails fast): blocking is acceptable because it's opt-in behind a configured limit, with try_send as the non-blocking escape.

StefanoD:
Great — thanks, that's all confirmed. Here's what I'll implement, one pass:

1. **No silent drop** → no change needed for correctness; the three new methods are added as flow-control ergonomics, not a data-loss fix.
2. **`async fn try_send` / `async fn try_send_text`** — fail fast with `ErrSendBufferFull` when over the limit, never await capacity.
3. **`data_channel_send_buffer_limit` defaults to `usize::MAX`** (zero behavior change unless opted in).
4. **`send`/`send_text` block on `writable()`** only when a limit is configured; otherwise (the `usize::MAX` default) they stay exactly as today. Mirrors `tokio::mpsc`: `send().await` awaits capacity, `try_send` fails fast.

Two implementation notes so there are no surprises in the diff:

- **All three new methods get default bodies** on the object-safe `DataChannel` trait (like `outstanding_bytes` does), so this stays non-breaking for external `impl DataChannel`: `writable` defaults to `Ok(())`, `try_send`/`try_send_text` default to delegating to `send`/`send_text`.
- **Blocking `send()` re-checks under the lock in a loop**, since `writable()` is advisory (it means "outstanding is below the limit," not "room for this exact message"); it also admits a lone oversized message when the buffer is empty so a single large send can't deadlock. This reinstates the `Notify` + close/`Drop` wake + lost-wakeup backstop, active only when a limit is set.

One consequence of (3) I'll fold in: with `usize::MAX` as the default the gate is a no-op unless opted in, so I'll **re-title and rewrite the PR** away from "bound peak RSS by default" toward opt-in flow-control primitives (the −50%/−78% naive-flood RSS numbers become "when a limit is configured"). I'll also add the `OnBufferedAmountLow`-paced `data-channels-flow-control` example we discussed, with `outstanding_bytes()` as the query primitive.

Will push shortly.

rainliu:
> Blocking send() re-checks under the lock in a loop, since writable() is advisory (it means "outstanding is below the limit," not "room for this exact message"); it also admits a lone oversized message when the buffer is empty so a single large send can't deadlock. This reinstates the Notify + close/Drop wake + lost-wakeup backstop, active only when a limit is set.

Why not moving such complexity into new writable()? in current send(), just call writable().await?

StefanoD:
Good call — that's cleaner, adopting it. `writable()` will own all of it (the `Notify` wait + 50 ms lost-wakeup backstop + `closing` check + a `usize::MAX` fast-path that returns immediately), and `send()`/`send_text()` become just `self.writable().await?` then lock-and-enqueue. `try_send()`/`try_send_text()` stay the non-blocking peek.

One semantic note so it's explicit: `writable()` is level-triggered (resolves when `outstanding < limit`), not a reserved permit like `mpsc::Sender::reserve`. So `send()` doesn't re-check after it returns — for a single sender that's exact; with multiple concurrent senders on one channel the limit is soft (overshoot up to ~one in-flight message per sender), same as `bufferedAmount`-based flow control. The oversized-lone-message case is handled for free since `writable()` returns as soon as `outstanding` hits 0.

StefanoD:
Pushed the redesign (tip `f394ed1a`); PR title/description rewritten to match.

**What landed**, exactly as we settled:

- `async fn writable()` / `try_send()` / `try_send_text()`, all with **default bodies** so external
  `impl DataChannel` keep compiling (`writable` → `Ok(())`, `try_send*` → delegate to `send*`).
- **Default `usize::MAX`** (unbounded) — zero behaviour change unless a limit is configured.
- With a limit set, `send()`/`send_text()` **block** via `writable().await?` then enqueue
  (`mpsc::Sender::send`); `try_send*()` **fail fast** with `ErrSendBufferFull` (`mpsc::Sender::try_send`).
  All the block-until-room complexity lives in `writable()`, per your suggestion.

**Independently reviewed before pushing**; it turned up one real liveness gap I've fixed:

- A blocking `send()` parked at the limit could hang if the driver **error-exited** (fatal SCTP on a
  timer tick, network drop) — `closing` was only set by `close()`/`Drop`. Now the driver sets
  `closing` + wakes parked senders whenever its event loop returns **for any reason**, so a parked
  send always terminates with `ErrDataChannelClosed`.
- Also guarded the driver's per-loop wake behind "a limit is configured", so the default unbounded
  path stays completely lock-free on the hot loop.

**A/B (this box):**
- Naive flood (30×32 MB, no `bufferedAmount` pacing) — peak RSS **−51.8%** @ 16 MiB, **−76.8%** @ 4 MiB.
- Flow-control happy path (`bufferedAmount`-paced, the #101 harness) — **throughput-neutral** vs
  `master` (N=1 ~726 vs ~722 Mbps, N=10 agg ~3306 vs ~3261), all `poop` counters within noise.

Tests (blocking-send bounded+delivered, `try_send` fail-fast, close-wakes-parked-send liveness,
terminal-after-close for all five methods, unbounded escape hatch, trait-default coverage) are green
on **tokio and smol**; `clippy`/`fmt` clean. Ready for another look whenever you have a moment.

src/data_channel/mod.rs
```diff
@@ -85,10 +88,62 @@ pub trait DataChannel: Send + Sync + 'static {
...
```
rainliu, line 98: Why trait need this? Can’t implement it as private fn?

src/data_channel/mod.rs
```diff
@@ -179,6 +234,23 @@ where
...
+    /// fires. Called only on the slow path, when [`writable`](DataChannel::writable) finds
+    /// the buffer at/over the limit.
+    ///
+    /// Holds no lock while waiting: the driver applies SCTP buffer releases (acknowledged
+    /// or abandoned bytes) to the per-channel `outstanding_bytes` counter and then wakes
+    /// `data_channel_backpressure`, so a blocked sender re-checks and proceeds as soon as
+    /// the peer acknowledges data. The 50 ms timeout is a lost-wakeup / liveness backstop
+    /// — and, for a peer that never acks, the sender correctly stays blocked (bounding
+    /// send-side memory) while still re-checking `closing` so `close()`/`Drop` release it.
+    async fn await_send_capacity(&self) {
```
rainliu, line 248: Same question for this trait fn?

src/data_channel/mod.rs
StefanoD: It's on the trait as a public *observability* query, not an internal helper — the counterpart to `bufferedAmount`, but counting the full app→core→SCTP send pipeline (pre-packetization) that `bufferedAmount` misses. It's added with a default body (`Ok(0)`) so it's non-breaking for external `impl DataChannel`, and it's **not** used internally: `writable()`/`try_send()` read the counter directly under the core lock, so nothing in the gate depends on this trait method. Its consumers are apps doing their own pacing and the back-pressure tests (which assert the send queue stays bounded — there's no other public signal that sees the pre-packetization queue).

That said, if you'd rather keep the public trait minimal, I'm happy to drop it — `writable()`/`try_send()` + `bufferedAmount`/`OnBufferedAmountLow` cover flow control without it, and I'd move the tests onto a `#[cfg(test)]` accessor. Keep it as a public query, or remove it?

src/data_channel/mod.rs
StefanoD: This one isn't on the trait — `await_send_capacity` is a private inherent method on `DataChannelImpl` (in the `impl<I> DataChannelImpl<I>` block, not `impl DataChannel for …`), called only by `writable()`. So it's already private, not part of the public API; it just sits right above the trait impl in the file. Happy to inline it into `writable()` if that reads cleaner.

rainliu (approved):

PR #824, opened 2026-07-25
fix(data): raise PollDataChannel default read buffer to 16 KiB

### LINKED ISSUE #823: PollDataChannel default read buffer rejects valid messages larger than 8 KiB
Thank you for maintaining webrtc-rs.

We encountered a `PollDataChannel` default read-buffer issue while integrating a libp2p WebRTC transport.

## Observed behavior

`PollDataChannel` uses an 8 KiB temporary read buffer by default. When the DataChannel receives a complete, valid SCTP message larger than that buffer, the read fails with a short-buffer error, for example:

```text
Short buffer (size: 8192) to be filled
```

For callers using the default configuration, this can subsequently close the DataChannel.

WebRTC DataChannels deliver complete SCTP messages. In our case, libp2p's WebRTC transport may produce framed messages close to 16 KiB, so the 8 KiB default cannot receive a valid message from that caller.

Increasing the default temporary buffer to 16 KiB resolved the short-buffer failure in our integration.

## Request for guidance

Would one of the following approaches be preferred upstream?

1. Increase the default temporary read-buffer capacity.
2. Keep the default, but make the buffer-capacity configuration clearer and more reliable for callers.
3. Use a growable temporary buffer for a received message instead of rejecting it solely because of a fixed default capacity.

If this is considered a bug, I can submit a small PR with an integration-style regression test that sends and reads a message larger than 8 KiB using the default `PollDataChannel`.

yexiyue:
## Summary

- raise `PollDataChannel`'s default temporary read buffer from 8 KiB to 16 KiB;
- add a regression test that locks the default at 16 KiB; and
- correct the public setter documentation to match the new default.

Closes #823.

## Motivation

`PollDataChannel` adapts message-oriented SCTP DataChannels to `AsyncRead`. Each read starts by allocating its internal temporary buffer. SCTP delivers one complete user message, so a message larger than that fixed default produces `ErrShortBuffer` before `PollDataChannel` can split it across the caller's `ReadBuf`.

The prior 8 KiB default therefore rejected valid messages from callers that do not override `set_read_buf_capacity`. libp2p's WebRTC transport can use frames up to 16 KiB, which made this observable as a transport reset in browser-to-browser transfers.

16 KiB is a conservative default that accepts those frames while keeping the existing opt-in setter for applications needing a different bound. This does not change SCTP's negotiated maximum message size or the behaviour of explicitly configured channels.

## Validation

- `cargo check -p webrtc-data --tests`
- `cargo test -p webrtc-data data_channel::data_channel_test::poll_data_channel_default_read_buffer_supports_16_kib_messages -- --exact`

## Attribution

Prepared with assistance from Codex 5.6; the reproducer, design decision, and validation were reviewed by the contributor.

### DIFF
--- a/data/src/data_channel/data_channel_test.rs
+++ b/data/src/data_channel/data_channel_test.rs
@@ -5,6 +5,11 @@ use util::conn::conn_bridge::*;
 use util::conn::*;
 
 use super::*;
+
+#[test]
+fn poll_data_channel_default_read_buffer_supports_16_kib_messages() {
+    assert_eq!(DEFAULT_READ_BUF_SIZE, 16 * 1024);
+}
 use crate::error::Result;
 
 async fn bridge_process_at_least_one(br: &Arc<Bridge>) {
--- a/data/src/data_channel/mod.rs
+++ b/data/src/data_channel/mod.rs
@@ -470,7 +470,7 @@ impl PollDataChannel {
         self.data_channel.buffered_amount_low_threshold()
     }
 
-    /// Set the capacity of the temporary read buffer (default: 8192).
+    /// Set the capacity of the temporary read buffer (default: 16384).
     pub fn set_read_buf_capacity(&mut self, capacity: usize) {
         self.read_buf_cap = capacity
     }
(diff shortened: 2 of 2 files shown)


yexiyue:
Closing this draft: the related libp2p transport now enforces an 8 KiB connection-level DataChannel message limit, which fits the existing default buffer. The original 16 KiB assumption no longer applies, so increasing webrtc-rs's global default is unnecessary.

Issue #827, opened 2026-07-28
No API for the remote DTLS certificate: 0.17's get_remote_certificate has no 0.20 equivalent


yexiyue:
## Summary

In 0.17, reading the peer's DTLS certificate was one call:

```rust
// webrtc 0.17, src/dtls_transport/mod.rs
/// get_remote_certificate returns the certificate chain in use by the remote side
/// returns an empty list prior to selection of the remote certificate
pub async fn get_remote_certificate(&self) -> Bytes
```

0.20 has no equivalent. There is no `dtls_transport` module in the public surface at all, and
nothing on `PeerConnection` exposes the remote certificate.

The only route left is the statistics report, which works but is a fair amount of ceremony for
what used to be an accessor.

## Getting there today

```rust
// `StatsSelector` is `use`d, not `pub use`d, by `webrtc::peer_connection`, and `webrtc` does not
// re-export `rtc` (only `rtc::shared::error` via `webrtc::error`). So calling `get_stats` at all
// requires taking a direct dependency on `rtc` just to name the argument type.
use rtc::statistics::StatsSelector;
use rtc::statistics::report::RTCStatsReportEntry;

let report = pc.get_stats(std::time::Instant::now(), StatsSelector::None).await;

// Two Certificate entries come back -- local and remote -- and nothing on the entry itself says
// which is which. You have to find the Transport entry first and match its id.
let mut remote_id = String::new();
for entry in report.iter() {
    if let RTCStatsReportEntry::Transport(t) = entry {
        remote_id = t.remote_certificate_id.clone();
    }
}

let remote_cert = report.iter().find_map(|entry| match entry {
    RTCStatsReportEntry::Certificate(c) if c.stats.id == remote_id => Some(c),
    _ => None,
});
```

Verified on master, loopback connection:

```
transport.remote_certificate_id = "remote-certificate-99:7f:31"
cert id="2e:a2:99:26:db:81:..."                matches_remote=false fp=2e:a2:99:26:db:81 der_len=720
cert id="remote-certificate-99:7f:31"          matches_remote=true  fp=99:7f:31:8c:ae:47 der_len=720
```

So the data is all there. What is missing is a way to ask for it.

## Why it matters

Certificate-pinned protocols need this. libp2p's [WebRTC-Direct][spec] runs a Noise handshake
over the data channel whose prologue binds **both** DTLS fingerprints:

```
libp2p-webrtc-noise:<client-fingerprint><server-fingerprint>
```

[...]

PR #828, opened 2026-07-28
fix(peer_connection): export the stats types `get_stats` names

yexiyue:
Addresses #827.

> **Revised after review.** The original version added
> `PeerConnection::remote_certificate_fingerprint`. @rainliu pointed out that it is not a W3C API
> and that `PeerConnection` should stay compact and compliant — a fair call, so the method is gone.
> What remains is the part that makes the W3C-shaped route actually usable from outside the crate.

## Problem

0.17 had a one-line accessor:

```rust
// webrtc 0.17, src/dtls_transport/mod.rs
pub async fn get_remote_certificate(&self) -> Bytes
```

0.20 has no equivalent. There is no `dtls_transport` module in the public surface, and nothing on
`PeerConnection` exposes the peer's certificate. The statistics report is the intended route — but
`get_stats` is a public trait method whose signature names types the crate does not export:

```rust
async fn get_stats(&self, now: Instant, selector: StatsSelector) -> RTCStatsReport;
```

`StatsSelector` and `RTCStatsReport` are `use`d rather than `pub use`d by `webrtc::peer_connection`,
and `webrtc` does not re-export `rtc` (only `rtc::shared::error`, via `webrtc::error`). So a caller
cannot name the argument or the return type without adding `rtc` to their own `Cargo.toml` — which
means restating a version this crate already pins and keeping the two in step by hand, while `rtc`
is still on release candidates.

And once you have the report, the lookup is indirect: it carries one certificate entry per side and
neither says which is which. You have to find the transport entry, read `local_certificate_id` /
`remote_certificate_id`, and match those back against the certificate entries — a traversal that is
not discoverable from the types.

## Change

- **`pub use` for `StatsSelector`, `RTCStatsReport` and `RTCStatsReportEntry`**, so `get_stats` is
  callable and its result readable. These types already appear in this crate's public signatures, so
  re-exporting them doesn't widen the API — it makes the existing one usable. The neighbouring
  `pub use rtc::interceptor::…` and `pub use rtc::peer_connection::…` lines in the same file follow
  the same principle.
- **Docs on `get_stats`** describing the certificate lookup and pointing at the test as a worked
  example.
- **`tests/remote_certificate_fingerprint.rs`**, which doubles as that example.

## Why it matters

libp2p's [WebRTC-Direct][spec] runs a Noise handshake whose prologue binds **both** DTLS
fingerprints:

```
libp2p-webrtc-noise:<client-fingerprint><server-fingerprint>
```

Both ends must derive the same prologue or the handshake fails. Each side knows its own fingerprint,
and the dialer learns the listener's from the multiaddr it dialed (`/webrtc-direct/certhash/...`).
That leaves exactly one value nobody can supply out of band: the listener needs the dialer's, and it
exists only in the certificate presented during DTLS. The 0.17-based transport in rust-libp2p does
precisely this:

```rust
async fn get_remote_fingerprint(conn: &RTCPeerConnection) -> Fingerprint {
    let cert_bytes = conn.sctp().transport().get_remote_certificate().await;
    Fingerprint::from_certificate(&cert_bytes)
}
```

On 0.20 the application performs the stats traversal itself. That is the intended shape — it just
needs the types to be nameable.

[spec]: https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md

## Test

`tests/remote_certificate_fingerprint.rs` connects two peers over loopback and cross-checks both
directions: what one peer reports as *remote* must equal what the other reports as *local*. It also
asserts the two certificates differ, so the cross-check cannot pass by coincidence, and that nothing
is returned before the handshake.

That shape is deliberate. Reading the local fingerprint by mistake is a silent failure — a pinning
check would compare a fingerprint against itself and accept every peer. Swapping
`remote_certificate_id` for `local_certificate_id` in the helper makes the test fail with the two
fingerprints printed side by side; a simple "is it non-empty" assertion would not have caught it.

`cargo fmt --check`, `cargo test` and `cargo clippy` are clean locally.

## Disclosure

This change was developed with AI assistance (Claude Code). I hit the gap building a WebRTC-Direct
libp2p transport on 0.20, worked around it with the stats traversal above, and opened #827 before
writing this. I have read every line of the diff and can explain it in review.

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -81,8 +81,8 @@ use rtc::rtp_transceiver::rtp_sender::RtpCodecKind;
 use rtc::rtp_transceiver::{RTCRtpTransceiverId, RTCRtpTransceiverInit};
 use rtc::sansio::Protocol;
 use rtc::shared::error::{Error, Result};
-use rtc::statistics::StatsSelector;
-use rtc::statistics::report::RTCStatsReport;
+pub use rtc::statistics::StatsSelector;
+pub use rtc::statistics::report::{RTCStatsReport, RTCStatsReportEntry};
 
 use crate::media_stream::track_local::TrackLocalEvent;
 use crate::media_stream::track_local::static_rtp::TrackLocalStaticRTP;
--- a/tests/remote_certificate_fingerprint.rs
+++ b/tests/remote_certificate_fingerprint.rs
@@ -1,0 +1,210 @@
+//! Worked example: reading the peers' DTLS certificate fingerprints out of `get_stats`.
+//!
+//! 0.17 had `RTCDtlsTransport::get_remote_certificate`; on 0.20 the statistics report is the
+//! route. Certificate-pinning protocols need the *remote* fingerprint — libp2p's WebRTC-Direct,
+//! for one, binds both peers' fingerprints into its Noise handshake prologue, and a listener
+//! can only learn the dialer's from the certificate the peer actually presented.
+//!
+//! The lookup is a two-step because the report identifies certificates indirectly: it carries a
+//! `Certificate` entry per side and neither says which side it belongs to — only the `Transport`
+//! entry's `local_certificate_id` / `remote_certificate_id` tell them apart. A cross-check pins
+//! that down: what one peer sees as *remote* must equal what the other reports as its *local*
+//! certificate, in both directions.
+//!
+//! Getting this backwards is not a loud failure. A pinning protocol would compare a fingerprint
+//! against itself and happily accept every peer.
+
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use anyhow::Result;
+use webrtc::peer_connection::{
+    PeerConnection, PeerConnectionBuilder, PeerConnectionEventHandler, RTCStatsReportEntry,
+    StatsSelector,
+};
+use webrtc::peer_connection::{RTCIceGatheringState, RTCPeerConnectionState};
+use webrtc::runtime::{Sender, block_on, channel, default_runtime, timeout};
+
+/// Which end of the connection a certificate belongs to.
+#[derive(Clone, Copy)]
+enum Side {
+    Local,
+    Remote,
+}
+
+/// The SHA-256 fingerprint of one side's DTLS certificate, colon-separated hex — the same shape
+/// as an SDP `a=fingerprint:` value ([RFC 4572 §5]). `None` before the handshake has completed.
+///
+/// [RFC 4572 §5]: https://www.rfc-editor.org/rfc/rfc4572#section-5
+async fn certificate_fingerprint(
+    pc: &dyn PeerConnection,
+    now: Instant,
+    side: Side,
+) -> Option<String> {
+    let report = pc.get_stats(now, StatsSelector::None).await;
+
+    let id = report.iter().find_map(|entry| match entry {
+        RTCStatsReportEntry::Transport(transport) => {
+            let id = match side {
+                Side::Local => &transport.local_certificate_id,
+                Side::Remote => &transport.remote_certificate_id,
+            };
+            (!id.is_empty()).then(|| id.clone())
+        }
+        _ => None,
+    })?;
+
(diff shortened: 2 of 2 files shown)


rainliu (changes requested):
remote_certificate_fingerprint is not defined in W3C API. we need to make PeerConnection's API more compact and compliance to W3C APIs

If application needs it, just use get_stats get it by itself, like this function does.

In application, just 
```
use rtc::statistics::StatsSelector;
use rtc::statistics::report::{RTCStatsReport, RTCStatsReportEntry};
```
since those struct or types are public in rtc crate.

yexiyue:
Thank you for the review — and for the clear reasoning behind it. You're right, and I've dropped `remote_certificate_fingerprint`. Keeping `PeerConnection` aligned with the W3C surface is the correct default and this method doesn't meet that bar; a convenience helper isn't worth widening a public trait.

The lookup now lives in `tests/remote_certificate_fingerprint.rs` as a worked example, and `get_stats`'s doc comment points at it — so the two-step (transport entry → `remote_certificate_id` → certificate entry) is at least discoverable by someone who goes looking for the peer's certificate and finds `get_stats` first.

One piece I did keep, and I'd like to make the case for it separately since it isn't the same question as API compactness: the `pub use` of `StatsSelector`, `RTCStatsReport` and `RTCStatsReportEntry`.

`get_stats` is a public trait method whose signature already names the first two, so as things stand a caller cannot name its argument or return type from outside the crate. Reaching them through their own `rtc` dependency does work, as you say — but it means restating a version this crate already pins and keeping the two in step by hand, which is a coupling worth avoiding while `rtc` is still on release candidates. Re-exporting types that already appear in this crate's public signatures doesn't widen the API surface; it just makes the existing one callable. This file already does it a few lines below, with `pub use rtc::interceptor::…` and `pub use rtc::peer_connection::…`, so it also seemed like the established pattern here.

That said, it's entirely your call — if you'd rather callers depend on `rtc` directly, say the word and I'll drop those three lines too and leave only the docs and the test.

Thanks again for the time on this, and for merging #825.

src/peer_connection/mod.rs
```diff
@@ -522,6 +522,14 @@ pub trait PeerConnection: Send + Sync + 'static {
         init: Option<RTCRtpTransceiverInit>,
     ) -> Result<Arc<dyn RtpTransceiver>>;
     /// Get a snapshot of accumulated statistics.
+    ///
```
rainliu: revert this comments

tests/remote_certificate_fingerprint.rs
```diff
@@ -0,0 +1,210 @@
+//! Reading the peers' DTLS certificate fingerprints out of the statistics report.
```
rainliu: just keep this new integration test to show an example how to extract peer DTLS  certificate fingerprint

src/peer_connection/mod.rs
yexiyue: Both done — the `get_stats` doc comment is reverted to its original single line, and the integration test stays as the worked example. The PR is now just that test plus the three `pub use` lines.

On those three lines: I've left them in for now, but there's a concrete consequence I should put in front of you rather than decide myself. Dropping them doesn't only affect downstream users — **this crate's own integration test stops compiling.** Integration tests under `tests/` see the crate exactly as an external caller does, and `rtc` is not in `[dev-dependencies]`. Reverting the re-exports locally to check:

```
error[E0603]: enum `RTCStatsReportEntry` is private
  --> tests/remote_certificate_fingerprint.rs:22:72
error[E0603]: enum `StatsSelector` is private
  --> tests/remote_certificate_fingerprint.rs:23:5
```

So keeping the test and dropping the re-exports would mean adding `rtc` to `[dev-dependencies]` as well — a change to `Cargo.toml` that you didn't ask for, which is why I haven't made it.

If you'd still rather callers reach the types through `rtc` directly, just say so and I'll do exactly that: drop the three lines and add the `[dev-dependencies]` entry so the test keeps building. Happy either way — it's your call on the API surface.

Thanks again for the review.

rainliu (approved):

PR #833, opened 2026-07-30
clean up stale data channel event senders on drop

dimitri-lenkov:
Hey sorry about #829; I made the PR, but when I opened the draft, someone had completely modified the section of the code I was working on so I quickly closed it till I had more time to revise my work with the latest version of webrtc.

I encountered this issue when migrating from v0.17 to v0.20. The sans-io driver fires `on_data_channel` for both locally-created and remote-opened channels; my application opens 5 data channels per connection, and during teardown some DataChannel Impl instances are dropped without a clean `OnClose` driver event, leaving zombie senders in the map.

As for the "draft" status:

I followed the `webrtc` crate's convention of placing tests inline at the bottom of the module, since that's what I observed here (as opposed to the `rtc` crate's external test files). Happy to adjust if there's a preferred structure.

## Summary

Implements W3C WebRTC Sec.6.2.4 _Closing procedure_, step 3: "Remove `channel` from `connection.[[DataChannels]]`."

> <https://www.w3.org/TR/webrtc/#closing-procedure>

The `webrtc` crate mirrors `[[DataChannels]]` with an internal `data_channel_events_tx` map that routes driver events (OnOpen, OnMessage, OnClose, …) to each channel's `DataChannelImpl`.

When a `DataChannelImpl` was dropped without a clean `OnClose` driver event (e.g. during teardown), its event sender remained in the map. Dead senders accumulated, and if a new channel later reused the same id, the driver's `OnOpen` handler saw `Entry::Occupied` and silently discarded all events for that channel.

## Changes

- `src/runtime/mod.rs`, `tokio.rs`, `smol.rs`: add `try_lock()` to `AsyncMutex` and `is_closed()` to `AsyncSender`.
- `src/data_channel/mod.rs`: add `Drop for DataChannelImpl` that best-effort removes the event sender via `try_lock()`. The `try_lock()` is intentionally non-blocking: a contended mutex skips cleanup rather than blocking in `Drop`. The `OnOpen` closed-sender replacement is the fallback that covers the race.
- `src/peer_connection/driver.rs`: extract the sender-insertion logic into `insert_data_channel_event_sender` (a `pub(crate)` function for unit testability), which replaces an `Occupied` sender when its receiver is closed. The driver `OnOpen` handler uses the helper instead of the inline match.

## Tests

- `data_channel::tests::drop_removes_event_sender`: creates a sender, inserts it, drops the `DataChannelImpl`, and asserts the map entry is removed.
- `driver::tests::insert_data_channel_event_sender_replaces_closed_sender`: populates the map with a closed sender, calls the helper with a fresh sender, and asserts `true` (the sender was inserted and the channel should be announced).
- `driver::tests::insert_data_channel_event_sender_skips_live_sender`: same with a live sender; asserts `false` (no regression for the alive path).

No integration test is added because the bug (id reuse after stale sender) cannot be forced through the public WebRTC API: in-band channel IDs are partitioned by DTLS role parity, and negotiated channels do not emit cross-wire `OnOpen` events. The unit tests cover both the `Drop` cleanup and the `OnOpen` replacement paths directly.

## Verification

```bash
cargo test --lib
cargo clippy -- -D warnings
```

### DIFF
--- a/src/data_channel/mod.rs
+++ b/src/data_channel/mod.rs
@@ -44,7 +44,7 @@
 //! ```
 
 use crate::peer_connection::PeerConnectionRef;
-use crate::runtime::{Mutex, Receiver};
+use crate::runtime::{AsyncMutex as _, Mutex, Receiver};
 use bytes::BytesMut;
 use futures::FutureExt;
 use rtc::interceptor::{Interceptor, NoopInterceptor};
@@ -249,6 +249,17 @@ where
         futures::select! {
             _ = self.inner.data_channel_backpressure.notified().fuse() => {}
             _ = crate::runtime::sleep(Duration::from_millis(50)).fuse() => {}
+        }
+    }
+}
+
+impl<I> Drop for DataChannelImpl<I>
+where
+    I: Interceptor,
+{
+    fn drop(&mut self) {
+        if let Some(mut data_channels) = self.inner.data_channel_events_tx.try_lock() {
+            data_channels.remove(&self.id);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


rainliu:
Please fix fmt/clippy

rainliu (approved):

PR #850, opened 2026-08-04
feat(runtime): expose the GRO buffer-sizing and recv-error helpers

yexiyue:
# feat(runtime): expose the GRO buffer-sizing and recv-error helpers

## Problem

`examples/custom-runtime` promises that writing a `Runtime` takes "No `#[cfg]` edits, no
fork, and no changes to library internals", and its README points anyone who wants a
production-grade one straight at the built-ins:

> A production runtime would enable UDP GSO/GRO the way the built-in ones do — construct a
> `webrtc::runtime::UdpSocketState` at wrap time, report its `max_gso_segments()` /
> `gro_segments()`, pass the whole `Transmit` to its `send`, and adapt its scatter/gather
> `recv` to the single-datagram shape `poll_recv` returns (see `src/runtime/tokio.rs` or
> `smol.rs`).

Following that advice runs into two things the built-ins rely on but downstream cannot
reach:

| helper | needed for |
|---|---|
| `gro_recv_buf_len` | sizing a receive buffer for a socket that may coalesce datagrams via GRO |
| `is_retryable_socket_recv_error` | telling "one peer went away" apart from "this socket is dead" |

Both are `pub(crate)`. On current `master`, with a throwaway probe in `tests/` (not added
by this PR):

```
error[E0432]: unresolved imports `webrtc::runtime::GRO_RECV_SEGMENT_LEN`,
`webrtc::runtime::MAX_GRO_SEGMENTS`, `webrtc::runtime::UDP_RECV_BUF_LEN`,
`webrtc::runtime::gro_recv_buf_len`, `webrtc::runtime::is_retryable_socket_recv_error`
 --> tests/downstream_visibility.rs:4:5
```

So the only option is to reimplement them from reading the source, and both are easy to
get subtly wrong.

**Buffer sizing.** `max_gro_segments()` documents that it is "Used to size each receive
buffer", but not the formula. The correct one needs `GRO_RECV_SEGMENT_LEN` (a coalesced
segment is at most one wire MTU — *not* the caller's max datagram size) and the
`MAX_GRO_SEGMENTS` clamp (`max_gro_segments()` comes from an arbitrary socket impl and is
used here as an allocation multiplier). Nothing catches a wrong segment length: too large
only wastes memory, and too small makes the kernel silently drop the tail datagrams, which
presents as unexplained packet loss.

**Error classification.** Omitting one variant means a receive loop tears down a healthy
listener. `ConnectionRefused` is the one to watch: it is how an ICMP port-unreachable can
surface on Linux, so dropping it lets any peer that goes away close a listening socket,
with no way back short of rebinding.

I hit both building a WebRTC-Direct libp2p transport that muxes one UDP port across many
`PeerConnection`s (a custom `Runtime` whose `wrap_udp_socket` hands each connection a
branch of a shared socket). I reimplemented both helpers and got both wrong: the buffer by
~5.5x (used my own max-datagram size as the segment length, and no clamp), and the error
set by omitting `Interrupted`, `WouldBlock` and `ConnectionRefused`.

## Change

- `gro_recv_buf_len`, `is_retryable_socket_recv_error`, and the three constants the former
  is defined in terms of (`GRO_RECV_SEGMENT_LEN`, `MAX_GRO_SEGMENTS`, `UDP_RECV_BUF_LEN`)
  become `pub`, re-exported from `runtime` next to the socket traits they serve.
- `is_retryable_socket_recv_error` gains a doc comment — it had none, and `missing_docs` is
  warned on, so it needs one to be public.
- A test pins the retryable set and its complement.

Purely additive, so it is a minor-version change; no behaviour change, and no existing call
site moves. The GSO-side constants (`MAX_GSO_SEGMENTS`, `MAX_GSO_BATCH_BYTES`,
`MIN_GSO_RUN`) deliberately stay private — those are the driver's send-batching policy,
not part of the socket contract.

## On placement

I left the definitions in `peer_connection::transports` and only added the re-export, to
keep this diff to the visibility change.

There is an argument for moving them into `runtime` outright: `is_retryable_socket_recv_error`
serves both the UDP and the TCP receive path, and all three socket traits live in
`runtime`. I kept that out of this PR so a move would not be mixed in with an API change —
happy to do it that way instead if you prefer.

## Testing

`cargo build`, `cargo test`, `cargo clippy --all-targets`, `cargo fmt --check` and
`cargo doc --no-deps` all pass (macOS). clippy reports no new warnings in `src/`; the
pre-existing ones are all in `examples/` and `tests/`. `cargo doc` is clean, so the new doc
comment satisfies `missing_docs` and its intra-doc links resolve.

Reachability was verified with the throwaway probe quoted above: it fails to compile on
`master` with that error and passes on this branch.

## Disclosure

This change was developed with AI assistance (Claude Code). I ran into the gap building a
WebRTC-Direct libp2p transport on 0.20, and I have read every line of the diff and can
explain it in review.

### DIFF
--- a/src/peer_connection/transports/mod.rs
+++ b/src/peer_connection/transports/mod.rs
@@ -7,11 +7,11 @@ pub(crate) mod tcp_transport;
 pub(crate) mod turn_relayer;
 
 /// Plain single-datagram UDP receive buffer size (no GRO coalescing).
-pub(crate) const UDP_RECV_BUF_LEN: usize = 2000;
+pub const UDP_RECV_BUF_LEN: usize = 2000;
 
 /// Upper bound on the number of datagrams the kernel may coalesce into one UDP GRO
 /// receive (`UDP_SEGMENT`/GRO cap is 64 per buffer).
-pub(crate) const MAX_GRO_SEGMENTS: usize = 64;
+pub const MAX_GRO_SEGMENTS: usize = 64;
 
 /// Upper bound on datagrams coalesced into one UDP GSO send. The kernel caps
 /// `UDP_SEGMENT` at 64 segments per `sendmsg`; a socket may report fewer.
@@ -43,7 +43,7 @@ pub(crate) const MIN_GSO_RUN: usize = 16;
 /// datagrams well under this (DTLS/SCTP MTU ~1200); the 1500 headroom covers a peer
 /// sending up to standard-MTU-sized datagrams. Jumbo-frame paths (MTU > 1500) are not
 /// supported for GRO and would truncate.
-pub(crate) const GRO_RECV_SEGMENT_LEN: usize = 1500;
+pub const GRO_RECV_SEGMENT_LEN: usize = 1500;
 
 /// Size a UDP receive buffer for a socket that may coalesce `max_gro` datagrams via
 /// GRO. Falls back to the plain single-datagram size when GRO is unavailable.
@@ -54,7 +54,7 @@ pub(crate) const GRO_RECV_SEGMENT_LEN: usize = 1500;
 /// [`GRO_RECV_SEGMENT_LEN`]); the buffers are zero-initialized so pages stay unmapped
 /// until actually written. Measured net effect is still an RSS *reduction* under load
 /// because batching cuts per-packet allocator churn far more than the buffers cost.
-pub(crate) fn gro_recv_buf_len(max_gro: usize) -> usize {
+pub fn gro_recv_buf_len(max_gro: usize) -> usize {
     if max_gro > 1 {
         max_gro.min(MAX_GRO_SEGMENTS) * GRO_RECV_SEGMENT_LEN
     } else {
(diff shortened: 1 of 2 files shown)


rainliu:
please submit this PR to branch `v0.20.x`, since there is no breaking change.

After merged into branch `v0.20.x`, I will merge it into branch `master` for future version too.

rainliu (changes requested):
After reviewed this PR, I don't think we need to expose those as public. Both helpers are consumed by the **driver**, not by an `AsyncUdpSocket` implementor.

So they're driver policy rather than part of the socket contract, and that's what makes
me reluctant with 1.0 close. Making them `pub` freezes our internal allocation strategy.

On your actual use case — a custom `Runtime` demultiplexing one UDP socket across many`PeerConnection`s — you're effectively acting as a driver, not just supplying a socket, so your own receive loop does own the buffer sizing. The MTU bound is the thing you needed, and `max_gso_segments` / `max_gro_segments` now state it explicitly, including the case where a runtime allocates buffers to its own rule.

I just updated README.md and inline-doc for fn poll_rec() and fn max_gro_segments() to make clear.

yexiyue:
Thanks @rainliu — retargeted onto `v0.20.x` in #853. Same commit, cherry-picked; `cargo check --all-targets` and the added tests pass on that branch.

Closing this one in favour of #853.

PR #853, opened 2026-08-04
feat(runtime): expose the GRO buffer-sizing and recv-error helpers

yexiyue:
Retarget of #850 onto `v0.20.x`, per @rainliu's request there — no breaking change, so it belongs on the patch line.

Same commit, cherry-picked onto `v0.20.x`; the patch applies unchanged (only the surrounding context in `src/runtime/mod.rs` differs between the branches).

---

`AsyncUdpSocket` is a public trait, but the two pieces of knowledge needed to drive one the way this crate's own driver does are `pub(crate)`:

- **`gro_recv_buf_len`** — how large a receive buffer must be for a socket that may coalesce datagrams via GRO. `max_gro_segments()` documents that it is "used to size each receive buffer" but not the formula, which needs both `GRO_RECV_SEGMENT_LEN` and the `MAX_GRO_SEGMENTS` clamp. Undersizing it makes the kernel silently drop the tail datagrams.
- **`is_retryable_socket_recv_error`** — which receive errors are one peer's problem rather than a dead socket. Omitting a single variant turns a routine ICMP reply into a permanently closed listener.

Both stay where they are defined and are re-exported from `runtime`, next to the socket traits they belong to. Adds a doc comment to the error classifier (it had none) and a test pinning both the retryable set and its complement.

Neither has a feedback loop for downstream implementors: an undersized buffer drops data silently, and a missing error variant kills the listener permanently. We got both wrong when we copied them into our own `udp_mux`, which is what motivated this.

Verified on `v0.20.x`: `cargo check --all-targets` clean, and the added tests pass.

### DIFF
--- a/src/peer_connection/transports/mod.rs
+++ b/src/peer_connection/transports/mod.rs
@@ -7,11 +7,11 @@ pub(crate) mod tcp_transport;
 pub(crate) mod turn_relayer;
 
 /// Plain single-datagram UDP receive buffer size (no GRO coalescing).
-pub(crate) const UDP_RECV_BUF_LEN: usize = 2000;
+pub const UDP_RECV_BUF_LEN: usize = 2000;
 
 /// Upper bound on the number of datagrams the kernel may coalesce into one UDP GRO
 /// receive (`UDP_SEGMENT`/GRO cap is 64 per buffer).
-pub(crate) const MAX_GRO_SEGMENTS: usize = 64;
+pub const MAX_GRO_SEGMENTS: usize = 64;
 
 /// Upper bound on datagrams coalesced into one UDP GSO send. The kernel caps
 /// `UDP_SEGMENT` at 64 segments per `sendmsg`; a socket may report fewer.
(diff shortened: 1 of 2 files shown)


yexiyue:
That's a fair line to draw, and the updated docs cover what I actually needed — thanks for turning it around so quickly.

You're right that I'm acting as a driver rather than just supplying a socket, so owning the buffer sizing is my job. What I was missing wasn't the symbols, it was the two rules behind them, and `ef8ba660` states both: the MTU bound on a coalesced segment (with the explicit note that a runtime allocating to its own rule carries the same bound), and the transient set in `poll_recv`'s `# Errors`.

Those are enough to implement against, so I've dropped the patch pins, gone back to `0.20.0` from crates.io, and kept the two helpers in my own mux — written against the documented rules rather than copied from the source, with tests pinning both the retryable set and the segment-vs-datagram sizing. Documenting the contract turned out to be the better fix: it doesn't spend any of your 1.0 compatibility budget, and it's what downstream can actually rely on.

Closing this one. Thanks for the review.

PR #863, opened 2026-08-09
fix(driver): prevent hot loop on non-advancing timeout

### LINKED ISSUE #862: PeerConnection driver can hot-loop on a non-advancing expired timeout
## Summary

`PeerConnectionDriver` can enter an unbounded immediate-timeout loop when the earliest Sans-I/O deadline is already expired and does not advance after `handle_timeout(now)`.

This was observed on `webrtc` v0.20.0 after a relay-backed peer lost network connectivity without a graceful close. ICE transitioned to `Failed`, TURN transactions continued timing out, and the peer connection stopped responding to close/shutdown work. The same vulnerable loop is still present on current `master` (`0ea042e7aeb44b967f9edac42bd917fb30af796e`).

## Affected code

The driver currently does this:

```rust
let timeout = self.poll_timeout().await;
let now = self.inner.runtime.now();
let delay_from_now = timeout.checked_duration_since(now).unwrap_or_default();

if delay_from_now.is_zero() {
    self.handle_timeout(now).await?;
    continue;
}
```

Source: https://github.com/webrtc-rs/webrtc/blob/0ea042e7aeb44b967f9edac42bd917fb30af796e/src/peer_connection/driver.rs#L375-L386

An expired deadline is normal in an async driver: the executor can wake after the requested instant, and some protocols intentionally request an immediate callback. The problem is the unconditional `continue`. It assumes one `handle_timeout` call always moves the minimum deadline into the future.

If any of the core, STUN, or TURN timeout sources returns the same expired deadline again, the driver never reaches `select!`. It can therefore monopolize its executor and starve:

- peer-connection driver events, including `Close`;
- incoming UDP/TCP packets that could change protocol state;
- normal shutdown and cleanup.

## Reproduction

The production sequence was:

1. Establish a TURN-backed peer connection.
2. Remove network connectivity without closing the peer connection.
3. Wait for ICE to transition through `Disconnected` to `Failed`.
4. Observe recurring `TURN transaction timed out` messages.
5. Request peer close / process shutdown.
6. The stale peer does not complete cleanup and the process does not terminate gracefully.

The driver-loop failure itself is deterministic with this state transition:

```text
poll_timeout()          -> T, where T <= now
handle_timeout(now)     -> Ok(())
poll_timeout()          -> the same T
```

The current loop repeats those three operations forever without polling any external input.

## Expected behavior

The driver should remain responsive even if a Sans-I/O component fails to advance an expired deadline. It should still allow close commands and network input to win promptly, while avoiding a CPU hot loop.

## Proposed fix

After handling an immediate timeout:

1. Poll the next deadline again.
2. If it changed, continue normally so legitimate catch-up work remains fast.
3. If it is unchanged, arm a short capped exponential backoff and enter the normal `select!` instead of immediately continuing.
4. Do not invoke `handle_timeout` a second time merely because that defensive backoff elapsed.

This keeps the healthy path unchanged, makes a non-advancing protocol timer observable, and preserves responsiveness to close/network events.

I will follow with a PR containing the guard and regression tests.

## Environment

- Reproduced with `webrtc` v0.20.0
- Still present on `master` at `0ea042e7aeb44b967f9edac42bd917fb30af796e`
- Rust 1.96.0
- macOS arm64

alinejun:
## Summary

- re-poll the aggregate Sans-I/O deadline after handling an already-expired timeout
- detect an unchanged deadline and replace the unconditional immediate loop with a capped exponential backoff (10 ms to 1 s)
- keep the normal `select!` active during that backoff so close commands and network input remain responsive
- skip a duplicate `handle_timeout` call when the defensive backoff timer itself expires
- reset the guard immediately once the deadline advances, leaving normal timeout catch-up unchanged

## Why

An expired deadline is normal in an async driver, but the previous unconditional `handle_timeout(...); continue;` assumed every timeout source always advanced in one call. If the core, STUN, or TURN layer returned the same expired deadline, the driver could hot-loop before reaching `select!`, starving the inputs needed to recover or close the connection.

The guard only activates when the exact handled deadline is returned again. Advancing deadlines keep the existing fast path.

Fixes #862.

## Tests

- `cargo fmt --all -- --check`
- `cargo clippy --fix --lib --bins --all-targets --allow-dirty`
- `cargo test --quiet`
- `cargo test --lib --quiet --no-default-features --features runtime-smol,crypto-ring`
- `cargo test --lib --quiet --no-default-features --features runtime-mock,crypto-ring`
- new unit coverage for exponential capping and reset-on-progress behavior

### DIFF
--- a/src/peer_connection/driver.rs
+++ b/src/peer_connection/driver.rs
@@ -102,6 +102,49 @@ pub(crate) const DRIVER_TO_TRACK_REMOTE_EVENT_CHANNEL_CAPACITY: usize = 256;
 pub(crate) const DRIVER_TO_TRACK_LOCAL_EVENT_CHANNEL_CAPACITY: usize = 256;
 
 const DEFAULT_TIMEOUT_DURATION: Duration = Duration::from_secs(86400); // 1 day duration
+const NON_ADVANCING_TIMEOUT_INITIAL_BACKOFF: Duration = Duration::from_millis(10);
+const NON_ADVANCING_TIMEOUT_MAX_BACKOFF: Duration = Duration::from_secs(1);
+
+/// Prevent a non-advancing Sans-I/O timeout from monopolizing the async executor.
+///
+/// Expired deadlines are normal: the executor can wake late, and a protocol can request an
+/// immediate callback. After `handle_timeout`, however, returning the exact same deadline means
+/// another immediate call cannot make timer progress. Back off exponentially while it remains
+/// unchanged, but keep the driver's `select!` active so network input and close commands can
+/// still move the connection forward.
+#[derive(Debug, Default)]
+struct ImmediateTimeoutGuard {
+    stalled_deadline: Option<Instant>,
+    consecutive_stalls: u32,
+}
+
+impl ImmediateTimeoutGuard {
+    fn observe(&mut self, handled_deadline: Instant, next_deadline: Instant) -> Option<Duration> {
+        if next_deadline != handled_deadline {
+            self.reset();
+            return None;
+        }
+
+        if self.stalled_deadline == Some(handled_deadline) {
+            self.consecutive_stalls = self.consecutive_stalls.saturating_add(1);
+        } else {
+            self.stalled_deadline = Some(handled_deadline);
+            self.consecutive_stalls = 1;
+        }
+
+        let exponent = self.consecutive_stalls.saturating_sub(1).min(31);
+        Some(
+            NON_ADVANCING_TIMEOUT_INITIAL_BACKOFF
+                .saturating_mul(1_u32 << exponent)
+                .min(NON_ADVANCING_TIMEOUT_MAX_BACKOFF),
+        )
+    }
+
+    fn reset(&mut self) {
+        self.stalled_deadline = None;
+        self.consecutive_stalls = 0;
+    }
+}
 
 /// Insert `sender` for `channel_id`, returning `true` if the channel should be announced.
 pub(crate) fn insert_data_channel_event_sender(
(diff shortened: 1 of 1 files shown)


src/peer_connection/driver.rs
```diff
@@ -102,6 +102,49 @@ pub(crate) const DRIVER_TO_TRACK_REMOTE_EVENT_CHANNEL_CAPACITY: usize = 256;
 pub(crate) const DRIVER_TO_TRACK_LOCAL_EVENT_CHANNEL_CAPACITY: usize = 256;
 
 const DEFAULT_TIMEOUT_DURATION: Duration = Duration::from_secs(86400); // 1 day duration
+const NON_ADVANCING_TIMEOUT_INITIAL_BACKOFF: Duration = Duration::from_millis(10);
```
rainliu, line 105: let's use `Duration::from_millis(1)` as initial value

rainliu (commented):
minior changes.

Could you also submit another PR to v0.20.0 branch so that we can patch it too?

rainliu (changes requested):
in https://github.com/webrtc-rs/webrtc/issues/862, you mentioned
```
Reproduction
The production sequence was:

Establish a TURN-backed peer connection.
Remove network connectivity without closing the peer connection.
Wait for ICE to transition through Disconnected to Failed.
Observe recurring TURN transaction timed out messages.
Request peer close / process shutdown.
The stale peer does not complete cleanup and the process does not terminate gracefully.
The driver-loop failure itself is deterministic with this state transition:

poll_timeout()          -> T, where T <= now
handle_timeout(now)     -> Ok(())
poll_timeout()          -> the same T
The current loop repeats those three operations forever without polling any external input.
```

we should also identify the root cause of why poll_timeout always return expired T, in order to figure the underlying bug too.

Do you know which module cause such issue?

rainliu:
Actually, I submitted a PR to fix this issue in other way, instead of using back-off, could you check whether current master branch?

PR #869, opened 2026-08-11
Re-export the remaining public-signature argument types

yexiyue:
### Summary

- re-export `MulticastDnsMode`, `NetworkType`, `RTCDtlsRole`, `SctpMaxMessageSize` and
  `CertificateParams` from `webrtc::peer_connection`;
- state the rule the existing re-exports follow, and list what it currently covers;
- add a test that names each of those types through `webrtc::` and never imports `rtc`.

### Why

`rtc` is a private dependency — it is a git submodule of this repository, referenced as
`rtc = { version = "…", path = "rtc" }`. An application that needs to name one of these
types has to add its own `rtc` dependency, which resolves to a *different* source than the
submodule. The two copies are then distinct types and the call does not compile:

```text
expected trait `webrtc::peer_connection::rtc_crypto::RTCCrypto`,
   found trait `rtc::rtc_crypto::RTCCrypto`
```

Both spellings look identical in the error, which is a confusing way to discover that a
re-export is missing. `[patch.crates-io]` does not help either — it cannot reach a path
dependency.

The comment above the existing re-exports already states exactly this reason, for
`set_dtls_cipher_suites` and `set_srtp_protection_profiles`. The five types added here sit
in the same position — arguments to `SettingEngineBuilder` methods and to
`RTCCertificate::generate` — so this looks like the rule not having been applied
uniformly rather than a deliberate omission.

For what it's worth, `tests/sctp_receive_buffer.rs` already reaches around it in-tree
(`use rtc::peer_connection::configuration::setting_engine::SettingEngineBuilder;`), which
works only because the test shares this workspace.

I ran into this porting a libp2p WebRTC transport from 0.20 to 0.21: it configures
ICE-lite, disables mDNS, restricts the network types and answers as the DTLS server, so it
needs four of the five, plus `CertificateParams` for the persisted DTLS certificate.

**Not included:** `UDPNetwork`, `InterfaceFilterFn` and `IpFilterFn`. They belong to the
same rule, but their imports are still `//TODO:`-commented in `setting_engine.rs`, so I
left them for whoever wires those setters up.

### Validation

- `cargo test --test public_api_is_nameable_from_webrtc` — 2 passed
- `cargo test --lib` — 25 passed
- `cargo fmt --check`, `cargo clippy --lib` — clean
- Negative check: downgrading the new `pub use rtc::ice::{…}` to a private `use` makes the
  new test fail with `E0603: enum MulticastDnsMode is private`, so it does catch the
  regression it is meant to catch.

One incidental change: the private `use rtc::ice::mdns::MulticastDnsMode;` at the top of
`peer_connection/mod.rs` is removed, since the new `pub use` brings the same name into
scope and the two collide (`E0252`).

### AI disclosure

This PR was co-authored by Claude Opus 5 (Claude Code). The missing-symbol audit was done
by walking every `pub fn` on `SettingEngineBuilder` and `RTCCertificate` and checking each
non-primitive parameter type against the re-exports in `peer_connection/mod.rs`; the
patch, the test and the negative check above were produced and run in that session.

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -74,7 +74,6 @@ use transport::stun_gatherer::RTCStunGatherer;
 use transport::turn_relayer::RTCTurnRelayer;
 
 use rtc::data_channel::{RTCDataChannelId, RTCDataChannelInit};
-use rtc::ice::mdns::MulticastDnsMode;
 use rtc::mdns::MulticastSocket;
 use rtc::peer_connection::RTCPeerConnectionBuilder;
 use rtc::peer_connection::configuration::{RTCAnswerOptions, RTCOfferOptions};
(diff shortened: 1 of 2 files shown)


rainliu (changes requested):
You don't need to re-expose them, just directly 
```
use rtc::ice::mdns::MulticastDnsMode;
use rtc::ice::network_type::NetworkType;
use rtc::peer_connection::certificate::CertificateParams;
use rtc::peer_connection::configuration::setting_engine::SctpMaxMessageSize;
use rtc::peer_connection::transport::RTCDtlsRole;
use webrtc::peer_connection::{RTCCertificate, SettingEngineBuilder, crypto};
```
in your test file

PR #870, opened 2026-08-11
Apply the SettingEngine TURN allocation Refresh cap

duszmox:
Depends on [webrtc-rs/rtc#166](https://github.com/webrtc-rs/rtc/pull/166)

### Summary

- update the `rtc` submodule to include the TURN `SettingEngine` option;
- read the allocation Refresh interval cap during peer-connection construction;
- forward it through `PeerConnectionImpl` and `RTCTurnRelayer`;
- apply it to each `rtc-turn::ClientConfig`.

### Why

The async WebRTC layer owns TURN client construction, while advanced transport policy is configured through `SettingEngine`.

This change bridges those layers: the async wrapper reads the configured cap before moving the setting engine into the core peer connection, then applies the same value to every TURN client it creates.

A shorter opt-in cap can keep an otherwise idle client-to-server NAT mapping active. When no cap is configured, TURN retains its existing half-allocation-lifetime refresh cadence.

### Compatibility

The option remains disabled by default, so existing applications have no behavioral change.

### AI disclosure

This change was developed with GPT-5.6-sol assistance.

### DIFF
--- a/src/peer_connection/mod.rs
+++ b/src/peer_connection/mod.rs
@@ -57,7 +57,7 @@ use log::error;
 use std::collections::{HashMap, HashSet};
 use std::net::ToSocketAddrs;
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use crate::data_channel::{DataChannel, DataChannelEvent, DataChannelImpl};
 use crate::media_stream::{track_local::TrackLocal, track_remote::TrackRemote};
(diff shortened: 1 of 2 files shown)


rainliu (changes requested):
rebase rtc pointer to its master

duszmox:
I bumped it to the master's latest head, if you only want me to bump it to my commit's hash please let me know!

rainliu (approved):

PR #871, opened 2026-08-11
fix(peer-connection): rebind UDP sockets on ICE restart

### LINKED ISSUE #868: Android ICE restart stalls after ~10s background; rebinding UDP sockets restores recovery
## Summary

On Android, returning an application to the foreground after approximately 10 seconds in the background can leave an established `PeerConnection` unable to recover through ICE restart.

The signaling connection becomes available and the ICE restart offer/answer exchange succeeds, but the restarted ICE generation remains in `Checking`. The existing data channels do not become sendable. The application waits for its recovery deadline, destroys the stale `PeerConnection`, and establishes a completely new connection before traffic can resume.

In the captured 9.7-second background run:

- signaling was connected immediately after foregrounding;
- two ICE restart offer/answer exchanges completed, but neither reached `Connected`;
- the application remained unable to send for approximately 10 seconds;
- rebuilding the `PeerConnection` and completing the echo took another 2,468 ms; and
- foreground-to-response latency was approximately 12.5 seconds.

The failure is caused by a mismatch between the restarted logical ICE generation and the UDP transport retained by the existing `PeerConnection`. Android backgrounding can leave the old UDP socket or its network path unusable, while ICE restart updates credentials and candidate-pair state without creating a fresh UDP socket generation. Signaling and SDP negotiation therefore succeed, but no viable candidate pair is formed.

In one Android native capture the old socket returned `ConnectionAborted`, was removed by the driver, and subsequent connectivity-check writes for its local address were silently dropped as `Ok(0)`. This is one observable manifestation of the problem; the same transport-generation mismatch can also occur when the socket remains present but its old network path, NAT mapping, or TURN allocation is no longer usable.

An opt-in fork that binds fresh UDP sockets and discards the previous local candidate generation during ICE restart recovered the existing connection in 360-560 ms in two physical-device runs with approximately 10 seconds in the background.

## Environment

- `webrtc` crate: `0.20.1`
- Device: Huawei Mate 40 (`OCE-AN10`)
- Android version: 12
- Network: Wi-Fi for the complete test; no Wi-Fi/cellular handover
- Workload: WebRTC data channels
- ICE configuration: STUN and TURN configured, `RTCIceTransportPolicy::All`, relay not forced
- 10-second capture roles: Android is the Offerer; native service peer is the Answerer
- Native socket-failure capture roles: Android is the Answerer; native service peer is the Offerer

The physical-device runs use `webrtc` 0.20.1. Static inspection of `master` at `c1d3860d913f7a128cff75edb4a15b0cb4fb966d` (`0.21.0-alpha.1`) shows the same relevant socket lifetime and ICE-restart behavior. `master` has not been exercised in this physical-device test.

## Reproduction steps

1. Establish a `PeerConnection` between the Android application and the service.
2. Wait until ICE is `Connected` and all data channels are open.
3. Verify that an echo request succeeds.
4. Put the Android application in the background for approximately 10 seconds. Keep the service running and leave the device on the same Wi-Fi network.
5. Return the application to the foreground without recreating the `PeerConnection`.
6. Perform a normal ICE restart:
   - the Offerer calls `create_offer` with ICE restart enabled;
   - the remote peer applies the offer;
   - the Answerer creates and applies an answer; and
   - both peers apply the new descriptions.
7. Observe the restarted ICE state and attempt to resume data-channel traffic.

The failure is intermittent because Android does not invalidate the old UDP path in every suspension. A five-round stress run using approximately 59 seconds in the background reproduced the failure in 3/5 rounds.

## 10-second failure timeline

Android application log, UTC+08:00:

```text
14:48:03 App lifecycle: Background

14:48:13 signaling state=DISCONNECTED
14:48:13 WebRTC state=recovering, event=DISCONNECTED
14:48:13 App lifecycle: Foreground (bg_duration=9717ms)
14:48:13 foreground recovery state=complete
14:48:13 signaling state=CONNECTED

14:48:23 WebRTC state=idle, event=DISCONNECTED
14:48:23 sending echo
14:48:23 WebRTC state=connecting
14:48:25 WebRTC state=connected
14:48:25 echo done: send-to-response 2468ms, succeeded=true
```

The lifecycle reconciliation and signaling recovery finish at `14:48:13`, but WebRTC does not become usable. At `14:48:23`, the stale session reaches its recovery deadline and becomes idle. Sending the echo then creates a new `PeerConnection`; the request completes at `14:48:25`.

Service log for the same run, UTC:

```text
06:48:09.662 PeerConnection -> Disconnected
06:48:09.662 requested ICE restart from the Android Offerer

06:48:13.055 received ICE restart offer
06:48:13.056 completed ICE restart answer
06:48:13.056 ICE -> Checking

06:48:20.085 received another ICE restart offer
06:48:20.087 completed ICE restart answer
...ICE never reaches Connected...

06:48:23.260 received a fresh, non-restart offer
06:48:23.261 closed the stale PeerConnection and created a new one
06:48:23.914 new ICE connection -> Connected
06:48:25.480 new data channels opened; connection became sendable
```

The two endpoints agree on the failure sequence: restart signaling works, the restarted ICE generation cannot establish connectivity, and only a full `PeerConnection` replacement restores traffic.

## Android native socket observation

One native capture contains this sequence immediately after foregrounding:

```text
ERROR webrtc::peer_connection::driver:
  Socket recv error on <mobile-address>:
  Software caused connection abort (os error 103)

WARN webrtc::peer_connection::driver:
  None tcp/udp socket, drop the packet to <turn-server>
  from <mobile-address> for UDP

received ICE restart offer with new ICE credentials
created and applied ICE restart answer
ICE connection state changed to Checking

WARN webrtc::peer_connection::driver:
  None tcp/udp socket, drop the packet to <turn-server>
  from <mobile-address> for UDP

...the same warning continues for connectivity-check and STUN writes...
```

The corresponding service-side recovery loop sends seven ICE restart attempts. Every attempt times out, followed by:

```text
peer recovery deadline expired (elapsed_ms=60001)
```

This capture demonstrates the missing-socket variant of the transport-generation mismatch. A stale path does not necessarily produce the same explicit socket error, so socket removal is not a prerequisite for the restart failure.

In the five-round stress run, the outcome was:

| Run | Background duration | Echo result | Recovery deadline teardowns |
| --- | ---: | --- | ---: |
| r1 | duration field contaminated by retained state | succeeded in 687 ms | 0 |
| r2 | 59,124 ms | timed out after 30,095 ms | 7 |
| r3 | 59,115 ms | timed out after 30,105 ms | 7 |
| r4 | 59,117 ms | timed out after 30,094 ms | 7 |
| r5 | 59,139 ms | succeeded in 3,203 ms | 0 |

Raw log files are not attached because surrounding signaling records contain identity material and credentials. Endpoint addresses are redacted in the excerpts.

## Root cause analysis

The failure is caused by a mismatch between the restarted logical ICE generation and the UDP transport retained by the existing `PeerConnection`.

### 1. UDP sockets are bound for the lifetime of the PeerConnection

The `PeerConnection` binds its UDP sockets during construction. A normal ICE restart updates ICE credentials and candidate-pair state, but does not create a new UDP socket generation.

### 2. Android backgrounding can invalidate the existing UDP path

After the Android application has been suspended, the previously bound UDP socket or its associated network path may no longer be usable.

In one native capture, the first receive after foregrounding returned `ConnectionAborted` and the driver removed the socket. In other reproductions, the socket may remain present while its previous network path, NAT mapping, or TURN allocation is no longer usable. A missing socket entry is therefore one observable manifestation of the problem, but is not required for the failure.

### 3. ICE restart reuses the old physical transport

[`restart_ice()`](https://github.com/webrtc-rs/webrtc/blob/c1d3860d913f7a128cff75edb4a15b0cb4fb966d/src/peer_connection/mod.rs#L1028-L1040) restarts the sans-I/O ICE state and requests candidate gathering, but it does not bind a fresh UDP transport for the restarted generation.

If a socket was removed, the restarted ICE generation can retain or gather a logical local candidate whose address has no corresponding entry in `udp_sockets`. Connectivity-check writes for that address are logged as `None tcp/udp socket`, dropped, and returned as `Ok(0)`: [`driver.rs`](https://github.com/webrtc-rs/webrtc/blob/c1d3860d913f7a128cff75edb4a15b0cb4fb966d/src/peer_connection/driver.rs#L632-L642).

If the socket remains present but its path is stale, connectivity checks can be issued without producing a viable candidate pair.

In both cases, fresh ICE credentials and SDP are exchanged successfully, but the restarted ICE generation remains in `Checking`.

### 4. Rebuilding the PeerConnection creates a working transport

A new `PeerConnection` binds new UDP sockets, creates new STUN/TURN state, and gathers a new local candidate generation. It connects successfully on the same device and network where ICE restart on the existing `PeerConnection` failed.

## Validated fix

The test fork adds an opt-in builder setting; the default behavior remains unchanged:

```rust
PeerConnectionBuilder::new()
    .with_fresh_udp_sockets_on_ice_restart(true)
```

It is used together with:

```rust
setting_engine.set_discard_local_candidates_during_ice_restart(true);
```

Discarding the previous local candidate generation ensures that candidates backed by the old UDP sockets are not retained after those sockets have been replaced.

The implementation:

1. Records the configured UDP bind addresses when the `PeerConnection` is created.
2. Detects a new ICE restart generation:
   - Offerer: around `create_offer`;
   - Answerer: when `set_remote_description` applies new remote ICE credentials.
3. Marks UDP socket rebinding as pending for that generation.
4. Defers restarted candidate gathering until `set_local_description` is applied.
5. In the driver:
   - clears receive futures and all references to the old UDP sockets;
   - binds the configured UDP addresses again;
   - updates the STUN gatherer and TURN relayer with the new local addresses; and
   - gathers candidates for the restarted ICE generation over the new sockets.

For an ephemeral `:0` configuration, rebinding produces a fresh local port and mapping.

Relevant commits:

- https://github.com/alinejun/webrtc/commit/e7a18bf8917da41e0ca284612a0cc80a834c99bd
- https://github.com/alinejun/webrtc/commit/8b4ce31f9aaef2d97dc41acee77b9e36bce53faa
- Validated v0.20.1 ICE restart generation hook: https://github.com/alinejun/rtc/commit/970535ac87870773a25a28008eeea1aa9115d72a
- Master-compatible ICE restart generation hook: https://github.com/alinejun/rtc/commit/492959c3df52ae935c20b48ffffa1b751a87cdf6

## Physical-device validation

The following runs used the same Android device, service, network, and ICE configuration:

| Build | Background duration | ICE recovery | Echo call | Foreground-to-response |
| --- | ---: | --- | ---: | ---: |
| `webrtc` 0.20.1 | 9,717 ms | two restarts remained in `Checking`; full connection rebuild after about 10 s | 2,468 ms | about 12.5 s |
| fork at `8b4ce31f` | 10,472 ms | `Checking -> Connected` in 560 ms; existing data channels reused | 210 ms | about 1 s |
| fork at `8b4ce31f` | 10,013 ms | `Checking -> Connected` in 360 ms; existing data channels reused | 176 ms | about 1 s |

First patched run, service log:

```text
07:22:26.255 received ICE restart offer
07:22:26.256 applied answer; ICE -> Checking
07:22:26.816 ICE -> Connected
07:22:26.816 existing data channels are open; connection became sendable
07:22:27.023 echo request received
```

Second patched run, service log:

```text
07:23:40.445 received ICE restart offer
07:23:40.451 applied answer; ICE -> Checking
07:23:40.810 ICE -> Connected
07:23:40.810 existing data channels are open; connection became sendable
07:23:41.015 echo request received
```

The A/B result shows that creating a fresh UDP transport and local candidate generation allows the existing `PeerConnection` and data channels to recover without waiting for a full rebuild.

## Expected behavior

Applications should be able to request a fresh UDP transport generation for an ICE restart.

When this behavior is enabled:

1. the previous UDP sockets should be released;
2. new UDP sockets should be bound from the configured bind addresses;
3. STUN and TURN gathering state should use the new local addresses;
4. local candidates from the previous socket generation should be discarded; and
5. candidate gathering should begin only after the restart local description is applied.

The default behavior should remain unchanged for applications that do not enable this option.

## Proposed upstream change

The validated implementation is opt-in and leaves the current default behavior unchanged:

```rust
PeerConnectionBuilder::new()
    .with_fresh_udp_sockets_on_ice_restart(true)
```

When enabled, every applied ICE restart creates a fresh UDP socket and local candidate generation. This provides deterministic behavior for mobile applications where an operating-system suspension may invalidate a UDP path without providing a reliable persistent socket-failure signal.

Upstream PR: https://github.com/webrtc-rs/webrtc/pull/871

alinejun:
## Summary

- add an opt-in `with_fresh_udp_sockets_on_ice_restart` builder setting
- replace configured UDP sockets when an ICE restart is applied
- reset STUN gathering and TURN relaying with the new local addresses
- defer restart gathering until `set_local_description` so the new transport and SDP generation stay aligned
- keep the existing behavior unchanged by default

## Motivation

Android may leave the UDP socket or its network path unusable after an application is suspended and returned to the foreground. Signaling and the restart offer/answer exchange still succeed, but ICE can remain in `Checking` because the restarted logical ICE generation continues to use the previous UDP transport generation.

With this option enabled, the driver drops references to the previous UDP sockets, binds the configured addresses again, refreshes STUN/TURN local addresses, and gathers candidates over the replacement sockets. For an ephemeral `:0` binding this produces a fresh local port and network mapping.

The setting is intended to be paired with:

```rust
setting_engine.set_discard_local_candidates_during_ice_restart(true);
```

This prevents candidates backed by the previous socket generation from being retained.

Fixes #868.

## Compatibility

The new behavior is opt-in:

```rust
PeerConnectionBuilder::new()
    .with_fresh_udp_sockets_on_ice_restart(true)
```

The default remains `false`, so existing applications retain the current ICE restart behavior.

## RTC dependency

The async wrapper needs to observe when the sans-I/O core actually applies an ICE restart. This branch pins [`alinejun/rtc@47e6abd`](https://github.com/alinejun/rtc/commit/47e6abd5f8fe98b68f1f89dfb77b3073ddd2444f), which exposes the applied ICE restart generation. The prerequisite is proposed upstream in [webrtc-rs/rtc#167](https://github.com/webrtc-rs/rtc/pull/167).

## Validation

Physical Android tests and detailed endpoint timelines are documented in #868. With the option enabled, two approximately 10-second background runs recovered the existing PeerConnection and data channels in 560 ms and 360 ms respectively, instead of waiting for a full connection rebuild.

Local checks completed:

- `cargo check --lib`
- `cargo fmt`
- `cargo clippy --fix --lib --bins --all-targets --allow-dirty`
- `cargo clippy --lib --bins --all-targets`

The full test suite was not run.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ crypto-ring = ["rtc/crypto-ring"]
 crypto-aws-lc-rs = ["rtc/crypto-aws-lc-rs"]
 
 [dependencies]
-rtc = { git = "https://github.com/alinejun/rtc", rev = "492959c3df52ae935c20b48ffffa1b751a87cdf6", default-features = false }
+rtc = { git = "https://github.com/alinejun/rtc", rev = "47e6abd5f8fe98b68f1f89dfb77b3073ddd2444f", default-features = false }
 
 bytes = "1.12.0"
 async-trait = "0.1.89"
@@ -50,7 +50,7 @@ tokio = { version = "1.52.3", features = ["net", "time", "sync", "rt", "rt-multi
 smol = { version = "2.0.2", optional = true }
 
 [dev-dependencies]
-signal = { git = "https://github.com/alinejun/rtc", rev = "492959c3df52ae935c20b48ffffa1b751a87cdf6", package = "rtc-signal" }
+signal = { git = "https://github.com/alinejun/rtc", rev = "47e6abd5f8fe98b68f1f89dfb77b3073ddd2444f", package = "rtc-signal" }
 
 env_logger = "0.11.11"
 chrono = "0.4.45"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ crypto-ring = ["rtc/crypto-ring"]
 crypto-aws-lc-rs = ["rtc/crypto-aws-lc-rs"]
 
 [dependencies]
-rtc = { version = "0.21.0-alpha.1", path = "rtc", default-features = false }
+rtc = { git = "https://github.com/alinejun/rtc", rev = "492959c3df52ae935c20b48ffffa1b751a87cdf6", default-features = false }
 
 bytes = "1.12.0"
 async-trait = "0.1.89"
@@ -50,7 +50,7 @@ tokio = { version = "1.52.3", features = ["net", "time", "sync", "rt", "rt-multi
 smol = { version = "2.0.2", optional = true }
 
 [dev-dependencies]
-signal = { version = "0.21.0-alpha.1", path = "rtc/examples/signal", package = "rtc-signal" }
+signal = { git = "https://github.com/alinejun/rtc", rev = "492959c3df52ae935c20b48ffffa1b751a87cdf6", package = "rtc-signal" }
 
 env_logger = "0.11.11"
 chrono = "0.4.45"
--- a/src/peer_connection/driver.rs
+++ b/src/peer_connection/driver.rs
@@ -183,7 +183,9 @@ pub(crate) enum PeerConnectionDriverEvent {
         ice_servers: Vec<RTCIceServer>,
         ice_transport_policy: RTCIceTransportPolicy,
     },
-    IceGathering,
+    IceGathering {
+        rebind_udp_sockets: bool,
+    },
     Close,
 }
 
@@ -200,6 +202,7 @@ where
     tcp_transport: RTCTcpTransport,
     mdns_socket: Option<Arc<dyn AsyncUdpSocket>>,
     udp_sockets: HashMap<SocketAddr, Arc<dyn AsyncUdpSocket>>,
+    udp_bind_addrs: Vec<SocketAddr>,
     /// Reused scratch buffer for concatenating a run of same-destination datagrams
     /// into one UDP GSO send (see [`flush_writes`](Self::flush_writes)).
     gso_scratch: Vec<u8>,
@@ -232,6 +235,7 @@ where
         turn_relayer: RTCTurnRelayer,
         mdns_socket: Option<Arc<dyn AsyncUdpSocket>>,
         udp_sockets: HashMap<SocketAddr, Arc<dyn AsyncUdpSocket>>,
+        udp_bind_addrs: Vec<SocketAddr>,
         tcp_listeners: HashMap<SocketAddr, Arc<dyn AsyncTcpListener>>,
     ) -> Result<Self> {
         if udp_sockets.is_empty() && tcp_listeners.is_empty() {
@@ -244,6 +248,7 @@ where
             turn_relayer,
             mdns_socket,
             udp_sockets,
+            udp_bind_addrs,
             gso_scratch: Vec::new(),
             tcp_transport: RTCTcpTransport::new(tcp_listeners),
             ice_gathering_active: false,
@@ -252,6 +257,23 @@ where
             pending_ice_configuration: None,
             pending_data_channel_events: HashMap::new(),
         })
+    }
+
+    fn rebind_udp_sockets(&mut self) -> Result<()> {
+        let mut udp_sockets = HashMap::with_capacity(self.udp_bind_addrs.len());
+        for bind_addr in &self.udp_bind_addrs {
+            let socket = std::net::UdpSocket::bind(bind_addr)?;
+            socket.set_nonblocking(true)?;
+            let local_addr = socket.local_addr()?;
+            udp_sockets.insert(local_addr, self.inner.runtime.wrap_udp_socket(socket)?);
+        }
+
+        let local_addrs = udp_sockets.keys().copied().collect::<Vec<_>>();
+        self.stun_gatherer
+            .restart_with_local_addrs(local_addrs.clone());
+        self.turn_relayer.restart_with_local_addrs(local_addrs);
+        self.udp_sockets = udp_sockets;
+        Ok(())
     }
 
     /// Mark the connection closing and wake any sender parked in send back-pressure.
@@ -277,7 +299,7 @@ where
         mut driver_event_rx: Receiver<PeerConnectionDriverEvent>,
     ) -> Result<()> {
         // Collect socket info into a vec for indexed access
-        let udp_socket_list: Vec<(SocketAddr, Arc<dyn AsyncUdpSocket>)> = self
+        let mut udp_socket_list: Vec<(SocketAddr, Arc<dyn AsyncUdpSocket>)> = self
             .udp_sockets
             .iter()
             .map(|(addr, sock)| (*addr, sock.clone()))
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -30,7 +30,7 @@ crypto-ring = ["rtc/crypto-ring"]
 crypto-aws-lc-rs = ["rtc/crypto-aws-lc-rs"]
 
 [dependencies]
-rtc = { version = "0.21.0-alpha.1", path = "rtc", default-features = false }
+rtc = { git = "https://github.com/alinejun/rtc", rev = "492959c3df52ae935c20b48ffffa1b751a87cdf6", default-features = false }
```
rainliu: any change on rtc crate? if yes, please submit PR on rtc repo too.

rainliu (changes requested):

rainliu:
If your changes are against v0.20, you can change base branch from master to v0.20.x branch in this PR.

rainliu:
I drafted a PR https://github.com/webrtc-rs/webrtc/pull/872 to fix the issue #868. Please take a look. I will close this PR in favorite of PR 872

PR #885, opened 2026-08-29
wrap deferred SCTP stream-id data channels

### LINKED ISSUE #199: RFC: Defer SCTP stream-id assignment until DTLS role is resolved (breaking API change)
When both peers rapidly create data channels before the DTLS role is negotiated, both are in the `Auto` role state. The current implementation treats `Auto` as server parity (odd ids), so both peers allocate odd identifiers (1, 3, 5...). When the DTLS handshake completes and one peer becomes the client, that peer should use even ids per RFC 8832 §6, but its channels already have odd ids assigned. This causes stream-id collisions and violates the spec's parity requirement.

The fix defers stream-id assignment until the DTLS role is resolved, aligning with the W3C WebRTC spec which states that the `id` attribute "is initially null" and "will be populated during the RTCSctpTransport connected procedure." This is a breaking change: `RTCDataChannel::id()` now returns `Option<RTCDataChannelId>` (None until assigned), and stream ids are assigned during the SCTP connected procedure per RFC 8832 §6, ensuring correct parity based on the resolved role.

I've implemented this in two fork branches ([rtc](https://github.com/dimitri-lenkov/rtc/tree/pr-009-data-channel-stream-id-deferral) and [webrtc](https://github.com/dimitri-lenkov/webrtc/tree/pr-009-data-channel-stream-id-deferral)) and would like to gauge whether upstream would be open to this breaking change, or if there's an alternative approach that maintains API compatibility.

### Spec References

> From [RFC 8832 §6 - Procedures](https://www.rfc-editor.org/rfc/rfc8832.html#section-6):
>
> "The peer that initiates opening a data channel selects a stream identifier for which the corresponding incoming and outgoing streams are unused. **If the side is acting as the DTLS client, it MUST choose an even stream identifier; if the side is acting as the DTLS server, it MUST choose an odd one.**"

> From [W3C WebRTC §6.2 - RTCDataChannel.id](https://www.w3.org/TR/webrtc/#dom-rtcdatachannel-id):
>
> "The id attribute returns the ID for this RTCDataChannel. **The value is initially null, which is what will be returned if the ID was not provided at channel creation time, and the DTLS role of the SCTP transport has not yet been negotiated.** Otherwise, it will return the ID that was either selected by the script or generated by the user agent according to [RFC8832]."

> From [W3C WebRTC §6.2 - createDataChannel algorithm](https://www.w3.org/TR/webrtc/#dom-peerconnection-createdatachannel):
>
> "If the [[DataChannelId]] slot is null after this step, **it will be populated during the RTCSctpTransport connected procedure.**"

dimitri-lenkov:
This PR wraps the `rtc` stream-id deferral ([https://github.com/webrtc-rs/rtc/pull/209](https://github.com/webrtc-rs/rtc/pull/209)) for the async API. In [https://github.com/webrtc-rs/rtc/pull/209](https://github.com/webrtc-rs/rtc/pull/209), `rtc` defers data channel stream id assignment until the DTLS role is resolved (W3C section 6.1 step 18 / section 6.1.1.3, RFC 8832 section 6); a channel created while the role is `Auto` has no id until SCTP connects. 

This PR bridges that into the wrapper so `DataChannel::id()` returns `Option` and resolves once the role is known. Closes [https://github.com/webrtc-rs/rtc/issues/199](https://github.com/webrtc-rs/rtc/issues/199).

**Depends on** [https://github.com/webrtc-rs/rtc/pull/209](https://github.com/webrtc-rs/rtc/pull/209): the `rtc` deferral must merge first.

### Breaking

- `DataChannel::id()` returns `Option<RTCDataChannelId>` (`None` while unassigned).

### Handle-based wrapper identity (internal)

- `DataChannelImpl` holds a stable handle plus an async `runtime::primitives::Mutex<Option<RTCDataChannelId>>`:
  - The driver calls `set_id().await` once the core assigns the id.
  - `id()` is `async` and locks the runtime `Mutex` to read the assigned id.
- `PeerConnectionRef` routes events by handle:
  - Adds a handle-keyed `data_channel_impls` map and a stream-id reverse map (`data_channel_ids`).
  - Each `create_data_channel` captures the handle and pre-registers any immediately-assigned id (negotiated channels, or channels created after the SCTP association exists).
  - Pre-registration lets the driver route an `OnClose` that arrives before its `OnOpen` instead of dropping it.

### Reference-cycle fix

- `data_channel_impls` now holds `Weak<DataChannelImpl>` instead of a strong `Arc`, breaking the cycle (`DataChannelImpl` -> `PeerConnectionRef` -> impl map) that leaked the connection until the process exited.
- The driver upgrades the weak reference when routing an event and drops it afterwards.
- A dropped *local* channel (tracked in a new `locally_created_handles` set) is never re-announced.
- A *remote* channel without a live wrapper is re-created and announced as before.

### Fresh-handle `OnOpen` routing

- Events are routed by the handle re-derived from the updated reverse map after `OnOpen`, so an open that corrected a stale stream-id -> handle mapping reaches the resolved channel rather than the stale one.

### Reverse-map cleanup

- A closed or dropped channel no longer leaves a stale stream-id -> handle entry:
  - `DataChannelImpl::Drop` best-effort removes its handle from `data_channel_ids` (via `try_lock`, non-blocking) alongside the existing `data_channel_events_tx` / `data_channel_impls` cleanup.
  - The driver's `OnClose` handling removes the stream id from `data_channel_ids` instead of re-inserting it, so a reused stream id cannot alias a dead handle.
  - `Drop` is the fallback that covers the case where no clean `OnClose` arrives (e.g. teardown).

### Handle-keyed routing

- Core event/message routing (already stream-id based in `rtc`) is translated to handle-keyed wrapper maps via `data_channel_ids`, matching the handle-keyed `create_data_channel` path.
- **Handle-keyed overflow routing.** The driver routes the core's `OnErrorByHandle` / `OnCloseByHandle` overflow events (emitted when SCTP stream ids are exhausted) straight to the owning wrapper by handle:
  - The stream-id reverse map is not consulted.
  - The failing channel - which never received an id - is notified correctly rather than misrouted via stream id 0.

The wrapper's public event surface (`DataChannelEvent`, `on_data_channel`) is unchanged.

## Tests

- New integration test `tests/data_channel_id_resolves_after_open.rs`: `test_data_channel_id_is_none_before_open_and_resolves_after` asserts `id() == None` before connection and that `id()` resolves to a non-null, stable stream id after `OnOpen`; `test_negotiated_data_channel_has_id_immediately` asserts a negotiated channel (`RTCDataChannelInit { negotiated: Some(id) }`) reports `id() == Some(id)` immediately at creation, covering the pre-registered-id path in `create_data_channel`.
- New unit tests: `drop_data_channel_allows_peer_connection_drop` (no reference cycle), `onopen_uses_freshly_resolved_handle` (stale reverse-map entry is corrected), and `create_data_channel_after_sctp_no_duplicate_wrapper` (no duplicate wrapper after SCTP). `drop_removes_event_sender` was extended to exercise the real weak-reference path.
- Existing wrapper unit and integration suites pass.

## Spec references

- [W3C WebRTC section 6.2 RTCDataChannel.id](https://www.w3.org/TR/webrtc/#dom-rtcdatachannel-id): the id is null until the DTLS role of the SCTP transport has been negotiated.

  > The value is initially null, which is what will be returned if the ID was not provided at channel creation time, and the DTLS role of the SCTP transport has not yet been negotiated. Otherwise, it will return the ID that was either selected by the script or generated by the user agent according to [RFC8832]. After the ID is set to a non-null value, it will not change.

## Verification

```bash
cargo test --tests -- --test-threads=4
cargo clippy --lib -- -D warnings
```

### DIFF
--- a/src/data_channel/mod.rs
+++ b/src/data_channel/mod.rs
@@ -53,7 +53,8 @@ use std::sync::atomic::Ordering;
 use std::time::Duration;
 
 pub use rtc::data_channel::{
-    RTCDataChannelId, RTCDataChannelInit, RTCDataChannelMessage, RTCDataChannelState,
+    RTCDataChannelHandle, RTCDataChannelId, RTCDataChannelInit, RTCDataChannelMessage,
+    RTCDataChannelState,
 };
 
 /// Object-safe trait exposing all public DataChannel operations.
@@ -77,11 +78,11 @@ pub trait DataChannel: crate::sealed::Sealed + Send + Sync + 'static {
     async fn negotiated(&self) -> Result<bool>;
     /// Returns the identifier for this data channel.
     ///
-    /// The value is initially null — which is what is returned if the id was not provided
-    /// at channel creation time and the DTLS role of the SCTP transport has not yet been
-    /// negotiated. Otherwise it returns the id that was either selected by the application
-    /// or generated. Once set to a non-null value it does not change.
-    fn id(&self) -> RTCDataChannelId;
+    /// `None` (null) is returned if no id was provided at channel creation time and the DTLS
+    /// role of the SCTP transport has not yet been negotiated (W3C section 6.1 step 18). Otherwise it
+    /// returns the id that was either selected by the application or generated. Once set to a
+    /// non-null value it does not change.
+    async fn id(&self) -> Option<RTCDataChannelId>;
     /// Returns the current state of this data channel.
     async fn ready_state(&self) -> Result<RTCDataChannelState>;
     /// Returns the threshold at which the buffered amount is considered to be *high*.
@@ -259,8 +260,12 @@ pub enum DataChannelEvent {
 ///
 /// This wraps a data channel and provides async send/receive APIs.
 pub(crate) struct DataChannelImpl {
-    /// Unique identifier for this data channel
-    id: RTCDataChannelId,
+    /// Stable handle for this data channel, assigned at creation and never changed. The stream
+    /// identifier (`id()`) is only assigned once the DTLS role is negotiated / SCTP connected.
+    handle: RTCDataChannelHandle,
+
+    /// Stream identifier; `None` until the core assigns one once SCTP is connected.
+    id: crate::runtime::primitives::Mutex<Option<RTCDataChannelId>>,
 
     /// Inner PeerConnection Reference
     inner: Arc<PeerConnectionRef>,
@@ -272,15 +277,26 @@ pub(crate) struct DataChannelImpl {
 impl DataChannelImpl {
     /// Create a new data channel wrapper
     pub(crate) fn new(
-        id: RTCDataChannelId,
+        handle: RTCDataChannelHandle,
         inner: Arc<PeerConnectionRef>,
         evt_rx: Receiver<DataChannelEvent>,
     ) -> Self {
         Self {
-            id,
+            handle,
+            id: crate::runtime::primitives::Mutex::new(None),
             inner,
             evt_rx: Mutex::new(evt_rx),
         }
+    }
+
+    /// The stable handle for this data channel.
+    pub(crate) fn handle(&self) -> RTCDataChannelHandle {
+        self.handle
+    }
+
+    /// Stores the stream identifier assigned by the core once SCTP is connected.
+    pub(crate) async fn set_id(&self, id: RTCDataChannelId) {
+        *self.id.lock().await = Some(id);
     }
 
     /// Park until the driver signals send-buffer progress, or a 50 ms liveness backstop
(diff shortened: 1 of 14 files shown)


src/data_channel/mod.rs
```diff
@@ -272,17 +278,29 @@ pub(crate) struct DataChannelImpl {
...
     pub(crate) fn new(
-        id: RTCDataChannelId,
+        handle: RTCDataChannelHandle,
         inner: Arc<PeerConnectionRef>,
         evt_rx: Receiver<DataChannelEvent>,
     ) -> Self {
         Self {
-            id,
+            handle,
+            id: std::sync::Mutex::new(None),
```
rainliu: use runtime::primitives::Mutex, instead of std::sync::Mutex.

src/data_channel/mod.rs
```diff
@@ -372,21 +398,21 @@ impl DataChannel for DataChannelImpl {
...
-            .data_channel(self.id)
+            .data_channel_handle(self.handle)
             .ok_or(Error::ErrDataChannelClosed)?
             .negotiated())
     }
 
     // Documented on `DataChannel::id`.
-    fn id(&self) -> RTCDataChannelId {
-        self.id
+    fn id(&self) -> Option<RTCDataChannelId> {
```
rainliu: with runtime's Mutex, change this fn to async fn is fine.

dimitri-lenkov:
@rainliu thank you for the review --- I've force-pushed the requested changes plus a new test. CI flagged one of my files' patch coverage at only 69.56%. I was not sure how to re-run the coverage tool.

rainliu:
similarly for this PR. 

I am going to close this PR in favor of another PR https://github.com/webrtc-rs/webrtc/pull/886

Issue #894, opened 2026-09-16
turn: release Sans-I/O allocations during relayer retirement and ICE restart


alinejun:
## Summary

The Sans-I/O TURN relayer can leave server allocations active after ICE transport replacement or `PeerConnection` close.

The retirement path removes a managed TURN client and calls the protocol client's `close`. In `rtc-turn`, the protocol client's close only clears transaction state; releasing an allocation requires `Relay::close`, which queues an authenticated `Refresh` with `LIFETIME=0`. The queued release must also be flushed through the old five-tuple and socket before that transport is discarded.

## Reproduction

Run two relay-only peers against an instrumented TURN server and record the server allocation table:

1. Start the server with a 600-second default allocation lifetime.
2. Create two relay-only PeerConnections, one UDP socket per peer, and connect them.
3. Confirm one allocation per peer and complete one data/RPC exchange.
4. For the close case, close both PeerConnections and poll the server allocation table for three seconds.
5. For the restart case, make both peers unavailable, wait for the old ICE state to become non-sendable, restore the network, complete ICE recovery, and perform one data/RPC exchange.
6. Poll the server table for three seconds, retaining the original five-tuples so old allocations can be distinguished from the new generation.

Representative sanitized output from the regression probe:

```text
initial: connected=true  rpc=ok  allocations=2
initial tuples: [client-A:41001 -> turn:3478, client-B:41002 -> turn:3478]

close case:
  PeerConnection A close requested
  PeerConnection B close requested
  after 3.0s: old_allocations_present=2  total_allocations=2
  expected: old_allocations_present=0  total_allocations=0

ICE restart case:
  network unavailable: A=non-sendable  B=non-sendable
  network restored:    A=sendable      B=sendable
  recovery RPC:        success
  before restart:      allocations=2
  after restart:       old_allocations_present=2  total_allocations=4
  expected:            old_allocations_present=0  total_allocations=2
```

The result is a stale allocation window, not evidence of a permanent leak: the server eventually removes the entries when their lifetime expires. On a reachable network, however, the old generation should be released immediately with `Refresh(0)`; otherwise a rapid restart can reuse the same five-tuple while the server still owns the old allocation and return 437.

## Expected behavior

Retiring a TURN client should issue `Refresh(LIFETIME=0)` for each active relay, flush it using the old transport, and only then discard the client and socket. If the network is already unavailable, local retirement must finish within a bounded timeout and the server may reclaim the allocation by expiry.

## Related protocol behavior

[...]

PR #895, opened 2026-09-16
fix(turn): release allocations during Sans-I/O transport retirement

alinejun:
## Summary

Fix TURN allocation retirement in the v0.20 Sans-I/O relayer.

When a TURN client is removed, `Client::close()` only clears transaction state. The allocation release operation is `Relay::close()`, which queues an authenticated `Refresh` with `LIFETIME=0`. The old relayer dropped the client before generating and flushing that request, so the server allocation remained until its normal expiry. ICE rebinding also discarded the old UDP socket before any release request could be sent.

This change:

- clears old client transactions, then queues `Refresh(0)` for every active relay;
- drains those packets into a retirement queue before dropping the client;
- preserves retirement packets across configuration resets;
- flushes them while the old UDP sockets are still alive during ICE rebinding and PeerConnection shutdown; and
- skips the release attempt when the socket has already failed, allowing bounded local retirement and server-side expiry.

Related issue: #894

## Reproduction addressed

With two relay-only peers and a 600-second TURN allocation lifetime:

```text
initial: connected=true  rpc=ok  allocations=2

PeerConnection close:
  after 3.0s: old_allocations_present=2  total_allocations=2
  fixed:      Refresh(0) is sent before the driver exits

ICE restart:
  recovery RPC: success
  before restart: allocations=2
  after restart:  old_allocations_present=2  total_allocations=4
  fixed:          old generation is retired before old sockets are rebound
```

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy --fix --lib --bins --all-targets --allow-dirty --features runtime-tokio`
- `cargo test --lib peer_connection::transports::turn_relayer --features runtime-tokio` (5 passed)
- `git diff --check`


src/peer_connection/driver.rs
```diff
@@ -327,6 +327,12 @@ where
     async fn bind_transports(&mut self) -> Result<()> {
         let runtime = Arc::clone(&self.inner.runtime);
 
+        // Retire TURN allocations while the old UDP sockets are still alive.
+        // Rebinding drops those sockets below; sending Refresh(0) afterwards
+        // would leave the server allocation until expiry.
+        self.turn_relayer.close()?;
+        self.poll_writes().await?;
```
rainliu, line 334: add
```
 if !self.udp_sockets.is_empty() {
```
 A one-line guard. Today the core queue is empty at first bind so nothing is lost, but the call buys nothing there and leaves a silent-drop path for whoever changes queueing order later.

alinejun: Addressed in e77f6784c. bind_transports now calls TURN retirement and flush only when an existing UDP generation is present; the initial bind skips the no-op flush. The focused turn_relayer tests still pass (5 passed).

rainliu (approved):

PR #900, opened 2026-09-20
fix: skip STUN servers without a matching address family

kaito-harry:
A dual-stack host configured with an IPv4-only STUN server logs an ERROR for each IPv6 local address, even though that server simply has no matching address family. The same happens for IPv4 local addresses with an IPv6-only server.

Resolve each STUN URL once per gathering pass and skip local addresses without a matching resolved server address before creating a client or transaction. Preserve both IPv4 and IPv6 host candidates, and retain error reporting for actual resolution and client-creation failures. This follows the existing TURN gatherer's handling of address-family mismatches.

Add four regression tests using `MockRuntime`, covering single-stack and mixed-family resolved addresses, preservation of dual-stack host candidates, completion without transactions when all addresses are incompatible, and continued gathering after a server fails resolution. The existing `runtime-mock,crypto-ring` CI job includes these tests.

Related: https://github.com/Actrium/actrium/issues/247. This addresses the deterministic address-family ERRORs; transaction timeouts require separate investigation because mismatches return before a transaction is created.


rainliu:
please fix CI test [cargo / Unit Test with runtime-mock,crypto-ring](https://github.com/webrtc-rs/webrtc/actions/runs/35505144012/job/106112613263?pr=900)

rainliu (approved):

